PR #164, opened 2020-10-23
ST6RI-279 Enable JupyterLab syntax highlighting

ivan-gomes:
Also integrated into the multi-project Maven build with dependencies, e.g. node, yarn, downloaded dynamically (@himi).

@blake-regalia tag for credit and review. :)


himi:
It worked in my JupyterLab environment with 86d486e.  That's wonderful!
After the build, I noticed "version" in org.omg.sysml.jupyter.jupyterlab/package.json was updated to "0.5.0-SNAPSHOT"  It's better to update this file, I think.

ivan-gomes:
> It worked in my JupyterLab environment with [86d486e](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/commit/86d486e42e3216c23b23bfe6af9c134cd426d01b). That's wonderful!
> After the build, I noticed "version" in org.omg.sysml.jupyter.jupyterlab/package.json was updated to "0.5.0-SNAPSHOT" So it should be updated or added to .gitignore.

Thanks for the Gradle Windows fix and the note about version. Updating the version in `package.json` with `yarn version` is the only way to pass along the Maven build's version I could find. Updated it to the current version.

himi (approved):

seidewitz (approved):
I am not entirely comfortable with having to maintain two essentially identical Code Mirror configuration in two separate script files. However, I can accept this for this sprint, as long as we implement automatic update of at least the keyword configuration in the next sprint.

seidewitz:
@blake-regalia Were you going to review this?

blake-regalia:
Will follow up in another ticket to rename package to be published under `systems-modeling` npm org.

PR #165, opened 2020-10-31
ST6RI-181 Update implementation for 2020-10 metamodel baseline

seidewitz:
@ivan-gomes @manasbajaj 

The `SysML.uml` files on the ST6RI-281 branch is now updated for the 2020-10 metamodel baseline. As far as I can tell, this update really didn't make any substantive change in the metamodel.


ivan-gomes (approved):
Confirmed that there are no changes in the metamodel that affect Track 5 code generation.

PR #170, opened 2020-11-23
ST6RI-284 Effective names for requirement constraints

seidewitz:


himi (approved):

himi (approved):
I confirmed that it worked with PlantUML visualization.

PR #171, opened 2020-11-24
ST6RI-294 Verification case concrete syntax and library model (updated)

seidewitz:
Also covers ST6RI-295 Verification validation case.


himi (commented):
VerificationCase::subj is no longer redefined as Case::subj so that VerificationCase has two subj, Case::subj and VerificationCase::subj.  Is it intended?  Or, does it come from some limitations?

seidewitz:
> VerificationCase::subj is no longer redefined as Case::subj so that VerificationCase has two subj, Case::subj and VerificationCase::subj. Is it intended? Or, does it come from some limitations?

Any Case is required to have exactly one subject and one objective, which must redefine the subject and objective of any Cases it specializes. If such redefinitions are not modeled explicitly, then they are inserted implicitly. So, `VerificationCase::subj` and `VerificationCase::obj` still redefine `Case::subj` and `Case::obj`, even though this is not shown explicitly. Of course, the explicit redefinitions that had been given previously were not wrong either, but I changed them for a technical reason. 

All the `verify` requirements in the objective of a VerificationCase get an implicit Subsetting to `VerificationCase::obj::requirementVerifications`. To insert this Subsetting, the system needs to look up the target element using the full global qualified name `Verifications::VerificationCase::obj::requirementVerifications`. However, for performance reasons, the system does not use the full name resolution algorithm for looking up implicit generalization targets, but rather, uses a more limited (but much faster) Xtext shortcut -- which doesn't recognize the effective name mechanism. So, when the VerificationCase objective was defined as `objective :>> obj`, the qualified name for it was not resolving for the purpose of the implicit Subsetting. It needed to have an explicit (not just effective) name.

Now, I could have changed the objective declaration to `objective obj :>> obj`, but then it seemed simpler to just rely on the implicit redefinition. Once I did this, I also changed the declaration of `subj` for consistency. But, perhaps, in retrospect, it would be better to be more explicit in library models (as has actually already been done in some of the other library models).

himi (approved):
Thank you for the detailed explanations.  I could read how these are treated as redefinition.  And I confirmed the test cases worked properly.

himi:
This redefinitions look better and clearer.  Thank you for this improvement.

PR #175, opened 2020-12-03
ST6RI-301 Membership::memberName not properly deserialized in XMI

seidewitz:


seidewitz:
@rbarbau 

Please test this latest revision and see if it still works for you.

rbarbau (approved):
This fixed the deserialization issue

PR #177, opened 2020-12-05
ST6RI-210 Provide separate storage of default generalizations in EMF representation

seidewitz:
This is a significant update the handling of implicit generalizations (that is, generalizations, generally to library elements, that are inserted by the system, not explicitly by the modeler). The system now initially just caches a record of what implicit generalizations are required, with sufficient information to be used for name resolution and typing, but the physical generalization relationships are only added into the model if explicitly requested. This change is being made because most EMF-based tooling does not expect new meta-elements to be created simply as a result of reading a model, which is what was happening before for implicit generalizations. (For more details, see [https://openmbee.atlassian.net/browse/ST6RI-210](https://openmbee.atlassian.net/browse/ST6RI-210).)

This change primarily impacts how tooling interacts with a model via the in-memory EMF API. However, the intent is that, for the purposes of non-EMF interchange between, all implicit elements can still be included, as before. This has already been rather thoroughly tested, but we need to a final review before including it in the 2020-11 release.

@himi 
The PlantUML visualization seems to continue to work with this update. The main change is that the methods that return various kinds of generalizations (e.g., `getOwnedSuperclassing`, `getOwnedSubsetting`, etc.) no longer include implicit generalizations. There is another method that could be used to get the implicit generalizations, but since we generally don't do _not_ want to include these implicit generalizations in the visualization anyway, I don't think anything actually needs to be done here. Note, however, that `Feature::getTypes` still _does_ include the types that come from implicit generalizations. Please review and respond as to whether you concur.

@manasbajaj 
We had been planning to update the repository publishing so that it continued to include implicit generalizations, as before. However, I decided not to do that at this point. A model published without the implicit generalizations will also not include the dangling elements for the library elements referenced by those implicit generalizations, which eliminates a lot of the clutter we currently have in the published models (see, for example, the projects `Test` and `Test2` that I pushed to the `dev` repository today, which do not include implicit generalizations, and each only have one root, as would be expected). This seemed to be reasonable to me, at least until we implement proper resolution of library elements to reusable library models in the repository. (Note, however, that any explicit references to library elements made by the modeler, e.g., from the Quantities and Units libraries, will still be included, as before.) Please review and respond as to whether you concur.

@rbarbau
The XMI generation utility has been updated so that all implicit generalizations are added to the model before the XMI is generated. We believe that we have validated that the XMI generated by the updated utility is logically the same as before, except for non-consequential changes in ordering and the inclusion of some additional binding connectors. The additional binding connectors are because we are now also more consistently applying other model transformations than just the implicit generalizations -- not including those errors before was actually an error. But please check that the XMI generated by the utility on this branch still works with your tooling. (Note that this branch includes the bug fix made in ST6RI-301.)

Thanks in advance to everyone for a timely review of this change, so we can get it into the 2020-11 release.


himi:
@Ed, thank you for the clarification of the impacts by this change.  The current PlantUML-SysML visualization uses getOwnedGeneralization() to visualize subsettings and getOwnedRedefinition() for redefinitions, and getType() for definedBy/typeOf relationships.  So I do not think this change has no direct impact on PlantUML visualizations but that means only definedBy visualization would have implicit ones.   If you have any concern on this treatment, please let me know.  I'd like to check it later.

seidewitz:
@himi 
Actually, only directly owned FeatureTyping relationships should be visualized as "defined by" or "type of" (just as only directly owned relationships are visualized for superclassing, subsetting and redefinition). The `getType` method includes types derived from subsettings, which is redundant with showing the subsetting relationships. The  FeatureTypings returned by the`getTyping` method would not include implicit FeatureTypings.

org.omg.kerml.xtext/src/org/omg/kerml/xtext/util/KerMLRepositorySaveUtil.java
```diff
@@ -179,6 +180,9 @@ public void run(String[] args) {
 				this.initialize();				
 				this.read(args);
 				
+				System.out.println("Transforming...");
+				ElementUtil.transformAll(this.resourceSet, true);
```
ujhelyiz, line 184: @seidewitz @manasbajaj If we don't want to upload the implicit generalizations, as you have written in the pull request description, I believe we should remove this update from the repository save utility, as it is responsible for adding the implicit generalizations to the model (although it is possible that it does not work as intended, as I did not manage to test this functionality in detail).

seidewitz: Ah, yes, you are correct. I had been thinking of adding a command argument option on whether to add the implicit generalizations or not, but I forgot. Thanks for noticing this.

It would also be possible to add an option for this for the `%publish` command, of course, but that didn't seem like something our Jupyter users would particularly care about.

himi:
Can you merge it from the latest master?  I'll try to use `getTyping ()` to visualize `definedBy` relationships.  Or I can do it by myself if you're ok.

seidewitz:
@himi Merge with master is done.

himi:
@ed, thank you.  I checked the current code and I found I was wrong.  The current visualization code visualizes FeatureTyping and it works properly.  However, to show a text of the type, for example `Part` in `p: Part` is taken from the first item of `getType()` as you advised if my memory is correct.  However, recently Sandy claimed such general types should not be shown and maybe we should use `getOwnedTyping()` for this purpose instead of `getType()`

seidewitz:
@himi Go ahead and commit the change on this branch. Thanks.

himi:
Done.

himi (approved):

ujhelyiz (approved):
Looks good to me.

rbarbau (approved):

rbarbau (approved):

PR #179, opened 2020-12-09
ST6RI-308 Adds Xtend compiler to the maven build configuration

ujhelyiz:
This change adds required configuration to the maven build to generate
Java files to the xtend-gen folders. This required adding extra pom.xml
files to the projects, as it is not possible for the previously used
tycho-pomless to add this project-specific configuration otherwise.

### DIFF
--- a/org.omg.kerml.xtext.ide/pom.xml
+++ b/org.omg.kerml.xtext.ide/pom.xml
@@ -1,0 +1,34 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>org.omg.sysml</groupId>
+		<artifactId>org.omg.sysml.parent</artifactId>
+		<version>${revision}</version>
+	</parent>
+
+	<artifactId>org.omg.kerml.xtext.ide</artifactId>
+	<packaging>eclipse-plugin</packaging>
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-clean-plugin</artifactId>
+				<executions>
+					<execution>
+						<id>gen-clean</id>
+						<phase>clean</phase>
+					</execution>
+				</executions>
+			</plugin>
+			<plugin>
+				<groupId>org.eclipse.xtend</groupId>
+				<artifactId>xtend-maven-plugin</artifactId>
+			</plugin>
+		</plugins>
+	</build>
+</project>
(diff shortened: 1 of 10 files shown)


ivan-gomes (approved):
Success! `mvn package` works from a fresh clone with these changes and the `org.omg.sysml.jupyter.installer` installed JupyterLab environment is able to successfully execute the smoke test notebook @seidewitz provided.

@ujhelyiz wanted to call out a few kinds of warnings that Maven is now reporting though.

```
[INFO] --- xtend-maven-plugin:2.22.0:xtend-test-install-debug-info (default) @ org.omg.sysml ---
[WARNING] Directory /path/to/SysML-v2-Pilot-Implementation/org.omg.sysml/src/test/java is empty. Can't process.
[WARNING] Directory /path/to/SysML-v2-Pilot-Implementation/org.omg.sysml/src/test/generated-sources/xtend is empty. Can't process.
```
Perhaps we can configure the Maven plugin to exclude the empty dirs, if appropriate.


```
[INFO] --- xtend-maven-plugin:2.22.0:compile (default) @ org.omg.kerml.xtext ---
[WARNING] 
WARNING: 	KerMLScopeProvider.xtend - /path/to/SysML-v2-Pilot-Implementation/org.omg.kerml.xtext/src/org/omg/kerml/xtext/scoping/KerMLScopeProvider.xtend
85: Unnecessary cast from QueryPathExpression to QueryPathExpression
[WARNING] 
WARNING: 	KerMLGlobalScope.xtend - /path/to/SysML-v2-Pilot-Implementation/org.omg.kerml.xtext/src/org/omg/kerml/xtext/scoping/KerMLGlobalScope.xtend
108: Unnecessary cast from Package to Package
```

himi:
Looks good.  If we use xtend for more projects, I think it's better to put xtend-maven-plugin to the root pom.xml.  In the future, I think xpect test cases will be executed in the test lifecycle, and it's using xtends as well.  @ujhelyiz how do you think?  Or do you want to explicitly specify it in the child pom.xml?  It's just for clarification.

Also I think we should execute mwe workflow as well, but it should be in the separate thread.

ujhelyiz:
@ivan-gomes I have tried to remove the warnings from the compiler, partially by removing functions we are not relying on from Xtend. Now the mentioned warnings should not appear anymore.

@himi I put the configuration of the xtend-maven-compiler to the parent pom (into the pluginManagement section). We could try to add the same to build, but a few years ago that was not working (e.g. it failed when maven tried to execute the Xtend compiler on a non-Java project); and if most of our projects are not using Xtend, it also slows down the build.

About Xpect tests, those are not part of the build yet, so I did not add any configuration there, but yes, if it were part of the build, it should be added as well. About the testCompile lifecycle, it does not seem necessary to me as there are no testSources classpath entries for that project. If that were to change, then the configuration is to be updated accordingly.

The MWE execution is also possible from the build, but running that is a bit more tricky. Given the Xtext generated code is already uploaded to the repository thus there are no issues with running the build locally, I would not try to add its generation to the build before the current release is finished.

seidewitz:
I think we are good now for this release.

@ivan-gomes Please redeploy to JupyterHub, and then I will do a final check that everything works there.

Thanks for getting this fix in so expeditiously!

himi (approved):

ivan-gomes:
> @ivan-gomes Please redeploy to JupyterHub, and then I will do a final check that everything works there.

Done.

seidewitz (approved):
Looks good now!

PR #183, opened 2020-12-23
ST6RI-307 Move syntax transformations from getters to the derived state computer

seidewitz:
Please review when possible, given the holiday.


ujhelyiz (approved):
In general, I am fine with the updates here but the one difference is the `ExpressionImpl::getFeaturesRedefinedByType` method as if that is still called by a normal getter it would still cause problems with EMF transactions.

However, given I won't really have time to have a deeper look at that single part until January, it might be a good idea to merge this change as it is (to avoid potential issues with merging), and fix the related issue separately.

seidewitz:
@ujhelyiz 
The latest commit has an update that removes the calls to `getInput` and `getOutput` from `ExpressionImpl.getFeaturesRedefinedByType`.

ujhelyiz:
> @ujhelyiz
> The latest commit has an update that removes the calls to `getInput` and `getOutput` from `ExpressionImpl.getFeaturesRedefinedByType`.

This looks totally reasonable to me. Thanks.

PR #186, opened 2020-12-29
ST6RI-321: Spurious "(T,blue)" shown in header of various def elements (PlantUML)

himi:
To ensure to use the fixed version of PlantUML, set the dependent version to the latest one.

I also fixed the default style is not properly chosen in Jupyter environment when we specify --style comptree with 
9d60bf2

### DIFF
--- a/org.omg.sysml.plantuml/META-INF/MANIFEST.MF
+++ b/org.omg.sysml.plantuml/META-INF/MANIFEST.MF
@@ -7,7 +7,7 @@ Bundle-Version: 0.8.0.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Export-Package: org.omg.sysml.plantuml
 Import-Package: com.google.inject;version="1.3.0",
- net.sourceforge.plantuml;version="1.2020.14.himi5",
+ net.sourceforge.plantuml;version="1.2020.14.himi6",
  org.eclipse.emf.common,
  org.eclipse.emf.common.util,
  org.eclipse.emf.ecore,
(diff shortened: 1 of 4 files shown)


seidewitz:
When I attempt to visualize a model in a runtime Eclipse instance, I now get the exception

`org.eclipse.core.runtime.CoreException: Plug-in "net.sourceforge.plantuml.eclipse" was unable to instantiate class "net.sourceforge.plantuml.eclipse.views.PlantUmlView".`

I have 1.2020.14.himi6 installed, and it worked before the latest updates on this branch.

seidewitz:
So, I have confirmed that the latest updates on this branch work when deployed in Jupyter or installed in Eclipse from the update site. 

However, I still can't get the view to open in a runtime Eclipse instance, even when I go back to before the commits on this branch, where it used to work, and even after uninstalling and reinstalling the PlantUML plugins. It seems that the ultimate cause of the runtime `CoreException` is

```
 java.lang.LinkageError: loader constraint violation: when resolving overridden method "net.sourceforge.plantuml.eclipse.views.AbstractDiagramSourceView.createPartControl(Lorg/eclipse/swt/widgets/Composite;)V" the class loader (instance of org/eclipse/osgi/internal/loader/EquinoxClassLoader) of the current class, net/sourceforge/plantuml/eclipse/views/AbstractDiagramSourceView, and its superclass loader (instance of org/eclipse/osgi/internal/loader/EquinoxClassLoader), have different Class objects for the type org/eclipse/swt/widgets/Composite used in the signature
```

Something seems to be screwed up in my Eclipse installation now, but I am not sure what it is. Has anyone seen this before?

@himi Can you confirm that you can open the PlantUML view in a runtime Eclipse instance?

seidewitz (approved):

PR #187, opened 2020-12-29
ST6RI-306 (re-opened) Enumeration concrete syntax and library model

seidewitz:
**NOTE.** I am opening a new pull request for ST6RI-306, because I seem to have accidentally merged PR #185 for it without waiting for any review. Not quite sure how that happened...

This is an interim implementation of enumerations, in order to provide a basic capability and surface notation in SysML while related semantic discussions continue on including a more general concept of "enumeration" in KerML. Per the 2020-12 abstract syntax baseline, enumeration definitions are implemented as a kind of variation, with the allow enumeration values being declared as variants. This allows the basic typing semantics of variations on enumerated variants to be adopted for enumeration types without them necessarily being considered "variant modeling" to the user.

The following limitations of the current approach should be kept in mind during review.

1. Enumeration definitions are kinds of attribute definitions, so enumeration usages are limited to attributes.
2. The members of an enumeration definition are limited to the enumeration usages used to define the enumerated values of the enumeration. It is not possible to directly define other features as part of the enumeration definition. However, an enumeration definition can specialize a regular attribute definition and inherit common features for its enumerated values in that way.
3. Since an enumeration definition is already a variation, it is not possible to define a variation enumeration definition (though it is not clear that this would be that useful anyway).

Examples of the new enumeration notation can be found in `sysml/src/examples/Simple Tests/EnumerationTest.sysml` and `sysml/src/validation/15-Properties-Values-Expressions/15-10-Primitive Data Types.sysml`.

In the end there was no library model added for enumerations. This is because it doesn't really make sense to have a base enumeration definition in the sense of generalization hierarchy. Specialization restricts the extent of a subtype relative to its supertype, while extending an enumeration definition should _add_ values in the extension. (This is one of the areas still under discussion for KerML.)


sysml/src/examples/Simple Tests/EnumerationTest.sysml
```diff
@@ -0,0 +1,35 @@
...
+		}
+		enum blue {
+			:>> val = 1;
+		}
+		enum green {
+			:>> val = 2;
+		}
+	}
+	
+	enum color : ColorKind;
```
hpdekoning, line 24: For me, allowing this use of the `enum` keyword as an alternative for an enum attribute is potentially confusing, because in the future it would not be clear whether the usage is an attribute, an item, a part, ... Also it is effectively a homonym with the `enum` used for the 'enumeration literals'. I would propose to reserve the sole `enum` keyword only for declaration of those 'enumeration literals'.

I understand that with `enum def` the logical consequence is to also have `enum` as a usage. Perhaps we could instead use:
```
    enum attribute def ColorKind :> Color {
        enum red {...}
    ...
    }
```
as a kind of special / qualified attribute and then only allow usage of such an attribute type via the normal `attribute` keyword, like in the next line `attribute color1 : ColorKind = ColorKind::blue;`. The same pattern could later also be used for Occurrences and Objects, i.e. `enum item def`, `enum part def`, `enum port def`. Their related `item`, `part` and `port` usages would then use existing normal syntax.

sysml/src/examples/Simple Tests/EnumerationTest.sysml
```diff
@@ -0,0 +1,35 @@
...
+	attribute color1 : ColorKind = ColorKind::blue;
+	
+	/**
+	 * The "enum" keyword is optional for EnumerationUsages used to defined the
+	 * enumerated values of an EnumerationDefinition.
```
hpdekoning, line 29: typo: defined -> define

hpdekoning:
Certainly the surface syntax looks intuitive and readable for end-users, my only reservation is the comment above on use of the `enum` keyword to declare an attribute usage (of an EnumerationAttributeType).

Further, I wonder if it would already be possible to create a validation example that extends an existing enum def, i.e. by creating a wider/super type with more enumeration values. Similar to the MoreColorKind we had in the prototype.

sysml/src/examples/Simple Tests/EnumerationTest.sysml
seidewitz: The enumerated values of an enumeration definition _are_ enumeration usages, syntactically the same thing as enumeration usages used outside of an enumeration definition. So, the use of the `enum` keyword isn't really a "homonym" as such, but just a consistent use of that keyword to mark enumeration usages.

On the wider point of ultimately extending the syntax to allow enumerations of kinds of things other than attributes, I also felt some concern about that. But I am also loath to add complexity to the syntax at this point when we don't know for sure what the general capability will be yet, or even if we will have time to get it into the language. And, even if we ultimately have general syntax like `enum attribute def`, `enum part def`, etc., I think I would want to keep `enum def` as a shorthand for `enum attribute def`, because that is still going to be very common and should be simple.

On the other hand, I would agree that allowing enumeration usages outside enumeration definitions doesn't really provide any additional capability over just using an attribute usage typed by an enumeration definition. Currently, however, the abstract syntax allows the use of enumeration usages like any other usages, and if we allow that in the abstract syntax, we should have a way to notate it in the concrete syntax. But perhaps we shouldn't allow that in the abstract syntax, either -- it would be easy to require that an enumeration usage be owned by an enumeration definition.

It might be worth a discussion with Track 2 (or Sandy, at least) whether it would be more confusing at this point to allow enumeration usages outside of enumeration definitions for now, when that might not make sense in the future, or to introduce `enum def`s as the syntax now, but without allowing corresponding `enum`s outside of the defs.

sysml/src/examples/Simple Tests/EnumerationTest.sysml
seidewitz: Fixed.

seidewitz:
> Further, I wonder if it would already be possible to create a validation example that extends an existing enum def, i.e. by creating a wider/super type with more enumeration values. Similar to the MoreColorKind we had in the prototype.

Unlike KerML, the SysML notation does not allow for the definition of generalizations that are not owned by the subtype in the relationship. Therefore, it is not possible to model the "expansion" of an enumeration in the way it was done in the KerML prototype. I should probably have listed this as a limitation of the current SysML enumeration implementation.

PR #194, opened 2021-01-27
ST6RI-327: Implement implicit binding connectors

ujhelyiz:
The commits in this pull request update the creation of binding connectors during the transformation steps in multiple ways:

1. Instead of relying on checking for null values, the linker was updated to correctly remove the binding connectors in the clean phase. The removed elements will be readded by the linker/derived state computer.
2. Binding connectors are not added to the containment hierarchy but are stored as implicit entries. This required to update the validation, outline generation and the implicit element generation transformation in ElementUtil.
3. I have verified that all tests run successfully, all models in the repository compile correctly and when serializing the SysML library model elements to XMI the output is the same as master.


seidewitz (approved):
The implicit BindingConnectors where not showing up for OperatorExpressions, because they have a special method in `KerMLOutlineTreeProvider` that overrides the general one for `Type`. This was easily fixed by adding a call to `createImplicitBindingConnectorNodes` to the method for OperatorExpression.

Other than this, it all looks great!

PR #195, opened 2021-01-29
ST6RI-303 Element filter notation

seidewitz:
This should be considered a work-in-progress pull request for ST6RI-303. It includes the following updates:

- Adds the basic notation `filter <expression>;` that can be used to specify an element filter in a package, view definition or view usage.
- Updates the import notation to `import <qualified-name>::* [<expression>];` that combines a package import with filtering of the imported elements. (This can also be used with recursive import by replacing `*` by `**`.)
- Adds a symbolic shorthand `@` for the `istype` (previously `instanceof`) operator and allows it to be used as a unary operator by defaulting the first argument to `self`. (This is to provide a convenient notation for testing that a metadata annotation is applied, in way that looks similar to specifying that annotation.)

It does _not_ include any validation that a filter expression is model-level evaluable.

At this time, I don't plan to merge this pull request into master until the following issues are also completed:

- ST6RI-304 Metadata expression evaluation
- ST6RI-305 Package membership/name resolution with element filtering

However, I created the pull request in case there might be interest in reviewing the notation update, even though no actual filtering happens right now.


ujhelyiz (commented):
Looks reasonable to me in its current form.

PR #198, opened 2021-02-04
ST6RI-304 Metadata expression evaluation

seidewitz:
This is a work-in-progress pull request for ST6RI-304. It also includes the commits for ST6RI-303 (see also PR #195). It implements an initial set of model-level evaluable functions, but does not include package element filtering (ST6RI-305). Additional functions will likely be added in the future.

Without integration with element filtering, Xpect cannot be used for testing the results of model-level evaluations. Instead, some testing is provided by Junit tests based on SysMLInteractive.


org.omg.sysml/src/org/omg/sysml/expressions/AtFunction.java
```diff
@@ -0,0 +1,10 @@
+package org.omg.sysml.expressions;
```
ujhelyiz, line 1: This file is missing the copyright header. I am not sure how important it is here, but the other classes of this package include the header, so I wanted to mention it.

ujhelyiz (commented):
Looks reasonable to me, I only found a missing copyright header I wanted to mention here.

org.omg.sysml/src/org/omg/sysml/expressions/AtFunction.java
seidewitz: I added the missing header.

PR #199, opened 2021-02-05
ST6RI-318: Automate test execution

ujhelyiz:
This change adds the Xpect tests to the Maven build and executes it. Travis CI does not seem to be able to display test results, but as far as I can see the maven build log should contain enough information about failed tests to identify what we need to check.


seidewitz (approved):
Very nice. This will be quite useful for local builds as well as Travis CI.

For Travis, I executed a build with a failing test, and what failed was quite clear from the log, as one would expect.

PR #203, opened 2021-02-12
ST6RI-339 Separates common expression sublanguage for KerML and SysML

ujhelyiz:
This language initializes a new set of Xtext projects that contain the parts required for expressions that are shared between them. The new language only contains the grammar customizations, but does not include any customization related to linking or validation - those are still in the original language projects.


org.omg.sysml.xtext/src/org/omg/sysml/xtext/SysML.xtext
```diff
@@ -2173,348 +2092,17 @@ RenderingRefUsage returns SysML::RenderingUsage :
...
-	ownedFeatureMembership_comp += FeatureReferenceMember
-;
-
-FeatureReferenceMember returns SysML::ReturnParameterMembership :
-	ownedMemberFeature_comp = FeatureReference
+	 KerMLExpressions::OwnedExpression | QueryPathExpression
 ;
 
 // Note: Use subsetting here in order to inherit typeing of referent.
+@Override
```
ujhelyiz, line 2101: The instance in KerML was returning a SysML::FeatureReferenceExpression instead of ReferenceUsage - if the return types can be unified, it is possible to remove this overridden rule from sysml.xtext.

org.omg.kerml.xtext/src/org/omg/kerml/xtext/KerML.xtext
```diff
@@ -454,14 +423,14 @@ OwnedConjugation returns SysML::Conjugation :
 /* Feature Membership */
 
 FeatureMember returns SysML::FeatureMembership :
-	TypeMemberPrefix FeatureMemberFlags 
+	DefinitionMemberPrefix FeatureMemberFlags
```
ujhelyiz, line 426: The rename of TypeMemberPrefix to DefinitionMemberPrefix was done as sysml.xtext contained the exact same content with a different name and DefinitionMemberPrefix seemed the newer terminology, but the synchronization could happen the other way if necessary.

org.omg.sysml.xtext/src/org/omg/sysml/xtext/SysML.xtext
seidewitz: Actually, in KerML it was returning a SysML::Feature. ReferenceUsage is a kind of Feature. Returning just a plain Feature is fine in the SysML grammar, so the override is not needed.

org.omg.kerml.xtext/src/org/omg/kerml/xtext/KerML.xtext
seidewitz: "Type" is a KerML term. "Definition" is a SysML term. All Definitions are Types, but not all Types are Definitions. "TypeMemberPrefix" should be used in the KerML grammar, not "DefinitionMemberPrefix".

seidewitz (approved):
There were some productions moved into the KerMLExpressions grammar that I hadn't expected. On review, I see why that happened, but the result meant that the KerML and SysML grammars where harder to understand on their own, and they were less aligned with how the grammars are documented in the spec documents. Since the grammar specifications are still being maintained by hand, it is important for maintainability that the Xtext grammars stay reasonably aligned with them.

So, I made some changes to allow most of the productions in question to be moved back into the KerML and SysML grammars, at the expense of re-introducing some redundancy. But I already knew that there was more commonality that needed to be factored out than just Expressions. We can handle that later.

The only production that I really couldn't figure out how to move out of KerMLExpressions was OwnedFeatureTyping. So I left it in KerMLExpressions, but still overrode it in both KerML and SysML (in the places that it had previously been defined in those grammars) with the exact same definition as in KerMLExpressions. This is particularly redundant, but it seems OK for now.

Please review before I merge.

ujhelyiz:
The suggested changes seem reasonable to me - I was not sure about some of the changes I did, that was why I added the specific conversations making them stand out for your review.

I don't really see any good way either to avoid redundancy related to the rule `OwnedFeatureTyping` - either we need an alternative concept in the expression grammar to allow `InvocationExpression`, `BodyExpression` and `TypeReference` to remain general, or all three of these features have to be duplicated in both in the KerML and SysML languages. Neither seems like a good solution to me.

The refactored grammars seem to be reasonable to me, I believe they can be merged into master.

PR #205, opened 2021-02-17
ST6RI-334 Update implementation for 2021-02 metamodel baseline

seidewitz:
Changes in this baseline include:

1. Addition of `AllocationDefinition` and `AllocationUsage` (along with corresponding `Definition::ownedAllocation` and `Usage::nestedAllocation` attributes).
2. Changes to `SendActionUsage` and `AcceptActionUsage`. (`TransferActionUsage` was also deleted, but this was abstract.)
3. Addition of the `Element::qualifiedName` attribute, which is derived to be the string representation of the "ownership qualified name" of an element. (This may be null if the element has no name, or if any of the namespaces in its ownership hierarchy have no name.)

Note that the model library has _not_ yet been updated to add a model for allocations or change the model for send and accept actions. Those changes will be done as part of ST6RI-336 and 338.

**Note:** With the completion of ST6RI-339 (#203), there are now three new Xtext projects that factor out the common expression-language subset from the KerML and SysML branches (`org.omg.kerml.expressions.xtext` and associated `ide` and `ui` projects). These must be imported into the workspace in order to successfully build.

@ivan-gomes @manasbajaj 
The `SysML.uml` file on this branch should be ready for use in updating the API implementation.


himi:
Ed, thank you.  Can I make a branch for ST6RI-342 to test this improvement?  That will effectively check Items 2 and 3.

seidewitz:
@himi
Sure, but move ST6RI-342 into the sprint first and set it to "in progress". Note, however, that the construction of `SendActionUsage` from the textual syntax isn't going to be correct until I complete ST6RI-338.

himi:
I changed the status of ST6RI-342 and added AcceptActionUsage to its target as well.   However, as you noted, I'd need to wait the completion of ST6RI-338 to complete ST6RI-342.

ivan-gomes (approved):
LGTM

PR #207, opened 2021-02-19
ST6RI-338 Send and Accept Action syntax and model library update

seidewitz:
This pull request is for the Pilot Implementation update corresponding to [ST4MD-275 Update TransferAction model to better reflect asynchronous transfer semantics](https://openmbee.atlassian.net/browse/ST4MD-275).

The pull request includes the following Systems Library updates. 

1. In `Actions`, `SendAction` and `AcceptAction` have been added and `TransferAction` has been removed. `SendAction` and `AcceptAction` do *not* specialize `Transfer`.
    - `SendAction` has the following signature, where `items` is the payload to be sent and `receiver` is the target it is sent to.
       **`action def`**`SendAction(`**`in`**`items[0..*],`**`in`**`receiver: Occurrence[1] )`
    - `AcceptAction` has following signature, where `items` is the payload that is received and `receiver` is twhere the payload was actually received (i.e., the "context" object in UML terms, generally a port or some outer part or action containing the AcceptAction).
        **`action def`**`AcceptAction(`**`out`**`items[0..*],`**`in`**`receiver: Occurrence[1] )`
2. In `States`, some elements have been renamed because they were inconsistent with the Track 4 model.
    - `State` to `StateAction`
    - `states` to `stateActions`
    - `Transition` to `TransitionAction`
    - `transitions` to `transitionActions`
3. Also in `States`, the model of `TransitionAction` has been updated to replace the use of a `TransferAction` as a trigger and, instead, bind the `TransferAction::trigger` to the `incomingTransfersToSelf` of an `accepter` that is an `AcceptAction` (note that an `AcceptAction` is constrained to have exactly one `incomingTransferToSelf`).

The pull request includes the following syntax updates:

1. The send notation 
    **`send`***`items-expression`***`to`***`receiver-expression`*`;` 
    now parses to a usage of `SendAction` (with FeatureValues used to bind the parameters):
    **`action`**`: SendAction (`**`in`**`items =`*`items-expression`*`,`**`in`**`receiver =`*`receiver-expression`*`);`
    The methods `getItemsArgument` and `getReceiverArgument` return the *`items-expression`* and *`receiver-expression`*, respectively, for a SendAction.
2. The accept notation now has an additional **`for`** clause. The notation 
    **`accept`***`items-declaration`***`for`***`receiver-expression`*`;` 
    parses to a usage of `AcceptAction`:
    **`action`**`: AcceptAction (`**`out`***`items-declaration`*`:>> items,`**`in`**`receiver =`*`receiver-expression`*`);`
    The **`for`** clause is optional (corresponding to the legacy notation), but the proper defaulting (i.e., what the default "context" object is) is still to be discussed in Track 4.  The method `getReceiverArgument` returns the *`receiver-expression`* for an AcceptAction, if it is given explicitly.


conradbock:
I checked for the three items in [ST4MD-275-comment-12326](https://openmbee.atlassian.net/browse/ST4MD-275?focusedCommentId=12326) and could see the first and last were fixed.  Not sure about the second (transferIn/Out), I couldn't find those in the changes, maybe you redefined incoming/outgoingTransfersTo/FromSelf without renaming?

I noticed changes to SysML state machines that I would need some review to understand.  Even tho these changes in are SysML, I'd like to get back to more discussion between us for changes to such basic things like state machines and interactions.  When I gave input to your Send/AcceptAction model update, you said they were making changes to a model that "was already there", but it wasn't in the December release, so must be quite new.  I know you're very busy, but I'm concerned our normal working relationship on these things got derailed somehow in the last few months.  Can talk on the phone if it's easier.

seidewitz:
@conradbock 

> I checked for the three items in ST4MD-275-comment-12326 and could see the first and last were fixed. Not sure about the second (transferIn/Out), I couldn't find those in the changes, maybe you redefined incoming/outgoingTransfersTo/FromSelf without renaming?

Yes, I simply redefined `incomingTransfersToSelf` and `outgoingTransfersFromSelf` without renaming them. I believe you suggested doing this. I will update the T4 model when we generate the spec documents.

>When I gave input to your Send/AcceptAction model update, you said they were making changes to a model that "was already there", but it wasn't in the December release, so must be quite new.

Previously, both SendActionUsage and AcceptActionUsage were typed by `Actions::TransferAction`, which was both a `Transfer` and an `Action`. SendActionUsage just constrained the start of the Transfer, while AcceptActionUsage constrained the completion of it. This was based on discussions we had some time ago in T4, and it has been in the T6 version since then.

As noted in the PR description, the `TransferAction` model has been replaced by the new `SendAction`/`AcceptAction` model. This was also just what we discussed recently in T4.

> I noticed changes to SysML state machines that I would need some review to understand. 

The renaming of `State` and `states`, as noted in the pull request description, was simply a correction in the T6 implementation. They had been called `StateAction` and `stateActions` in the T4 model for some time, which are, I believe, the correct names we had agreed on in T4.

However, your comment made me realize that I forgot to mention that there was also a change to `TransitionAction` that I forgot to include in the PR description. Previously, `TransitionAction` was called `Transition`, and it used a `TransferAction` to specify its `trigger`, in order to be consistent with the syntax and semantics of accept actions as used in action models. This worked because`TransferAction` was a kind of `Transfer`, as required for a `trigger`. But this no longer works with the new model, because `AcceptAction` is no longer a kind of `Transfer`.

In order to fix this with a minimal change to the abstract syntax and library model, I updated `TransitionAction` to replace, as directly as possible, its owned `TransferAction` with an `AcceptAction`. However, instead of this now being itself the `trigger` for the `TransitionAction`, I updated the `TransitionAction` model so that the incoming Transfer of the `AcceptAction` is bound to the `TransitionAction` trigger. (This is consistent with the T4 model as updated for the 2020-02 baseline.) The intent is to have the same effect as the previous model, pending further discussion of send and accept actions.

Of course, even though I was trying to make a minimal change here, it was enough to break the PlantUML rendering of triggers on state diagrams. So for that reason, also, I really should have included this change in the PR description! I will update the description now, for the record. Sorry for the oversight.

seidewitz:
@himi 
As noted in the PR description, the parsing of the accept notation has changed. The item to be accepted is now parsed as a parameter redefinition, rather than as an ItemFeature. As a result, the PlantUML rendering of a transition trigger no longer works. I expect you are also dealing with this as part of the accept action visualization implementation for ST6RI-342. Would you like to also fix the visualization for transition triggers as part of ST6RI-342, or would you prefer to fix it on this branch?

@santileortiz 
I expect this will also impact accept action and transition visualization in your tool.

seidewitz:
I also added the resolution to ST6RI-346 (Transition sources are not being parsed correctly) to this branch, because it could be annoying to test updates to state model visualization with that bug being fixed.

conradbock:
Thanks for the details on the SysML state machine changes.  I'd like to discuss in a Wednesday call if possible, since T4 tends not to follow the library changes (even more so in the interaction calls with E,O,&Sandy attending).  These are in SysML, but have more complicated semantic issues than the usual specialization of Kernel for SE applications.  Can talk on the phone if it's easier.

conradbock (approved):

himi (approved):
I'll work on ST6RI-342 after the close of this pull request.

seidewitz:
@himi
I put in a quick fix for visualization of transition triggers, by simply setting `triggerString = getText(s)` in `VStateMembers.convertToDescription`. This way, the accept triggers at least show up, but the expectation is that the full fix will be made as part of ST6RI-338.

himi:
@seidewitz, I get noticed.  I'll work on that part after merging your fix.   Thank you.

PR #209, opened 2021-02-25
ST6RI-335 Create Xpect tests for package import with element filtering

seidewitz:


miyakowilson (approved):
I tested all kerml and sysml tests and all passed.

PR #212, opened 2021-03-02
ST6RI-350 Web page freezes in JupyterLab when an unmatched single quote is typed

seidewitz:
Please see description and analysis in Jira issue [ST6RI-350](https://openmbee.atlassian.net/browse/ST6RI-350).


ivan-gomes (approved):

blake-regalia:
CodeMirror's StringStream returns `undefined` instead of null when EOF is reached. Fix was to remove null comparison and allow truthy test to take place on assignment expression in while loop conditional.

blake-regalia (approved):

PR #216, opened 2021-03-18
ST6RI-351 Update implementation for 2021-03 metamodel baseline

seidewitz:
This pull request implements the 2021-03 update to the metamodel baseline.

**Owned Relationships**

The major change in the abstract syntax for 2021-03 is that all properties that directly or indirectly subset `Element::ownedRelationship` are now derived. In particular, `Namespace::ownedImport`, `Namespace::ownedMembership` and `Type::ownedFeatureMembership` are now derived as all `ownedRelationships` that are Imports, Memberships or FeatureMemberships, respectively. (Note that `ownedGeneralization` and properties that further subset it were already derived in prior metamodel baselines.)

The primary reason for this change is to ensure that the abstract syntax preserves the ordering of owned Relationships when parsed from the textual notation. Doing this is necessary so that it will be possible to serialize the parsed abstract syntax representation of a model back to textual notation and preserve the ordering of declarations in the original textual representation. Note that, even though `ownedRelationships` was already an ordered property, UML does not define the ordering across values inserted into different subsetting properties of an ordered subsetted properties. Indeed, ipreviously in the implementation, the ordering for the `ownedRelationships` of a Namespace had all Imports grouped at the beginning, followed by all Memberships, followed by all FeatureMemberships, regardless of how the declarations of these different kinds of Relationships were intermixed in the original textual representations.

In the new baseline, all owned Relationships are inserted into the `ownedRelationship` property with a total ordering regardless of the subtype of the Relationship. Specific properties like `ownedMembership` are then computed as the sublist of elements of `ownedRelationship` with the appropriate subtype (perserving the ordering of the original list). This also eliminates the need for the "`_comp`" versions of the subsetting properties.

**Owned Related Elements**

Relationships are owned by Elements via the `Element::ownedRelationship` property, and those Relationships can then own other Elements using the `Relationship::ownedRelatedElement` property. The new metamodel baseline does _not_ make properties that subset or redefine `ownedRelatedElement`. This means that our existing custom EMF importer for UML _would_ still generate "`_comp`" versions of these subsetting/redefining properties.

In order to eliminate the need for "`_comp`" altogether, the custom importer has been updated to treat all subsets and redefinitions of single-valued composite properties as derived, even if they are not marked as such in the original UML/MOF metamodel. The Xtext grammars are then updated to insert the owned Elements of a Relationship directly into `ownedRelatedElement` rather than using the subsetting or redefining property. Specific properties like `ownedMemberElement` are then computed as the first element of the `ownedRelatedElement` list of the Relationship.

**Impacts**

1. _EMF._ The Java interfaces generated for EMF no longer have methods for the `_comp` version of composite properties. 
   * Owned Relationships should now always be inserted into the list obtained from calling `getOwnedRelationship` on an Element. Getters for more specific sublists (e.g., `getImport`, `getMembership`, `getFeatureMembership`, etc.) work as before. Inserting into the lists obtained from these getters is not allowed.
    * Both the getters and setters can be used for the single-valued properties for owned Elements of a specialized Relationships. For example, on a Membership, `getOwnedMemberElement` will get the `ownedMemberElement` and `setOwnedMemberElement` will set it. These setters have the same effect as getting and setting the first Element of the list obtained using`getOwnedRelatedElement`.

2. _XMI_. Tags with `_comp` suffixes will no longer appear in the XMI representation of a model. All Relationships owned by an Element will be serialized using nested `ownedRelationship` XML elements. All related Elements owned by a Relationship will be serialized using `ownedRelatedElement` XML elements. For example, a Membership that was formerly serialized to XMI as
   ```
    <ownedMembership_comp memberName="A" ...>
        <ownedMemberElement_comp xsi:type="sysml:PartDefinition">
            ...
        </ownedMemberElement_comp>
    </ownedMembership_comp>
   ```
   will now be serialized as
   ```
    <ownedRelationship xsi:type="sysml:Membership" memberName="A" ...>
        <ownedRelatedElement xsi:type="sysml:PartDefinition">
            ...
        </ownedRelatedElement>
    </ownedRelationship>
    ```

3. _API_. The `_comp` properties from the in-memory representation of a model were already _not_ being saved to the repository, and derived properties _are_ being saved. So this change should not effect how models are stored. However, it may impact how the server-side metamodel validation and client-side API code are generated.

@manasbajaj @ivan-gomes 
The updated UML metamodel is available as usual at `org.omg.sysml/model/SysML.sysml` on the ST6RI-351 branch.

@YvesTLS 
The update Ecore metamodel is available at `org.omg.sysml/model/SysML.ecore` on the ST6RI-351 branch, if you want to take a look.


rbarbau (approved):

ujhelyiz (approved):
This change looks very nice to me; I especially like that the code became much simpler.

I did not see any specific issue that might result in performance regressions (as you have mentioned during yesterdays telco), and when running the interactive profiler I did not see any regressions between the current master and the PR branch (loading the library got significantly faster - 4 seconds vs 4.7 seconds; on the other hand parsing the simple model files was mostly the same even if somewhat slower, e.g. 8-Requirements.sysml file 0.9 seconds vs 1.1 seconds). There might be some non-trivial performance regression, but I have no idea what might have caused it.

himi (approved):
I found this change caused NullPointerException in PlantUML visualization and I fixed it.

ivan-gomes (approved):
lgtm

PR #218, opened 2021-03-20
ST6RI-341: Help option for the %viz command (PlantUML)

himi:
It's a corner-cutting solution but better than nothing ;)


seidewitz (approved):
I updated the help string with some simple formatting using tabs. This could also be done in a more sophisticated way using `String.format`, but that seemed unnecessary at this point. If you are OK with this, everything else looks good.

himi:
Thank you for your changes and comments, Ed.  In the future, that help string should be constructed from the actual configuration.  So could you please merge it?

seidewitz:
We should probably provide help for all the magics eventually.

PR #221, opened 2021-04-06
ST6RI-312 (and ST6RI-330) Enhanced Q&U library with vector and tensor quantities

hpdekoning:
Q&U library update with support for vector and tensor quantities. This pull request also resolves the bugs reported in ST6RI-330 concerning duplicate unit types defined in the ISQ packages.

The generated quantity attributes still contain subsetting the xxxQuantities features:
```
    attribute speedQuantities: SpeedValue[*] nonunique :> scalarQuantities;
    attribute speed: SpeedValue[1] :> speedQuantities;
```
This will be replaced shortly with the original pattern, i.e.:
```
    attribute speed: SpeedValue[1] :> scalarQuantities;
```

For the rest the pull request should be ready for review.


hpdekoning:
Note: For now, for vector quantities only Cartesian 3D coordinate systems are generated. Other regularly used coordinate systems (2D/3D Cylindrical, Spherical, and Geo Lat/Long) can be added in a next Sprint.

hpdekoning:
Added commit 27f1b770724840776931bf04483a5a5646906b85 to make a final correction: 

Removed intermediate xxxQuantities and xxxVectors attributes from all quantities declarations.
All quantities now subset directly scalarQuantities or vectorQuantities or tensorQuantities.

seidewitz (changes requested):
1. Minor: The documentation comment for `MeasurementUnit` has a note that says "MeasurementUnit directly specializes MeasurementReference". "MeasurementReference" should be "ScalarMeasurementReference".
2. Minor: `QuantityDimension` is defined twice, once in `Quantities` and once in `UnitsAndScales`. I think it is the latter one that is used.
3. Major: The use of arithmetic operators on units is now generating warning messages. This is because the functions for these operators are from `ScalarFunctions` whose arguments typed by `ScalarValue`. But `MeasurementUnit` now descends from `Array` which is _not_ a `ScalarValue`. So there is a typing problem. This needs to be resolved before we can release. We may need to have a quick Track 4 discussion about it this week.

seidewitz:
I have pushed a commit with what seems to me is the easiest way to fix issue 3 from my review. I created a DataFunctions` package that generalizes all the ScalarFunctions to operator on any DataValue, rather than just on ScalarValues.

Another option was to make NumericalValue not be a subtype of ScalarValue and then define a specialization of Array that was also a NumericalValue. This could have been used as the base for TensorMeasurementValue and TensorMeasurementUnit allowing them to be used with NumericalFunctions. However, that would leave some ScalarValues (the Numbers) as NumericalValues, while others were note, which seemed unnecessarily complicated. And String is a non-NumericalValue that currently uses `+` for concatenation. So I went with the DataFunctions approach.

This is still open for further discussion, though not for long.

Note that there are also still type-checking warnings in the SI and USCustomaryUnits packages that seem to be due to changes that are not directly related to the tensor and vector update. (Plus there is a warning in UnitsAndScales, but that looks like its one I may need to address.)

hpdekoning:
Thanks! Ah, I did not catch the type mismatch issue caused by MeasurementUnit no longer being a ScalarValue. I will have a look later this evening.

seidewitz:
I have fixed the remaining type-check warnings in the Quantities and Units library. Most of these were caused by two issues in changes made elsewhere during this sprint (which I merged into this branch), so it was good to find these.

I have also fixed the type from item 1 in my review. I have not fixed item 2, but there is another, more critical issue around QuantityDimension.

The new ISQBase package now introduces a set of base quantities to use in QuantityDimensions. ISQBase is then publicly imported into ISQ. However, the problem is that, if ISQ and SI are then both imported into a user model, the names for the base quantities `L` and `N` in ISQBase conflict with the SI units for litre and newton. This means that these need to be qualified, which will be counter-intuitive for users, since most won't be dealing with the base quantity kind features at all.

Actually, I thought we had agreed to not include QuantityDimensions with the resolution of this issue, because it was not necessary for introducing vectors and tensors. (The reason for not including too much in an increment is exactly to avoid having to deal with things like this late in the game...)

hpdekoning:
Indeed I missed the conflicts on L and N, should have found those, sorry about that. The reason to still introduce QuantityDimension was different than what I used QuantityDimension originally for. I needed to find a general way to enable verifying that a compatible set of unitPowerFactors is defined on each DerivedUnit without hardcoding the ISQ base quantities into a specialization like SIDerivedUnit. I could not find a way to do this 'statically'. I realized that the earlier attempt I had with creating private attributes upf1, upf2, etc for each of the individuals UnitPowerFactors and then binding unitPowerFactors = {upf1, upf2, ...} was cumbersome and did not enforce correct typing.

Is the pull request acceptable now, or should I still refactor the conflicting use of the base quantities?
I tried to load the latest commit in my Eclipse environment but it won't. I think I need to upgrade to catch up with the master branch.

seidewitz:
We need to fix the conflicting base quantity names. I would suggest simply moving their definition into the body of the `isq` feature. This makes some sense, since they are specifically ISQ base units. Since `isq` is itself at package level, they would still be universally accessible through paths like `isq::L`, etc. And, if a model is not importing the contents of `SI`, it can include `import ISQBase::isq::*;` so the base quantity names are available without qualification.

I already tried out the above approach, which only required minimal manual modifications. If you agree, I can check that in, and you can change your generator later.

Also, we need to delete one of the definitions of `QuantityDimension`. Would it be right to delete the one in `Quantities` and keep the one in `UnitsAndScales`?

hpdekoning:
OK, yes I agree with moving the base quantity names into the body of `isq`.
For `QuantityDimension` I think rather the one in `UnitsAndScales` should be deleted, since `QuantityDimension` is about quantities and also the declaration of `QuantityPowerFactor` is above `QuantityDimension` in `Quantities.sysml`. I think I forgot to remove it from `UnitsAndScales` after copying it.

I also noticed another left-over in lines 53-54 of Quantities.sysml:
```
	// TODO: Remove TEMPORARY HACK	
	alias quantities as quantity; 
```
That should have been removed as well. I am almost 100% sure that there is no more reference to `quantity`.
Thanks!

seidewitz (approved):

seidewitz:
OK, I made the final changes as you suggested. I think we are ready to go now.

PR #225, opened 2021-04-16
ST6RI-375 Importing a member using an alias does not work correctly

seidewitz:
It was noticed some time ago that importing an element using an alias name didn't work right. In fact, we had three KerML Xpect tests that failed, which we have had sequestered so they didn't run as part of the regular suite. Somehow, though, we never recorded this as a bug in Jira.

The issue has come up again, though, because of the increasing extent of the use of aliasing in the Quantities and Units library. So, I have create Jira issue [ST6RI-375](https://openmbee.atlassian.net/browse/ST6RI-375) for it, and this pull request proposes a resolution, to be included in the 2021-03a release with other bug fixes.

The intent is that an import such as `import X::Y::Z;` be effectively equivalent equivalent to `import X::Y::Z as Z;`, whether `Z` is the ownership name or an alias name of the element being imported. However, this is not straightforward to implement, since, once the qualified name is resolved, it is no longer available in the abstract syntax tree. That means that the only place that the last segment of the qualified name can be extracted is during name resolution.

This pull request contains a change to `KerMLScope.resolveInScope` to check whether the qualified name currently being resolved is in the context of a Membership that is parsed from an `import` or `alias` such as above. If so, the `memberName` of the Membership is set to the last segment of the qualified name. This is rather a hack, since it happens as a side-effect of the name resolution process, but it seems to be the best we can do right now without a metamodel change.

With this change, all three of the formerly failing Xpect tests pass.


seidewitz:
It turns out that this fixes the bug, but causes other sporadic problems. I now think that this cannot be resolved without a change to the metamodel. So I am going to withdraw this PR and put the issue back on the backlog.

PR #228, opened 2021-05-06
ST6RI-313 Q&U Library - ISO 80000-12 Condensed matter physics

hpdekoning:
ISO 80000 part 12 was added, and a number of documentation improvements as well. ST6RI-313 is built on top of ST6RI-196 that was merged in order to enable using the new sequence initializer syntax (parentheses instead of curly brackets). The ISO/IEC 80000 library package generator has been upgraded accordingly.

ST6RI-314 (Q&U Library - IEC 80000-13 Information science and technology) will follow shortly, and will be built upon ST6RI-313.


hpdekoning:
I see there are still a few places where parentheses are put around a 'single element sequence'. I will update the generator for ST6RI-314 to avoid those.

seidewitz (approved):

PR #229, opened 2021-05-07
ST6RI-314 Q&U Library - IEC 80000-13 Information science and technology

hpdekoning:
IEC 80000-13 added, plus some minor corrections in documentation and handling quantities and units of dimension one.


hpdekoning:
I think we get the same failure in CI as with #228.

seidewitz:
The build completes successfully after merging with the latest from ST6RI-196.

PR #230, opened 2021-05-10
ST6RI-362 Stakeholders, concerns and renderings

seidewitz:
This pull request includes the following:

1. _Stakeholders._ Notation for StakeholderDefinition and StakeholderUsage using keywords **`stakeholder def`** and **`stakeholder`** with corresponding base elements `Requirements::Stakeholder` and `Requirements::stakeholders`.
2. _Concerns._ Notation for
   - ConcernDefinition and ConcernUsage using keywords **`concern def`** and **`concern`** with corresponding base elements `Requirements::ConcernCheck` and `Requirements::concernChecks`. 
   - Referencing an affected stakeholder from a concern using the keyword **`affect`** (this replaces the earlier proposal of **`involve`**).
   - Referencing a framed concern from a requirement using the keyword **`frame`** (this replaces the earlier proposal of **`address`**). Note that viewpoints are kinds of requirements.
3. _View Renderings._ 
   - Notation for referencing a rendering for a view using the keyword **`render`**. 
   - New library elements for standard kinds of renderings: `asTextualNotation`, `asTreeDiagram`, `asInterconnectionDiagram`, `asElementTable`.
4. _Reflective Abstract Syntax._
   - New reflective abstract syntax models for KerML (Kernel Library model `KerML.kerml`) and SysML (Systems Library model `SysML.sysml`). These include _only metaclasses_ at this time, with no metaproperties.
   - Implicit annotation of any element with its abstract syntax metaclass, which may be tested in a filter expression.
3. _View Magic Command._ New `%view` Jupyter magic command that allows a view usage to be rendered. (The functionality of the `%viz` command is unchanged.)
4. _Magic Command Help._ Help messages for all Jupyter magic commands.


himi:
@seidewitz can you give me a pointer to the sample sysml file for view and rendering?  I'm now reading 11-View and Viewpoint validation case but it's not a good example to look into %view and rendering.

himi:
BTW, can I add the following entry to .gitignore? Or how do you cope with annoying garbage made by Finder?
```
# MacOS Finder
.DS_Store
```

seidewitz:
@himi Actually, if you change the render for the `system structure generation` view to be either `asTreeDiagram` or `asInterconnectionDiagram`, then it provides a reasonable example for `%view`. Or you can take the render off the view usage and use the `--render` option on the `%view` command (with the same values as for the `%viz --view` option).

I should also update the `11b-Safety and Security Feature Views` model to use the new library rendering elements, and then you can try the views in that, too. I don't really have much else.

seidewitz:
On `.DS_Store`, I thought this was already in `.gitignore`, because they are all ignored in my workspace. On further research, though, I found that the exclusion was configured in my local `~/.gitignore_global` file, but had never been added to the project `.gitignore`. So, by all means, feel free to add it to the project file in the repository!

himi:
@seidewitz, thank you.  I could successfully run it and it seems to work properly.  So, I approve this pull request, but I'm still wondering how you make directions in rendering definitions.  Currently, SysMLInteractive checks the name of rendering to choose which style is shown, however, I think you intend to add more formal specifications in the definitions.  I think we do not have sufficient time to discuss on it, and if it's true, let's defer the discussion.  So, I'm not so inclined to define columnView in asElementTable rendering.

org.omg.sysml.jupyter.kernel/src/main/java/org/omg/sysml/jupyter/kernel/magic/View.java
```diff
@@ -0,0 +1,68 @@
+/*
+ * SysML 2 Pilot Implementation
+ * Copyright (c) 2020 Mgnite Inc.
+ *
```
himi, line 4: Maybe you should add your copyright notice.

himi (approved):

himi:
> On `.DS_Store`, I thought this was already in `.gitignore`, because they are all ignored in my workspace. On further research, though, I found that the exclusion was configured in my local `~/.gitignore_global` file, but had never been added to the project `.gitignore`. So, by all means, feel free to add it to the project file in the repository!

Thank you.   Since my home directory is shared with other Linux servers, adding it to .gitignore is better.  So I added it.

org.omg.sysml.jupyter.kernel/src/main/java/org/omg/sysml/jupyter/kernel/magic/View.java
seidewitz: Oops. I fixed it.

seidewitz:
> I think %view and %viz magick should be integrated in the future, and --view option of %viz is now confusing. When view is given to %viz, it should visualize by following to the view specification.

I agree that the `--view` option on `%viz` is now confusing. It would be better if it was `--render`, like on `%view`, but I decided not to change it right now, since people are used to the option on `%viz`.

I do think the commands are different, though. The `%viz` command takes the elements to be visualized as arguments. A ViewUsage, however, already specifies the elements to be view through its Expose relationship. So it would not really make sense to specify a ViewUsage in the `--view` options on `%viz` and then also give the elements to be viewed as arguments. Both commands are useful, but they are different.

It is still worth more discussion in the future, though. The view model integration is very preliminary -- but more than I expected to get done on this sprint!

PR #232, opened 2021-05-17
ST6RI-380: Support MacOSX "aarch64" and Linux "arm" environment for Maven Build

himi:
This change should make the maven build work on MacOS "aarch64" (a.k.a. Apple Silicon) or Linux "arm" environment as well.  This change will be useful for OpenJDK 11 environment on Travis.  I think it should work on Linux aarch64 environment but I could not test it now (since my Linux server environment is still armv7l)

The most controversial point is that it uses Eclipse 2021-06, which has not yet been officially released, but so far only this version has the SWT bundle for MacOS aarch64.  If you change it to 2021-03, the build will fail on MacOS aarch64.

I also need to update Tycho to 2.3.0, node.js to 16.1.0, and frontend-maven-plugin to 1.12.0 to support MacOS aarch64.

For Travis OpenJDK11, I needed to set up -Dhttps.protocols=TLSv1.2 JVM flag.  Without that, Maven failed to connect to github.com due to the SSL authentication error.


.travis.yml
```diff
@@ -1,6 +1,9 @@
 language: java
 jdk:
-  - openjdk8
+  - openjdk11
```
ujhelyiz, line 3: This a massive change here! If we compile in Tycho with Java 11, at the very least, there may be issues with some Java standard library entries not being matched (e.g. if somebody uses a method that is only available in Java 11, it may compile but fail with runtime exceptions when using Java 8); in the worst, even the compiled bytecode may not work with Java 8. This can be worked around by the correct maven toolchain setup, but I don't recall that part being added to the build configuration.

ujhelyiz (changes requested):
Unless we are fine with requiring Java 11 as a minimum dependency, this change is very risky. On the other hand, if we want to require Java 11, I suggest updating the manifest files as well to explicitly require Java 11 - e.g., see https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/blob/master/org.omg.kerml.xtext.ui/META-INF/MANIFEST.MF#L26.

seidewitz:
We have a separate task on this sprint to move to Eclipse 2021-03 and Java 11 ([ST6Ri-383](https://openmbee.atlassian.net/browse/ST6RI-383). We will _not_ be moving to the pre-release Eclipse 2021-06 at this time, at least not as part of our official development configuration. Is the update in this pull-request compatible with Eclipse 2021-03, other than it won't actually work for MacOS aarch64?

.travis.yml
himi: This is just for build environment on Travis and it should not impact anything on build target.  OpenJDK11 is required by Tycho 2.3.0, and I do not think the latest Eclipse environment supports any JDK version older than 11.  Build source and target are controlled by maven the following configurations of pom.xml as
` <maven.compiler.target>1.8</maven.compiler.target>`
`<maven.compiler.source>1.8</maven.compiler.source>`
These are not change.

himi: Regarding MANIFEST.MF,  each bundle should properly specify which runtime environment is supported.  It should not be updated by the change of build environment because  we assume or trust JRE keeps backward compatibility.

himi:
> 
> 
> We have a separate task on this sprint to move to Eclipse 2021-03 and Java 11 ([ST6RI-383](https://openmbee.atlassian.net/browse/ST6RI-383)). We will _not_ be moving to the pre-release Eclipse 2021-06 at this time, at least not as part of our official development configuration. Is the update in this pull-request compatible with Eclipse 2021-03, other than it won't actually work for MacOS aarch64?

If you change the repository to 2021-03, build does not work on MacOS aarch64, but it works on other environment including Linux "arm" environment.

himi:
@ivan-gomes could you please review on the impact of the upgrade of Node.js to 16.1.0?

himi:
I quickly tested it with 2021-03
- It works on Windows x64
- It works on Linux arm 
- It works with .m2 cache on MacOS aarch64 ;).  It failed after clearing the cache.

.travis.yml
ujhelyiz: This setting is used just for the build environment and my issue is that it is not enough. I think (but I am not sure so it should be tested), the mentioned settings would cause the compiler to emit Java 8 specific byte code, so for now, it should work.

However, considering both the development environments (required by current Eclipse versions) and the build system uses Java 11 while we state that the built bundles work in Java 8, who will catch issues as somebody using newer standard library code (e.g. [List.of](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/List.html#of(E)) that got introduced in Java 9), it would compile fine but if somebody tried to use the bundles in a Java 8 environment the built-in checks in Eclipse would not flag it. If we are going to support only Java 11, why not state it explicitly in the code as well?

himi: Your concern that the runtime may mismatch with what we assume should be handled with `Bundle-RequiredExecutionEnvironment` of OSGi configuration in MANIFEST.MF, as you probably already noticed.

Each bundle we develop can still support Java8 by specifying JavaSE-1.8 in Bundle-RequiredExecutionEnvironment so that bundle users may use it with JRE8.   Practically it will be harder and harder because such bundle will depend on and be tested with the latest Eclipse bundles that support only Java11 or later.   However, that will not be a good reason to specify Java11 in Bundle-RequiredExecutionEnvironment because it's better to allow developers to port or use such bundles in different environment and it would be simply harmful in view of portability to specify Java11 if they do not use Java11 specific features.

Anyway, that is independent of the build environment we use.   We expect JDK keeps backward compatibility with the compiled bytecode.  So when the development environment, in this case, Tycho 2.3.0, requires JDK11 feature, we need to use JDK11 anyway.

Of course what runtime environment we officially support is another story.  It involves many other factors such as testing, deployment, and support.  We should discuss it separately.

ujhelyiz: If this requirement specification will be done in relation to [ST6RI-383](https://openmbee.atlassian.net/browse/ST6RI-383) I am fine with postponing this until that point, but I really believe it should be done as soon as possible.

ujhelyiz (commented):
As written on a limited discussion, I don't think the current implementation is in a stable form (when building with Java 11, there might be issues with compatibility with Java 8). However, considering in [ST6RI-383](https://openmbee.atlassian.net/browse/ST6RI-383) we want to change the required implementation to Java 11, this would need only a few dependency specification updates, so I am fine with this change being merged now.

.travis.yml
himi: @ujhelyiz, as I said, I disagree to change the bundles' Bundle-RequiredExecutionEnvironment for this reasons.

ujhelyiz: I understand your reasoning, as you are saying you want to build system to keep the built java classes compatible with Java 8, so not changing the BREE values makes sense.

However, my point is that the build system currently does not guarantee that everything will work in Java 8 when built on Java 11 as classes and methods of the JDK running the build might be resolved in a way that is not compatible with Java 8 (I am speaking from experience, such issues are really hard to find out). However, given that it is planned to update minimum requirement in a different ticket in the not too distant future, it might make sense to ignore this issue for now.

seidewitz:
I am going to hold off merging this branch until later in the sprint, at least until after ST6RI-383 and ST6RI-384. It makes sense to complete the migration to Java 11 first. And, just in case there are other issues with this PR, I want to make sure the update for the 2021-05 metamodel baseline is done first.

.travis.yml
himi: Indeed, I agree on your point.  Also in my experience, changing runtime environment made unexpected problems and it's really hard to find.  For us, I consumed almost one week about the conflicts of OSGi and runtime environment and adding only one single option of target environment solved the problems ;)  Lessons learned is we need ample time after the migration of runtime and build environment.

pom.xml
```diff
@@ -21,8 +21,8 @@
     <build-helper-maven-plugin.version>3.2.0</build-helper-maven-plugin.version>
     <maven-clean-plugin.version>2.5</maven-clean-plugin.version>
     <maven-assembly-plugin.version>3.3.0</maven-assembly-plugin.version>
-    <frontend-maven-plugin.version>1.9.1</frontend-maven-plugin.version>
-    <node.version>14.14.0</node.version>
+    <frontend-maven-plugin.version>1.12.0</frontend-maven-plugin.version>
+    <node.version>16.1.0</node.version>
```
ivan-gomes, line 25: I believe 16.2.0 is out now. May be good to update it while we're here.

ivan-gomes: It may be worth noting that Node v16 will be the **next** LTS ([ref](https://nodejs.org/en/about/releases/)) starting in October, so it's the best version to jump to since we wish to get a compatibility feature not offered in the current LTS.

ivan-gomes (approved):
Confirmed (without cache) compatibility with Linux x86_64.

pom.xml
himi: @ivan-gomes, thank you. I updated node.js to 16.2.0

seidewitz (approved):

PR #234, opened 2021-05-26
fix small typo

Rots:


seidewitz:
Thank you for noticing this and suggesting the patch. However, the `SysML.ecore` file is actually generated from the `SysML.uml` model, which is, in turn, an export for our normative MOF model for the abstract syntax. We never modify the Ecore file by hand.

However, I will correct this typo in our abstract syntax model, so the correction will be included in the Ecore file for our next release (and in the similarly generated documentation in the draft KerML spec document).

PR #237, opened 2021-06-02
ST6RI-384 Update implementation for 2021-05 metamodel baseline

seidewitz:
The following metaclasses have been added:

- OccurrenceDefinition
- OccurrenceUsage
- PortioningFeature
- PortionKind (enumeration)
- EventOccurrenceUsage _[added since original baseline]_

**Note:** The metaclass LifeClass was added to the normative abstract syntax in the 2021-05 baseline, but it already existed previously in the implementation.

The following metaclasses have been removed:

- IndividualDefinition
- IndividualUsage
- TimeSliceFeature
- SnapshotFeature

Other changes:

1. _Occurrences._ A basic textual syntax has been added for **`occurrence def`** and **`occurrence`**.  No syntax has been added for "events" or "messages" (that will be handled as part of ST6RI-385).
2. _Individuals._ The notations for **`individual def`**, **`individual`**, **`timeslice`** and **`snapshot`** are unchanged. However, it they now parse to OccurrenceDefinition and OccurrenceUsage (possibly with a PortioningFeature), rather than to IndividualDefinition and IndividualUsage. No syntax has been added for declaring more specific kinds of individuals like **`individual part`**, etc. (that will be handled as part of ST6RI-385).
3. _PlantUML._ Minimal updates have been made to PlantUML visualization to support the existing graphical notation for individuals using the new metamodel.
4. _ST6RI-368._ This bug has been resolved by renaming `Occurrences::successions` to `Occurrences::happensBeforeLinks` in the Kernel model library. `Transfers::flows` has also been renamed to `Transfers::transferBefores`.
5. _ST6RI-378._ This bug has been resolved by adding the base association `Occurrences::HappensLink` in the Kernel model library.

PR #239, opened 2021-06-08
ST6RI-385 Concrete syntax for occurrence definitions and usages

seidewitz:
This pull request adds SysML concrete syntax related to occurrences.

1. _Occurrences._ A basic textual syntax for **`occurrence def`** and **`occurrence`** was previously added as part of ST6RI-384 (PR #237).
2. _Individuals._ The syntax for individuals has been expanded.
   - Individual definitions and usages can now be made on all kinds of occurrences, i.e., **`individual part def`**/**`individual part`**, **`individual action def`**/**`individual action`**, etc. 
   - The notations **`individual def`** and **`individual`** default to **`individual occurrence def`** and **`individual occurrence`**. 
   - Every individual usage of any kind must have exactly one type that is an individual definition. However, individual usages may now also have additional types that are not individual definitions.
3. _Time slices and snapshots._ The time slice and snapshot notation has been updated.
   - The **`timeslice`** and **`snapshot`** notations are no longer restricted to individuals, allowing any kind of occurrence type. The kind of occurrence can be explicitly given in the declaration, e.g., **`timeslice part`**, **`snapshot action`**, etc.
   - A **`timeslice`** or **`snapshot`** usage without an explicit type that is nested in the body of occurrence definition or usage (of any kind) is implicitly a time slice or snapshot of its containing occurrence. 
   - A **`timeslice`** or **`snapshot`** usage with a type that is an individual definition implicitly represents a time slice or snapshot of the defined individual. To make a **`timeslice`** or **`snapshot`** declaration explicitly an individual usage, the notations **`individual timeslice`** and **`individual snapshot`** can be used.
4. _Event occurrences._ A notation has been added for event occurrences.
   - The notation **`event occurrence`**_`name`_`;` represents an identified suboccurrence of a containing occurrence. That is, it is equivalent to **`ref occurrence`**_`name`_`:> suboccurrences;`. 
   - The notation **`event`**_`qualifedName`_`;` declares an event occurrence by reference to an occurrence usage owned elsewhere. It is equivalent to **`ref occurrence`**`:>`_`qualifiedName`_`:> suboccurrences;`. 
   - The **`event occurrence`** notation can be considered a generalization of **`perform action`** and **`exhibit state`**. However, unlike **`perform`** and **`exhibit`**, the referential **`event`** notation does _not_ take its effective name from the occurrence usage it references. This is because it is expected that **`event`** will commonly be used to reference the `source` and `target` of messages (see below), which would result in name conflicts between multiple such events in one namespace.
5. _Message._ A notation has been added for _messages_, which are item flows without explicitly specified `sourceOutput` and `targetInput` features. 
    - Messages are notated as special kinds of connectors: **`message`**_`name`_**`of`**_`FlowType`_**`from`**_`source`_**`to`**_`target`_`;`. The _`name`_ and _`FlowType`_ are optional. 
    - As an item flow, the implicit type of a message is the Kernel library type `Transfer`. A message can also be explicitly typed with any specialization of `Transfer`.
    - The ends of a `Transfer` have been renamed from `transferSource`/`transferTarget` to `source`/`target`. This means that the ends of a named message can be referenced as `messageName::source` and `messageName::target` (e.g., when used with the **`event`** notation).


seidewitz:
@himi
There has been no further update made to the PlantUML visualization over what was in ST6RI-384. This means that base OccurrenceDefinitions and OccurrenceUsages with `isIndividual = true` or a `portionKind` are rendered using the `<<individual def>>`, `<<individual>>`, `<<timeslice>>` or `<<snapshot>>` keywords. Any specialized kind of occurrence definition or usage that uses the default keyword rendering act similarly (for example, the textual notation **`individual part`** renders graphically as simply `<<individual>>`). However, kinds of definitions or usages with special keyword rendering ignore `isIndividual` or `portionKind` (for example, **`individual analysis`** renders as `<<analysis>>`).

Do you want to make an update to the visualization on this branch, or do you want to defer that to another issue on the next sprint?

himi:
I'd like to work on that.  By when do you want to finish this PR?

seidewitz:
> I'd like to work on that. By when do you want to finish this PR?

No later than the end of this week (11 June). 

Alternatively, we could finish this PR sooner and put a new task on the current sprint for your work. You could perhaps have until 14 June to complete that before it would need to be rolled over to the next sprint.

himi:
I think we should consider `<<timeslice>>` and `<<snapshot>>` as well as `<<individual>>` to add keywords because for example we can say `individual snapshot part ... `.   So I would need to rewrite the entire part of caseOccurenceUsage().  I'll work on this week because deferring it does not save my work.

himi:
By considering the combination of `individual` and `snapshot` or `timeslice`, I do not think "non-individual" snapshot and timeslice do not make sense.   But I could write either `individual snapshot` or `snapshot`.  Do those have differences between these?  If not, ou.isIndividual() should be true if ou.getPortionKind() is TIMESLICE or SNAPSHOT.

seidewitz:
The 2021-05 metamodel baseline, specifically does NOT require that snapshots and time slices be only for individuals. This was never an issue in the kernel semantics, but only a restriction due to how snapshot and times slice modeling was previously incorporated into individual modeling in the SysML syntax.

After Track 2 review, there was a specific request to allow more general snapshot and timeslice modeling. For example, you can model the lifetimes of Persons in general as having snapshots for birth and death and time slices for childhood and adulthood. These would be different of each Person and an individual Person would have corresponding individual snapshots and time slices.

So, as noted in the PR description, isIndividual and portionKind are independent properties.

Sent from my iPad

On Jun 9, 2021, at 1:04 AM, Hisashi Miyashita ***@***.***> wrote:

﻿

By considering the combination of individual and snapshot or timeslice, I do not think "non-individual" snapshot and timeslice do not make sense. But I could write either individual snapshot or snapshot. Do those have differences between these? If not, ou.isIndividual() should be true if ou.getPortionKind() is TIMESLICE or SNAPSHOT.

himi:
@seidewitz, thank you for the answer.  I almost finished the implementation.  I'll commit it this night.

himi (approved):
I could test individual and event and write what I need.

oysteiha (approved):
This is not really a format where I can give sensible feedback, but I approve the general style presented in Track 4 on Thursday 10. June, in particular the more concise message-oriented style. For the record, I would appreciate a straight forward way to give arguments to the message payload as mentioned yesterday.

PR #241, opened 2021-06-09
ST6RI-389 Update XMI export to use UUIDs for XMI IDs

seidewitz:
This PR updates the `KerML2XMI` and `SysML2XMI` utilities to generate XMI that uses the UUID `identifier` of an Element as its XMI ID. These IDs are then used for cross-references, rather than the default Xpaths.

It also includes a fix for the bug reported in [ST6RI-402 Member names are sometimes not written to XMI](https://openmbee.atlassian.net/browse/ST6RI-402).


wpiers (approved):
Implementation looks fine to me, I also tested the bridge and it works well.

PR #242, opened 2021-06-10
ST6RI-388 Boolean expression operator aliases

seidewitz:
This PR includes the following changes related to Boolean expressions.

1. _Exclusive or._ The exclusive-or operator has been changed from `^` to `^^`. This allows `^` to be used as an exponentiation operator. (The existing `**` exponentiation operator has also been retained for now.)
2. _Keyword operators._ Keywords have been introduced as alternatives for certain existing symbolic operators.

Keyword | Symbol
---------|---------
**`and`** | `&&`
**`or`** | `\|\|`
**`xor`** | `^^`
**`not`** | `!`

3. _Implies._ A new `implies` operator has been add for logical implication. Note that this is a conditional operator: the expression _`x`_**`implies`**_`y`_ is equivalent to **`not`**_`x`_**`or`**_`y`_ or `!`_`x`_`||`_`y`_. That is, if _`x`_ evaluates to false, then _`y`_ is not evaluated and the implication results in true.
4. _Precedence._ The precedence of the logical operators has been changed so that **`and`** and `&&` have the same precedence as `&`, and **`or`** and `||` have the same precedence as `|`. The precedence of **`implies`** is lower than the or operators.
5. _Conditional._ A new syntax **`if`**_`test`_`?`_` x`_**`else`**_`y`_ has been introduced as an alternative for the conditional expression _`test`_`?`_`x`_`:`_`y`_.


hpdekoning:
Very nice! I agree with the approach. I do not have the infrastructure ready perform a build and test, but in a quick scan of the changed files I only saw sensible / expected updates.

seidewitz:
@hpdekoning I assume that means you will approve the PR...

himi (approved):
I quickly tested this PR and it worked.  I like `implies` preserves as a distinct operator.

PR #248, opened 2021-06-21
ST6RI-410 Name resolution can fail for PathStepExpression

seidewitz:
Updated `KerMLScopeProvider.featureRefNamespace` to return the `referent` of FeatureReferenceExpression as the relative namespace, rather than the `result`. The `result` should nominally subset the `referent`, giving it implicitly the same typing. But that implicit subsetting is only set when the `result` is transformed, which my not have happened yet during name resolution.


himi (approved):
I confirmed it worked.

PR #249, opened 2021-06-22
ST6RI-411: Dot-notation expression visualization (PlantUML)

himi:
Visualize connectors with dot-notation expressions.
This PR includes a bug fix of interconnection view.

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/SysML2PlantUMLText.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/SysML2PlantUMLText.java
@@ -489,4 +489,14 @@ public class SysML2PlantUMLText {
     Integer getId(Element e) {
         return idMap.getId(e);
     }
+
+    private boolean inherited;
+
+    void setInherited(boolean flag) {
+        this.inherited = flag;
+    }
+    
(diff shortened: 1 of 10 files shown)


seidewitz (changes requested):
Consider the following model:
```
package ConnectionTest {
	part p {
		part x {
			part x1;
		}
	}
	part def P {
		part y;		
		connect p to y;		
		part p1 :> p;
		connect p1.x to y;		
		connect p1.x.x1 to y;
	}
}
```
The interconnection visualization is:
image
This has the following problems:
1. The second and third connections are shown as being on nest features of `p`, not `p1`.
2. The third connection is shown as connecting to `x` rather than `x1`.

himi:
I have not yet addressed the inherited case but first fix the case of `p1.x.x1`

seidewitz:
I can confirm that the test model I gave earlier does now visualize with `y` connected to `x1`.

If you are going to address the visualization of connections to inherited elements in this PR, please have the update pushed no later than then end of Monday afternoon (28 July). Otherwise, you can include it in ST6RI-412 for 2021-06.

himi:
Done.  It should work with inherited features.

seidewitz (approved):
Looks good!

PR #254, opened 2021-07-16
ST6RI-406 Dot notation for feature chains

seidewitz:
Formerly, dot notation of the form `f_1.f_2. ... .f_n`, for _n > 1_, mapped to a hierarchy of (left associative) PathStepExpressions, one for each dot. This notation could be used in two ways:

1. As an evaluable Expression in its own right, either directly as an owned Expression or as a sub-Expression of some other Expression. In this case, `f_1` can be any Expression for which `f_2` is a Feature accessible from the result. All the other `f_i` must be Feature names.
2. To specify an end feature of a Connector (including ItemFlows). In this case, all the `f_i` must be feature names.

This PR replaces the use of PathStepExpressions in the second case with the new FeatureChaining construct available in the 2021-06 baseline. No change is being made at this time to the use of PathStepExpressions for the first case. This case will be addressed further as part of the implementation of [ST6RI-387 Concrete syntax for assignment and control structures](https://openmbee.atlassian.net/browse/ST6RI-387). Dot notation will also now be usable in a third way:

3. As the target of a Subsetting or Redefinition relationship.

A _chained Feature_ is a Feature that represents the relational composition of a sequence of two more more other _chaining Features_. The chained Feature is related to its chaining Features using FeatureChaining relationships. FeatureChainings are always owned by the (source) chained feature.
* The operation `Feature.getOwnedFeatureChaining` returns a list of the FeatureChainings for a Feature (if any), ordered from first chaining Feature to last. 
* The operation `Feature.getChainingFeature` returns a list of the chaining Features, also ordered from first to last. 
* The `Feature.getFeaturingType` operation has been updated to include all the `featuringTypes` of the first chaining Feature.
* The `Feature.getType` operation has been updated to include all the `types` of the last chaining Feature. 
* The static utility operations `FeatureUtil.getFirstChainingFeatureOf` and `FeatureUtil.getLastChainingFeatureOf` can be used to get the first and last chaining Features of a Feature (they avoid doing the full derivation of the `getChainingFeature` list). They return `null` if the Feature has no FeatureChainings.
* Features that subset or redefine a chained Feature inherit members from the last chaining Feature.

The dot notation is implemented as follows using feature chaining:

1. Notation of the form `f_1.f_2. ... .f_n` is normally parsed to a single Feature that owns FeatureChainings that reference each of the `f_i`, in order (but see the special case for ItemFlows below).
2. A chained Feature that is the target of a Subsetting or Redefinition is parsed as an `ownedRelatedElement` of the relationship, as well as being the `subsettedFeature` or `redefinedFeature`. 
   * The end of a Connector is parsed (as usual) as the `subsettedFeature` of a redefining end Feature of the Connector, so a chained Feature used to specify the end of a Connector is owned by the relevant Subsetting relationship for that end.
   * When used to specify the source or target end of an ItemFlow (other than a message), `f_n` is taken to specify the source input or target output feature, and, so, is not included in the chained Feature for the end. If _n = 2_, then the ItemFlow end is parsed as a reference to the Feature named by `f_1`, without chaining. Otherwise, the end Feature is parsed as the chained Feature given by `f_1.f_2. ... .f_m`, for _m = n - 1_.
3. There is a KerML-specific notation **`feature`**_`memberName`_**`is`**_`f_1.f_2. ... .f_n`_`;` that can be used to give a chained Feature a name. It is parsed to a Membership (_not_ a FeatureMembership) with the given `memberName` and the chained Feature as its `ownedMemberElement`.

### DIFF
--- a/org.omg.kerml.xpect.tests/library/Objects.kerml
+++ b/org.omg.kerml.xpect.tests/library/Objects.kerml
@@ -36,10 +36,7 @@ package Objects {
 	 * BinaryLinkObject is the most general binary association structure, being both a 
 	 * BinaryLink and a LinkObject.
 	 */
-	assoc struct BinaryLinkObject specializes BinaryLink, LinkObject {
-		end feature source: Anything[0..*] nonunique subsets BinaryLink::source;
-	    end feature target: Anything[0..*] nonunique subsets BinaryLink::target;
-	}
+	assoc struct BinaryLinkObject specializes BinaryLink, LinkObject;
 	
 	/**
 	 * objects is a specialization of occurrences restricted to type Object.
--- a/org.omg.sysml.xpect.tests/library.kernel/Objects.kerml
+++ b/org.omg.sysml.xpect.tests/library.kernel/Objects.kerml
@@ -36,10 +36,7 @@ package Objects {
 	 * BinaryLinkObject is the most general binary association structure, being both a 
 	 * BinaryLink and a LinkObject.
 	 */
-	assoc struct BinaryLinkObject specializes BinaryLink, LinkObject {
-		end feature source: Anything[0..*] nonunique subsets BinaryLink::source;
-	    end feature target: Anything[0..*] nonunique subsets BinaryLink::target;
-	}
+	assoc struct BinaryLinkObject specializes BinaryLink, LinkObject;
 	
 	/**
 	 * objects is a specialization of occurrences restricted to type Object.
--- a/sysml.library/Kernel Library/Objects.kerml
+++ b/sysml.library/Kernel Library/Objects.kerml
@@ -36,10 +36,7 @@ package Objects {
 	 * BinaryLinkObject is the most general binary association structure, being both a 
 	 * BinaryLink and a LinkObject.
 	 */
-	assoc struct BinaryLinkObject specializes BinaryLink, LinkObject {
-		end feature source: Anything[0..*] nonunique subsets BinaryLink::source;
-	    end feature target: Anything[0..*] nonunique subsets BinaryLink::target;
-	}
+	assoc struct BinaryLinkObject specializes BinaryLink, LinkObject;
 	
 	/**
 	 * objects is a specialization of occurrences restricted to type Object.


seidewitz:
@conradbock @rbarbau 
FYI

seidewitz:
@conradbock @rbarbau 
The validation checks that have been implemented on this branch are now tight enough to result in warnings about some of the issues that Raphael has noticed with associations in the Kernel library models. So I have made a number of renamings of end features and adjustments of redefinitions (including fixing [ST6RI-333 SelfLink participant feature redefinitions/name clash](https://openmbee.atlassian.net/browse/ST6RI-333)). These can be reviewed in commit d8313fd (plus one additional correction in 1ce9c65.)

himi:
At the latest of this branch, ST6RI-406, FeatureChaining did not appear in the outline view and getOwnedFeatureChaining() returned an empty list. Am I doing something wrong?

seidewitz:
> At the latest of this branch, ST6RI-406, FeatureChaining did not appear in the outline view and getOwnedFeatureChaining() returned an empty list. Am I doing something wrong?

Oops. I had a stack overflow happen because of an edge case in displaying the outline tree. However, the fix I put in for that unintentionally also resulted in FeatureChainings no longer being displayed either. I have now pushed a fix for that.

As far as I can tell, though, `getOwningFeatureChaining` is working for me. Note that feature chains are not being used in expressions yet (PathStepExpressions are still used there), only for as targets of subsettings and redefinitions (which includes how they are used for connector ends).

himi:
Thank you. I thought end feature owns FeatureChaining but actually the feature of the subsetting of the end feature owns FeatrureChaining.  I updated VPath to support that.

seidewitz:
@himi 
>  I updated VPath to support that.

I have found at least one issue with the connector visualization after your update to VPath. However, I would prefer not to deal with it on this branch. Please create a new branch ST6RI-412 for your visualization update, then revert commit 3c0550d on this branch. I would like to get this PR merged and closed.

seidewitz:
> Does the parser add redefinitions when a feature has the same name as the one inherited (for source/target in this case)?

It doesn't normally do that (yet), but it does for association ends (actually based on end ordering, not name right now). So, implicit redefinitions get added for BinaryLink::source and BinaryLink::target, making the subsettings entirely redundant. Now, in the kernel library models, I generally try to show redefinitions explicitly, to make it clearer, so these subsettings were probably intended to be redefinitions. However, it is not actually necessary to declare new ends for BinaryLinkObject at all, since it can just inherit them from BinaryLink (this didn't always work right in the past, but it should work now). So I can just update the declaration of BinaryLinkObject to
```
assoc struct BinaryLinkObject specializes BinaryLink, LinkObject;
```

seidewitz:
@conradbock 

> Logical or would allow both sides of the vertical bar to be true (or != implication)

"x implies y" is logically equivalent to "(not x) or y", so the constraint is logically equivalent to 
```
isMove implies (size(moving) == size(sourceOutputLink))
```

himi:
Sorry I wrongly pushed the version. I'm now fixing.

seidewitz:
@himi
All I expected was for you to create a new branch, cherry-pick your one commit to it, then revert the commit on this branch...

seidewitz:
> Sorry I wrongly pushed the revert to master. I restored ST6RI-406 and master branches.

Looks OK now. Thanks.

himi:
I'll set ST6RI-412 to 3c0550d. (Git's beauty or beast :)

conradbock:
> [Conrad] Does the parser add redefinitions when a feature has the same name as the one inherited (for source/target in this case)?
> [Ed] It doesn't normally do that (yet), but it does for association ends (actually based on end ordering, not name right now). 

Does the same apply to connectors:
```
connector sourceOutputLink: BinaryLinkObject[1..*] {
  end transferSource subsets source[1];
  end transferPayload subsets item[1..*] subsets transferSource.sourceOutput; }
connector targetInputLink: BinaryLinkObject[1..*] {
  end transferTarget subsets target[1];
  end transferPayload subsets item[1..*] subsets transferTarget.targetInput;}
```
(connector participant features redefine association participant features based on order)?

seidewitz:
> connector participant features redefine association participant features based on order

Yes. Redefinitions of the ends of BinaryLinkObject are added implicitly. Here I didn't put them in explicitly, because the notation was already getting cluttered doing it this way. (Note that the subsetted `source` and `target` here are _not_ the ones from BinaryLinkObject, they are the ones from Transfer. The ones from BinaryLinkObject aren't inherited, because they have been redefined by ends with different names.)

PR #255, opened 2021-07-20
ST6RI-422 Positive and negative invariants

seidewitz:
This PR implements the revisions in the 2021-06 baseline relating to invariants.

**KerML**

1. *Library.* The features `trueEvaluations` and `falseEvaluations` have been added to the Performances library model as subsets of `booleanEvaluations`, which always evaluate to `true` or `false`, respectively.
2. *Abstract Syntax.* The 2021-06 metamodel baseline adds an `isNegated` flag to the Invariant metaclass. The default is `isNegated = false`, in which case the Invariant implicitly subsets `trueEvaluations`. If `isNegated = true`, then the Invariant implicitly subsets `falseEvaluations`. 
2. *Concrete Syntax.* The syntax for Invariants has been extended to allow **`inv true`**`...` and **`inv false`**`...`, for Boolean expressions that are invariantly true or invariantly false, respectively. **`inv true`** is the defaults, corresponding to the current **`inv`** syntax. **`inv false`** corresponds to setting `Invariant::isNegated = true`.
3. *Transformation.* 
   a. *InvariantAdapter.* It is no longer necessary to add an implicit assertion binding connector to each Invariant.
   b. *TypeAdapter.* The `removeUnnecessaryImplicitGeneralTypes` transform has been updated so that it checks each implicit general type against all general types from owned Specializations _and_ all other implicit general types. This makes it easier to add multiple implicit general types to a type, and then clean up any redundancies later. This is used, for example, to allow an Invariant subexpression to have separate subsets to `trueEvaluations` (or `falseEvaluations` if negated) and `subevaluations`, while a regular Expression would only need to subset `subevaluations`

**SysML**

1. *Constraints.*
   a. *Library.* The feature `assertedConstraintChecks` has been added as a subset of `constraintChecks` and `trueExpressions`. Similarly, the feature `negatedConstraintChecks` has been added as a subset of `constraintCheckChecks` and `falseExpressions`. The feature `Items::assertedConstraintChecks` has been renamed to `Items::checkedConstraints`, because the former name was inaccurate.
   b. *Abstract Syntax.* An AssertConstraintUsage is a kind of Invariant. If it has `isNegated = false`, then it implicitly subsets `assertedConstraintChecks`. If it has `isNegated = true`, then it implicitly subsets `negatedConstraintsCheck`.
   c. *Concrete Syntax.* The syntax for AssertConstraintUsages has been extended to also allow **`assert not`**`...`, corresponding to `isNegated = true`.
2. *Requirements.*
    a. *Library.* The feature `satisfiedRequirements` has been added as a subset of `requirementChecks` and `assertedConstraintChecks`. Similarly, the feature `notSatisfiedRequirementChecks` has been added as a subset of `requirementChecks` and `negatedConstraintChecks`.
   b. *Abstract Syntax.* A SatisfyRequirementUsage is a kind of AssertConstraintUsage. If it has `isNegated = false`, then it implicitly subsets `satisfiedRequirementChecks`. If it has `isNegated = true`, then it implicitly subsets `notSatisfiedRequirementChecks`.
   c. *Concrete Syntax.* The syntax for SatisfyRequirementUsage has been extended to also allow **`not satisfy`**`...`, corresponding to `isNegated = true`.


seidewitz:
@himi @santileortiz 
FYI, in case you want to create visualizations of this.

PR #256, opened 2021-07-21
ST6RI-412: Feature chain visualization (PlantUML)

himi:


seidewitz (changes requested):
Initial review seems to indicate that the visualization of the dot notation works as before. However, I noticed a problem that happens even without the dot notation (so it probably was also there in 2021-05, though I hadn't noticed it before).

The model
```
package Q {
  part def A {
  	part f;
  	part g;
  }
  
  part def B {
	part a : A {
	  connect f to g;
	}  
  }	
}
```
is visualized as
image
This is not correct, because it shows the connector as being in part def `A`, when it really is owned by part usage `B::a`. The connection should really be between the inherited features `^f` and `^g` within the box for `B::a`.

himi:
I examined the parsed result of "connect f to g" in your example and it points to A::f and A::g
Screen Shot 2021-07-23 at 5 11 41 PM

Is it the intended parsed result?

seidewitz:
> Is it the intended parsed result?

Yes, it is. `f` and `g` are owned by `A` and inherited by `a`. But the connection usage is owned by `B::a`, _not_ by `A`. Usages of `A` other than `B::a` (and its specializations) will not have this connection.

seidewitz:
If the model is changed to
```
package Q {
  part def A {
  	part f;
  	part g;
  }
  
  part def B {
	part a : A;
 	connect a.f to a.g;
  }	
}
```
then the visualization is essentially correct
image
Well, the connector is really owned by `B`, not `B::a` in this case, but I don't think there is a good way to make this distinction in a graphical visualization. But the diagram above is how the previous version of the model should be visualized, too.

himi:
Hmmm... It's not easy to understand the result.  Ok.  To correctly process connectors, visualizers need to resolve the inherited features.  In this example, `A::f` is actually inherited by the owning part usage, in this example, `a`.  It is possible but I think it looks quite tricky to understand the semantics of the parsed models

In your example, `connect a.f to a.g` works correctly.  If they are parsed to feature chains, I think the meaning of the parsed results looks clearer.

seidewitz:
> It is possible but I think it looks quite tricky to understand the semantics of the parsed models

This is the way inheritance works. This was already the case in SysML v1. The innovation in SysML v2 is that usages also inherit from, and can extend, their types. In SysML v1 you would have to do something like
```
package Q1 {
  part def A {
  	part f;
  	part g;
  }
  
  part def A1 :> A {
  	connect f to g;
  }
  
  part def B {
	part a : A1;
  }	
}
```
But then the connector would appear in an IBD for `B`, inside `a`, but _not_ in an IBD for `A`. So it's not really a new problem.

However, I have confirmed that this problem already existed in 2012-05 and was not introduced by the update for ST6RI-412 (which you probably already knew). So, if you would prefer, I can put in a separate bug report for the problem I have identified, we can complete this PR (and the subsequent ones), and work on the bug resolution after 2012-06.

himi:
I do not have any objection to the expected results of these examples.  I'm worrying that application programs will confuse the parsed results.  They need to check each referred element is an inherited member by tracing all of the specializations of owning the connector.  And if it is not, they must not treat it as an inherited member.

In SysMLv1, there is no name resolution and  application programs do not need to worry about parsed models.   Through the IBD creation, users need to explicitly specify inherited members to be connected or shown.

I think I can deal with this because VPath can traverse and check inheritances, but I think this specification will impose a heavy burden to application programs that need to process connectors.  If referent is an inherited feature, I think parsing it to FeatureChaining would be better to explicitly specify that it is referring an inherited one.

However, I do not strongly stick to it.  If you want to leave it as it is, I'll modify VPath to deal with this.

himi:
Just note that
`package Q {
  part def A {
  	part f;
  	part g;
  }
  
  part def B {
	part a : A {
	  connect a.f to a.g;
	}  
  }	
}` 
is visualized correctly.

seidewitz:
> However, I do not strongly stick to it. If you want to leave it as it is, I'll modify VPath to deal with this.

You do have a point, but I don't believe there is any semantic reason why references to inherited features should map to feature chains. So, the parsing is correct as it is, and I don't think it is worth complicating the parse tree further in this case.

You can updated VPath on this branch, if you can complete the revision by Monday (26 July). Otherwise, we can open a bug issue for the change.

himi:
Also in the following example, we should connect B::f to B::g as well, I think.  Do you agree?  (Notice that the current visualizer connects A::f to A::g)

```
package Q {
  part def A {
  	part f;
  	part g;
  }
  
  part def B :> A {
	  connect f to g;
  }	
}
```

seidewitz:
Yes, that is correct.

himi:
Now it works with
```
package Q {
  part def A {
  	part f;
  	part g;
  }
  
  part def B :> A {
  	  part h;
	  connect f to g;
	  connect g to h;
  }
  part def C {
  	  part h;
  	  connect h to a.f;
  	  part a: A {
  	  	part b {
  	  	  connect f to g;
  	  	}
  	  	connect b to f;
  	  }
  }
}
```
Screen Shot 2021-07-25 at 12 09 20 AM

himi:
I tested this version with various patterns and it works as expected.

seidewitz:
Close. However, the model
```
package Q1 {
  part def A {
  	part f;
  	part g;
  }
  
  part def B :> A;

  part def C {
  	  part a: A {
  	  	part b: B {
  	  	  connection c_b connect f to g;
  	  	}
  	  	connection c_a connect b to f;
  	  }
  }
}
```
is visualized as
image
But connection `c_a` should connect `b` to `a.f`, not `b.f`.

himi:
Thank you.  I fixed the bug.
Screen Shot 2021-07-25 at 3 56 50 PM

seidewitz:
Looks good!

seidewitz (approved):

PR #259, opened 2021-07-23
ST6RI-363 Multiplicity Defaults

seidewitz:
This PR implements 1..1 multiplicity defaults for certain kinds of SysML usages:

1. AttributeUsages
2. ItemUsages, including PartUsages but _not_ ConnectionUsages
3. PortUsages

The default is added only if the following conditions are met:

1. The usage is owned by a type (i.e., a definition or usage, not a package).
2. The usage does _not_ have any _explicit_ owned Subsettings/Redefinitions.

If a default is not added, then the multiplicity is inherited from subsetted features, as previously (which means, for the case of implicit subsetting of a library element, the default is then effectively 0..*).

The default multiplicities are implemented as follows.

1. The KerML syntax has been extended to allow the declaration of named multiplicities, both in packages and in types.
2. Named multiplicities are added to `Base` package in the Kernel Library for `exactlyOne`, `zeroOrOne`, `zeroOrMore` and `oneOrMore`.
3. When one of the three kinds of usages identified above meets both of the stated conditions, a transformation adds a Multiplicity (_not_ a MultiplicityRange) as an owned member.
4. The multiplicity then gets an implicit subsetting to the `Base::exactlyOne` multiplicity.

**Note:** While not directly related to the new default multiplicities, the 0..1 multiplicity previously added to a LifeClass has been updated so that now has the same form as for a default multiplicity above (i.e., an owned Multiplicity that is not a MultiplicityRange), but subsetting `zeroOrOne` rather than `exactlyOne`.


seidewitz:
@hpdekoning 
Please review the updates the Quantities and Units models in commit 526c315.

seidewitz:
@himi
You may want to consider how these default multiplicities should be visualized, since they don't actually appear in the text. There is a utility method `FeatureUtil.getMultiplicityRangeFor` that will follow the subsetting of a default Multiplicity to get the effective MultiplicityRange.

@santileortiz 
Since implicit generalizations are not being saved to the repository, a default multiplicity, when read from the repository, will appear as an owned Multiplicity member that is not a MultiplicityRange, with no further relationship to anything else. Any such multiplicity on a Usage element should be interpreted as representing a default multiplicity of 1..1.

seidewitz:
@Weilkiti @YvesTLS 
In XMI generated without implicit generalizations, a default multiplicity will appear something like the following:
```
<ownedRelatedElement xsi:type="sysml:AttributeUsage" ... >
    <ownedRelationship xsi:type="sysml:Membership" ... >
        <ownedRelatedElement xsi:type="sysml:Multiplicity" ... >
    </ownedRelationship>
</ownedRelatedElement>
```
Once 2021-06 is released, this is how you should map the default multiplicity for a SysML v1 value property (mapped to an AttributeUsage), part property (mapped to a PartUsage) or reference property (mapped to a referential PartUsage).

himi:
@seidewitz  It sounds useful and it would be easy to support.  Do you want to include the update to the PlantUML visualizer in this PR?

seidewitz:
> @seidewitz It sounds useful and it would be easy to support. Do you want to include the update to the PlantUML visualizer in this PR?

If it is easy, it may be worth including in this PR. But lets get the other outstanding PRs for visualization reviewed and approved first.

seidewitz:
@himi
If you would like to update the visualization for default multiplicities on this branch, please do so.

himi:
Done.  Just made it use FeatureUtil.getMultiplicityRangeOf().

PR #251, opened 2021-07-08
ST6RI-403 Update implementation for 2021-06 metamodel baseline

seidewitz:
This PR implements the 2021-06 baseline metamodel, without updating the surface notation implementation for all new features (which will be done later). It also resolves ST6RI-375 (Importing a member using an alias does not work correctly) and ST6RI-392 (Directed features are not treated as parameters for requirements).

**Abstract Syntax Update**

_Metaclasses Added_

- ActorMembership
- AssignmentActionUsage
- BindingConnectorAsUsage
- ConnectorAsUsage _(abstract)_
- Disjoining
- FeatureChaining
- FlowConnectionUsage
- ForLoopActionUsage
- FramedConcernMembership _(renamed from AddressedConcernMembership)_
- IfActionUsage
- IncludeUseCaseUsage
- LiteralInfinity _(renamed from LiteralUnbounded)_
- LiteralRational _(renamed from LiteralReal)_
- LoopActionUsage _(abstract)_
- Specialization _(renamed from Generalization)_
- StakeholderMembership
- Subclassification _(renamed from Superclassing)_
- SuccessionAsUsage
- SuccessionFlowConnectionUsage
- UseCaseDefinition
- UseCaseUsage
- WhileLoopActionUsage

_Metaclasses Removed_

- AddressedConcernMembership _(renamed to FramedConcernMembership)_
- Generalization _(renamed to Specialization)_
- LiteralUnbounded _(renamed to LiteralInfinity)_
- LiteralReal _(renamed to LiteralRational)_
- StakeholderDefinition
- StakeholderUsage
- Superclassing _(renamed to Subclassification)_

**KerML Language Updates**

1. _Member Import._ Updated Import to handle single-membership imports. The notation **`import`**_`packageQualifiedName`_::`_`elementName`_`;` now sets the `importedNamespace` of the Import to the resolution of _`packageQualifiedName`_ and sets the `importedElementName` to _`elementName`_. The result is to import the Membership in the `importedNamespace` with the `memberName` given by the `importedElementName`, even if given name is an alias name. (This resolves ST6RI-375.)
2. _Import All._ Adding the **`all`** keyword to **`import`** now sets the `isImportAll` flag to true. Memberships are then imported without regard to visibility. Expose now always has `isImportAll = true`.
[...]

PR #260, opened 2021-07-26
ST6RI-405 Concrete syntax for parallel states

seidewitz:
The 2021-06 metamodel baseline adds an `isParallel` flag to StateDefinition and StateUsage. A "parallel" state (`isParallel = true`) is one whose _substates_ are intended to run concurrently ("in parallel"). This is in contrast to a "non-parallel" state (`isParallel = false`), which should have exactly one substate being performed at any point in time.

This PR implements a notation for StateDefinitions and StateUsages with `isParallel = true`:

- **`state def`**_`Name`_**`parallel`**`{ ... }`
- **`state`**_`name`_`:`_`Definition`_**`parallel`**`{ ... }`

The positioning of the **`parallel`** keyword is intended to indicate that it is the substates in the body that are "parallel".

The following additional validation checks have also been implemented for StateDefinitions and StateUsages that are not abstract:

1. If the StateDefinition or StateUsage is parallel, then it may not contain any Successions or TransitionUsages.
2. If the StateDefinition or StateUsage is _not_ parallel, then
    a. It must have at least one transition whose source is its entry action.
    b. Every composite substate must have at least one incoming transition.

The PR also includes the following changes that are also related to state modeling, though not specifically to the **`parallel`** notation.

1. Previously, entry, do and exit actions had to appear at the very beginning of the body of a StateDefinition or StateUsage. Now they may appear anywhere in the body, in any order (but at most one of each). However, to avoid ambiguity, it is no longer possible to use the shorthand form **`do`**_`action`_**`then`**_`target`_`;` for a TransitionUsage from the lexically preceding state with an effect action but no trigger or guard. Instead the form **`transition do`**_`action`_**`then`**_`target`_`;` must be used. (The **`transition`** keyword can also be used if the TransitionUsage has a **`accept`** and/or **`if`** clause, but it is then optional.)

2. Prior to the implementation of ST6RI-403 in PR #251,  the output parameter part of the trigger action of the TransitionUsage (i.e., **`accept`**_`outName`_`:`_`Type`_) was parsed as a parameter owned by the AcceptActionUsage acting as the trigger for the TransitionUsage, redefining the payload parameter from the base AcceptAction library type. PR 251 changed this so that the parameter was parsed as being directly by the TransitionUsage, redefining the payload parameter from the base library type StateTransitionAction (which is bound to the payload parameter of the trigger AcceptAction). This PR changes the parsing of TransitionUsage back so that the payload parameter is owned by the trigger AcceptActionUsage. The TransitionUsage still gets its own payload parameter, but this now subsets the trigger action payload parameter (for typing) and takes on its name (for convenience of reference).


seidewitz:
@himi
Note that this PR reverses the change to the parsing of TransitionUsage that was causing errors with the PlantUML state machine visualization. With this reversal, the visualization now seems to work correctly again.

himi:
@seidewitz I confirmed the fix.  I can update the visualization for parallel state.  Would you like to include it in this PR?

seidewitz:
> I can update the visualization for parallel state. Would you like to include it in this PR?

Go ahead.

himi (approved):

PR #261, opened 2021-07-26
ST6RI-419 Time instant quantity and ISO 8601 domain library model

hpdekoning:
Basic upgrade of the Time library package to support:
* time instant
* time scale
* Coordinated Universal Time (UTC) scale
* ISO 8601-1 date and time representation

A number of TODOs are recorded, mainly concerning the implementation of more details.

Also validation case 15_13 "Discretely Sampled Function Value" is updated to make use of the new Time package for the definition of a mission elapsed time scale. It shows how a specialization of a TimeScale can be mapped to another TimeScale (here UTC) by using placement attribute that defines a CoordinateSystemTransformation. This is an example of the unification of the treatment of scalar and higher order quantity types.


seidewitz (approved):
The functionality included is fine for this release (and much needed). The names of ISO 8601-related types seem a bit verbose, though, and perhaps the model could be made a bit simpler for simply specifying a date/time using an ISO 8601 string.

However, we can consider going forward will working on the other TODOs. So I am going to approve the PR for now.

PR #262, opened 2021-07-27
ST6RI-404 Concrete syntax for use cases

seidewitz:
**Concrete Syntax**

1. _Use cases._ Adds notation for UseCaseDefinition and UseCaseUsage using the keywords **`use case def`** and **`use case`**, respectively.
2. _Actors._ Adds notation for actors parameters of requirements and cases (not just use cases). An actor parameter is declared as a PartUsage, but using the keyword **`actor`** instead of **`part`**, i.e., with the form **`actor`**_`usageDeclaration`_`;`. A requirement or cases can have multiple actor parameters (similar to stakeholder parameters).
3. _Included use cases._ Added notation for IncludeUseCaseUsage, similar to PerformActionUsage, but using **`include`** instead of **`perform`**.

**Systems Library**

1. _UseCases._ Added library package `UseCases` with base definition `UseCase` and base usage `useCases`.
2. _Cases._ Added an `actors` feature to `Case`, which is subsetted by all actor parameters nested in CaseDefinitions or CaseUsages.
3. _Requirements._ Added an `actors` feature to `RequirementCheck`, which is subsetted by all actor parameters nested in RequirementDefinitions or RequirementUsages. (Also replaced the `ConcernCheck::concernedStakeholders` feature with `RequirementCheck::stakeholders`, since requirements in general can have stakeholder parameters, not just concerns.)

**Validation Case**

Implemented validation case `18-Use Case` validation case, except for the following, which are not yet directly supported by the syntax and are, therefore, currently mocked-up.

1. Pre- and post-conditions.
2. Use case extension.


seidewitz:
@himi @santileortiz 
FYI, for future implementation of use case diagram visualization.

@Weilkiti @YvesTLS 
FYI, for mapping from SysML v1.

PR #273, opened 2021-09-03
ST6RI-437: Render unnamed actions and states, and remove inappropriate satisfying feature links. (PlantUML)

himi:
This PR fixes the visualization problems of:

- Unnamed actions and states are not rendered;
- Wrong `<<satisfy>>` links may appear when SatisfyRequirementUsage is incomplete;
- PlantUML error in sequence view reported in ST6RI-436.

### DIFF
--- a/org.omg.sysml.interactive/src/org/omg/sysml/interactive/SysMLInteractiveHelp.java
+++ b/org.omg.sysml.interactive/src/org/omg/sysml/interactive/SysMLInteractiveHelp.java
@@ -67,7 +67,7 @@ public class SysMLInteractiveHelp {
     	    + "   ORTHOLINE\t\tOrthogonal line style\n"
     	    + "   COMPTREE\t\tShow tree structures in compartments\n\n"
     	    + "Example:\n"
-    	    + "  %viz --view Tree --style LR --style ortholine Pkg1::PartDef Pkg1::Pkg2::partUsage\n"
+    	    + "\t%viz --view Tree --style LR --style ortholine Pkg1::PartDef Pkg1::Pkg2::partUsage\n"
     	    + "should visualize Pkg1::PartDef and Pkg1::Pkg2::partUsage with a tree view ordered in the left-to-right direction with orthogonal lines.\n";
 
     private static final String VIEW_HELP_STRING =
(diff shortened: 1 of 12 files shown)


himi:
In the review, it is pointed out that messages should be rendered as open arrows because they are defined as asynchronous ones.  So I fixed it in 6bfdcd3.

seidewitz:
In all the examples I have tried so far, none of the SatisfyRequirementUsages are being visualized at all. Has the "satisfy" visualization been removed, or are there specific cases in which it is still shown?

himi:
@seidewitz the situation I encountered is not a normal situation.  We have many SatisfyRequirementUsage for example, in 8-Requirements validation case.  Normally, such SatisfyRequirementUsage has satisfiedRequirement and satisfyingFeature.  However, typically during the incomplete build, these cannot be resolved and these might return null.  In this situation, wrong <<satisfy>> links had been rendered.  This fix prevents such wrong link rendering.

seidewitz:
The 8-Requirements validation case model includes the following:
```
	part 'vehicle1_c2 Specification Context' {
		import 'vehicle1-c2 Specification'::*;
		
		requirement 'vehicle1-c2 Specification' {
			subject vehicle : Vehicle;
			require vehicleMass2;
			require vehicleReliability2;
		}
		
		satisfy 'vehicle1-c2 Specification' by vehicle1_c2;
	}
```
The satisfy in this model used to be rendered incorrectly (as a dashed arrow pointing to a final node symbol). Now it is not rendered at all:
image
This about as normal a satisfy as there is. Nothing is incomplete.

himi:
I confirmed it in 8-Requirements validation case... By attaching a debugger, I found satisfyingFeature was null... I'm investigating the reason.

seidewitz:
You are right, `basicGetSatisfyingFeature` was implemented incorrectly. I created a separate issue for this, in order to track it as a bug (see [ST6RI-444](https://openmbee.atlassian.net/browse/ST6RI-444)). I have also made the fix and create PR #275 for it. 

When this fix is merged into the ST6RI-390 branch, the satisfy arrows are then rendered between the correct elements. However, the arrow is going in the wrong direction.

seidewitz (changes requested):
The `<<satisfy>>` arrows are being rendered in the wrong direction. They should go _from_ the satisfying feature _to_ the satisfied requirement.

himi:
@seidewitz Thank you.  I corrected the direction.

himi:
BTW, Sandy requested to add some example in the help string.  I'll add it.

seidewitz (approved):

PR #277, opened 2021-09-12
ST6RI-445:Visualize parameters, inherited members, and more members in compartments

himi:
This change adds three new style options to specify how members are visualized.

```
SHOWINHERITED    Show inherited members
SHOWLIB                 Show elements of the standard libraries
COMPMOST            Show as many memberships in a compartment as possible
 
```


himi:
I made this PR mainly for the documentation.  Some of them are Sandy's request.  It's not an urgent matter and please review when you have time.

seidewitz (changes requested):
With the style "Show as many memberships in a compartment as possible", there does not seem to be any "parts" compartment. Instead the part memberships simply disappear. Either there should be a parts compartment, or the part memberships should still be shown graphically if it is not "possible" to show them in a compartment

seidewitz:
Nice!

PR #280, opened 2021-09-28
ST6RI-453 Derived values are no longer being written when a model is saved using %publish

seidewitz:
For 2021-08, a flag was added to the `JSONElementProcessingFacade` to control whether derived property values are included when generating the JSON representation of a model. This was to allow more compact export of JSON for file-based interchange. However, the default for the flag is false, but the `SysMLInteractive` implementations of `%publish` and `%show` were not updated to set the flag to true when writing to the repository or otherwise generating JSON.

This PR fixes `SysMLInteractive` so that it works as before. It also updates the repository save utilities to add a `-d` flag for including derived property values, as had been previously added to the JSON file export utility (the XMI utility still exports without derived values, at it always has). The Eclipse launchers for the save utilities have been updated to add the `-d` flag, so they also operate as before.

**This is a critical fix to be included in a 2021-08.1 maintenance release.**

### DIFF
--- a/org.omg.kerml.expressions.xtext.ide/META-INF/MANIFEST.MF
+++ b/org.omg.kerml.expressions.xtext.ide/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Automatic-Module-Name: org.omg.kerml.expressions.xtext.ide
 Bundle-ManifestVersion: 2
 Bundle-Name: org.omg.kerml.xtext.ide
 Bundle-Vendor: SysML v2 Submission Team
-Bundle-Version: 0.15.1.qualifier
+Bundle-Version: 0.15.2.qualifier
 Bundle-SymbolicName: org.omg.kerml.expressions.xtext.ide; singleton:=true
 Bundle-ActivationPolicy: lazy
 Require-Bundle: org.omg.kerml.expressions.xtext,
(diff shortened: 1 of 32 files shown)


seidewitz:
@santileortiz 

I discovered this bug when saving to the repository using the 2018-08 (v0.15.1) release. Without derived property values, the visualization does not work, of course. You can see the problem in the `sysml2-sst` repository by opening the project, which was published from Jupyter last night:

`ActionTree Tue Sep 28 01:45:49 EDT 2021`

This is republished using the fix in this PR in the following project (v0.15.2), which seems to visualize correctly:

`ActionTree Tue Sep 28 15:34:29 EDT 2021`

(The saves in between these times are using an interim update I did to support the SE DSIG demo today.)

I have now also pushed the entire usual suite of tests models to the `sysml2-dev` repository with the updated save utility. Please confirm that these work correctly with the visualization tool.

seidewitz:
@ivan-gomes
Once we do the 2021-08.1 release, we will need to update the npm and conda-forge repositories for 0.15.2.

ivan-gomes (approved):

ivan-gomes:
> @ivan-gomes Once we do the 2021-08.1 release, we will need to update the npm and conda-forge repositories for 0.15.2.

No problem.

santileortiz (commented):
The projects on `sysml2-dev` look good. We noticed that the models were also pushed to the public server without derived property values. Are we going clear this repository later?, I think it would be good, so we avoid people thinking the visualization is broken.

seidewitz:
I pushed those models to the public server by accident. I have asked Manas to have the public server cleared tomorrow, before I announce the new point release.

santileortiz (commented):
Great, thanks!

PR #288, opened 2021-10-28
ST6RI-407 Revised concrete syntax for connections

seidewitz:
The concrete syntax for connections in SysML has been updated as follows. (The abstract syntax is unchanged.)

1. _Bindings_
   - Old:
      **`bind`**`b1`**`as`**`p1.a1.x = p2.a2.y;`
      **`bind`**`p1.a1.x = a2.y;`
   - New:
      **`binding`**`b1`**`bind`**`p1.a1.x = p2.a2.y;`
      **`bind`**`p1.a1.x = p2.a2.y;`
2. _Successions_
   - Old:
      **`succession`**`s1`**`first`**`p1.a1`**`then`**`p2.a2;`
      **`succession`**`p1.a1`**`then`**`p2.a2;`
   - New:
      **`succession`**`s1`**`first`**`p1.a1`**`then`**`p2.a2;`
      **`first`**`p1.a1`**`then`**`p2.a2;`
3. _Flow Connections_
   - Old:
      **`stream`**`f1`**`of`**`T`**`from`**`a1.x`**`to`**`a2.y;`
      **`stream`**`a1.x to a2.y;`
   - New:
      **`flow`**`f1`**`of`**`T`**`from`**`a1.x`**`to`**`a2.y;`
      **`flow`**`a1.x`**`to`**`a2.y;`
4. _Succession Flow Connections_
   - Old:
      **`flow`**`f1`**`of`**`T`**`from`**`a1.x`**`to`**`a2.y;`
      **`flow`**`a1.x`**`to`**`a2.y;`
   - New:
      **`succession flow`**`f1`**`of`**`T`**`from`**`a1.x`**`to`**`a2.y;`
      **`succession flow`**`a1.x`**`to`**`a2.y;`

The KerML concrete sytnax for item flows and succession item flows has also been updated to be consistent with items 3 and 4 from above.


himi:
I tried the following but it did not work with `no viable alternative at input 'flow'`.

```
package TestConnections {
	
	action a1 {
		out port pa1;
	}
	action a2 {
		in port pa2;
	}
	
	item def T;
	
	flow a1.pa1 to a2.pa2;
	flow from a1.pa1 to a2.pa2;
	flow f1 of T from a1.pa1 to a2.pa2;
}
```

seidewitz:
The grammar requires that flows be nested in a definition or usage, not just in a package. Whether this makes sense any more is probably a different issue.

seidewitz:
@himi
Actually, considering this more closely, KerML allows item flows directly in packages, so I am not sure why the SysML grammar was still disallowing flow connections in packages. Therefore, I have updated the SysML grammar to allow that. With this update, your `TestConnections` model should parse without error.

himi (approved):
Thank you.  Now it woks with
```
package TestConnections {
	attribute x;
	attribute y;
	
	port def P;
	
	port p1;
	port p2;
	
	action a1 {
		out port pa1;
	}
	action a2 {
		in port pa2;
	}
	
	connection def B {
		end attribute ba;
		end attribute bb;
	}
	
	interface def I {
		end port p1 : P;
		end port p2 : P;
	}
	
	occurrence def OC;
	
	item def T;
		
	binding b1 : B bind x = y;
	interface i1 : I connect p1 to p2; 
	succession s1 : OC first a1 then a2;
	first a1 then a2;
	
	flow f1 of t1 : T from a1.pa1 to a2.pa2;
	flow a1.pa1 to a2.pa2;
	flow of T from a1.pa1 to a2.pa2;
	flow from a1.pa1 to a2.pa2;
	
	succession flow sf1 of t2 : T from a1.pa1 to a2.pa2;
	succession flow from a1.pa1 to a2.pa2;
	succession flow of T from a1.pa1 to a2.pa2;
	succession flow a1.pa1 to a2.pa2;
}
```

kerml/src/examples/Behavior Examples/TakePicture.kerml
```diff
@@ -12,7 +12,7 @@ behavior TakePicture() {
 	step step1: Focus[1] (out xrsl);	
 	step step2: Shoot[1] (in xsf);
 	
-	flow exposure[1] from step1.xrsl to step2.xsf;
+	succession flow exposure[1] from step1.xrsl to step2.xsf;
```
hpdekoning, line 15: This has nothing to do with the current PR, but just a question for my understanding: is multiplicity [1] on Focus, Shoot and exposure mandatory here? Or would they be the default when omitted?

hpdekoning (approved):
All good. Did not see any quirks.

kerml/src/examples/Behavior Examples/TakePicture.kerml
seidewitz: KerML does not specify any special multiplicity defaults. Unless otherwise give explicitly, the multiplicity is as inherited from the implicit base feature element, which is 0..*. In this case, though, even if the model was done in SysML, with actions instead of steps, the default multiplicity would still be 0..*. That is because a behavioral feature is supposed, by default, to be able to be performed zero or more times, unless otherwise constrained (and a flow is also a behavioral feature, as well as a connection).

PR #292, opened 2021-11-08
ST6RI-467 Include Use Case names could not be resolved

seidewitz:
Note: The commit related to ST6RI-461 should have been pushed separately, but now that it is accidentally included in this PR, it is not worth the trouble to change it. The second commit contains the resolution to ST6RI-467.


himi (approved):
I confirmed the fix.  IncludeUseCaseUsage.getEffectiveName() also works.

PR #294, opened 2021-11-22
ST6RI-470 Update to Eclipse version 2021-09

seidewitz:
Updated for Eclipse 2021-09:

1. `org.omg.sysml.installer/SysML2.setup`
2. Top level `pom.xml`
3. Top level `README.adoc`

Verified a clean install using the Oomph SysML2 install (for Mac x86_64). (Also requires a regeneration of all Xtext grammars and a rebuild of the workspace.)

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -19,7 +19,7 @@ The project provides a configuration file for the https://wiki.eclipse.org/Eclip
 +
 image:installer-advanced.png[switch to advanced mode]
 
-4. In the product list, select "Eclipse Modeling Tools" with Product Version: 2021-03.
+4. In the product list, select "Eclipse Modeling Tools" with Product Version: 2021-09.
 
 5. In the Projects window, to the right of the text box, click the "+" to the add the Sysml2 project as user project. In the dialog, choose Catalog: Github Projects and browse the file system to select the setup file from the local git repository.
    * e.g. `file:/C:/git/sysml/SysML-v2-Pilot-Implementation/org.omg.sysml.installer/SysML2.setup`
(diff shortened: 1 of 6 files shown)


himi:
Basically it worked in my environment.  But one Xpect test case failed only in maven execution (it does not fail in Eclipse Xpect JUnit environment.)  Does it occur in your environment?  
```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running 15_01-Constants.sysml.xt: src/org/omg/sysml/xpect/tests/expression
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.903 s <<< FAILURE! - in 15_01-Constants.sysml.xt: src/org/omg/sysml/xpect/tests/expression
[ERROR] 15_01-Constants.sysml.xt: src/org/omg/sysml/xpect/tests/expression  Time elapsed: 0.015 s  <<< FAILURE!
org.junit.ComparisonFailure:
ERROR:Couldn't resolve reference to Element ''m/s?''. (file:/C:/Users/himi/git/SysML-v2-Pilot-Implementation/org.omg.sysml.xpect.tests/src/org/omg/sysml/xpect/tests/expression/15_01-Constants.sysml.xt line : 114 column : 91)
ERROR:Must be a valid feature (file:/C:/Users/himi/git/SysML-v2-Pilot-Implementation/org.omg.sysml.xpect.tests/src/org/omg/sysml/xpect/tests/expression/15_01-Constants.sysml.xt line : 114 column : 91)
ERROR:Must be a valid feature (file:/C:/Users/himi/git/SysML-v2-Pilot-Implementation/org.omg.sysml.xpect.tests/src/org/omg/sysml/xpect/tests/expression/15_01-Constants.sysml.xt line : 114 column : 91) expected:<...80665['m/s?'];
    []}

    package 'Mo...> but was:<...80665['m/s?'];
    [                                                                                      777777
1: error - Couldn't resolve reference to Element ''m/s?''.
2: error - Must be a valid feature
4: error - Must be a valid feature
    ]}

    package 'Mo...>
```

himi:
I think we should update xtext.version and tycho.version but I think we should resolve the issue of the test case fail before that.

seidewitz:
The Xpect test runs successfully in my environment, both as a Junit test and during the Maven build. As you noted, it also passed during the Travis build from the repository.

Perhaps there is a character encoding problems. The name 'm/s?' is actually supposed to be 'm/s²'.

You are right that the Xtext and Tycho versions should be updated. Xtext 2.25.0 (the latest) is actually installed locally, but I forgot to update it in the POM.

himi:
@seidewitz, the problem apparently stems from encoding.  Java default encoding depends on the environment and the default encoding wrongly read *.sysml and *.xt files.  I added -Dfile.encoding=UTF-8 to the configuration to avoid this problem.  I'm not sure why this problem happened in 2021-09.  I'm now testing it with Xtext 2.25.0 and Tycho 2.5.0 to align with 2021-09 environment.

himi (approved):
Everything works right.

PR #295, opened 2021-11-23
ST6RI-469 Update implementation to 2021-10 metamodel baseline

seidewitz:
This update makes only one substantive change to the metamodel: The derived attribute `Type::disjointType` has been removed. As a result, the method `getDisjointType` has been removed from the Java classes `Type` and `TypeImpl`.


seidewitz:
@manasbajaj @ivan-gomes 
I am going to close this pull request. This is just a heads up that, even though the change to the metamodel is very small, I expect it will require an update to the API implementation.

PR #299, opened 2021-11-25
ST6RI-468 Variation cases cause an error of "A variation must only have variant owned members."

seidewitz:
Updated the validation check for variants to ignore objective memberships.


himi (approved):
It worked.

PR #301, opened 2021-11-28
ST6RI-480: Show connections in compartments and render flow connections (PlantUML)

himi:
Show connections in compartments.  Flow connections are properly rendered with `sourceOutput` and `targetInput`.

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VActionMembers.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VActionMembers.java
@@ -31,7 +31,6 @@ import org.omg.sysml.lang.sysml.FeatureDirectionKind;
 import org.omg.sysml.lang.sysml.ForkNode;
 import org.omg.sysml.lang.sysml.JoinNode;
 import org.omg.sysml.lang.sysml.MergeNode;
-import org.omg.sysml.lang.sysml.ParameterMembership;
 import org.omg.sysml.lang.sysml.Type;
 
 public class VActionMembers extends VBehavior {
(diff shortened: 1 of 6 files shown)


himi:
```
package TestConnections {
    part p1 {
        attribute x;
        attribute y;
        
        port def P;
        
        port p1;
        port p2;
        
        action a1 {
            out port pa1;
        }
        
        action a2 {
            in port pa2;
        }
        state st1;
        state st2;
        
        
        connection def B {
            end attribute ba;
            end attribute bb;
        }
        
        interface def I {
            end port p1 : P;
            end port p2 : P;
        }
        
        occurrence def OC;
        
        item def T;
        
        connection con : B connect a1.pa1 to a2.pa2;
                
        binding b1 : B bind x = y;
        interface i1 : I connect p1 to p2; 
        interface p1 to p2;
        succession s1 : OC first a1 then a2;
        first a1 then a2;
        
        flow f1 of t1 : T from a1.pa1 to a2.pa2;
        flow a1.pa1 to a2.pa2;
        flow of T from a1.pa1 to a2.pa2;
        flow from a1.pa1 to a2.pa2;
        
        succession flow sf1 of t2 : T from a1.pa1 to a2.pa2;
        succession flow from a1.pa1 to a2.pa2;
        succession flow of T from a1.pa1 to a2.pa2;
        succession flow a1.pa1 to a2.pa2;
    }
}
``` gives
![Screen Shot 2021-11-28 at 3 15 38 PM](https://user-images.githubusercontent.com/10537/143784343-04841bc5-4081-49b1-a963-aa5fe789fcaa.png)

himi:
Basically this works but as show in the example, item flow always has an item feature named `item`.  So `item' name is annoying even if we do not specify any names in textual notations.  Do we have any appropriate way to remove such default name?  I tried to check the type of the item feature, but it does not work with `flow of T from a to b`, which also has `item` name.

seidewitz:
On further review, I realized that a simple fix for the `item` name issue is simply to use the regular name, not the effective name, for the flow item feature. That is, simply replace `getFeatureName(f)` in `VDefault.itemFlowDesc` with `f.getName()`. The syntax for these features does not allow explicit redefinition anyway, so I think this small change will give us the expected visualization for now, until (and if) we think of something better.

seidewitz:
I also noticed that your compartment text for succession flows uses the `then` keyword. However, the textual notation we decided on still uses `to` in succession flows (as is the case in your `TestConnections` example). The visualization should be consistent with the textual notation.

himi:
@seidewitz, thank you.  I fixed the code to address your comments.

seidewitz (approved):

PR #304, opened 2021-12-04
ST6RI-487 Concrete syntax for assignment

seidewitz:
This PR implements a concrete syntax for assignment, as a surface notation for an AssignmentActionUsage.

1. _Assignment to feature in scope._ Assigns the result of _`valueExpression`_ to the feature with a _`featureName`_ in the current scope.
    
    **`assign`**_`featureName`_`:=`_`valueExpression`_`;`

    **Example:** **`assign`**`currentStatus := StatusEnum::Nominal;`
   
2. _Assignment to feature of a target occurrence._  Assigns the result of a _`valueExpression`_ to the feature with name _`featureName`_ of the occurrence that is the result of a _`targetExpression`_.
    
    **`assign`**_`targetExpression`_`.`_`featureName`_`:=`_`valueExpression`_`;`
    
    **Example:** **`assign`**`vehicleTrailerSystem.vehicle.hitch := trailer.hitchBall;`

    **Note:** In this case the _`targetExpression`_ is a path step expression, `vehicleTrailerSystem.vehicle`, and the _`valueExpression`_ is also a path step expression `trailer.hitchBall`.

An assignment action can also have an **`action`** declaration part of the usual form, e.g.,

**`action`**`updateStatusToNominal :>> updateStatus`**`assign`**`currentStatus := StatusEnum::Nominal;`

An assignment actions with any of the above syntax can appear in the body of another action, or directly as a do, entry or exit action (similarly to send and accept actions).

~~This PR does _not_ include syntax for including an assignment of an "initial value" in the declaration of a feature, pending further consideration of the abstract syntax and semantics.~~


hpdekoning:
I've read through all updates and did not spot any issue However I also do not master the Xtext environment sufficiently to fully review the consequences of all details involved, therefore this does not account for a full detailed review. From my side the change can be merged.

himi (approved):
I quickly tested and confirmed it worked.

seidewitz:
I added a preliminary implementation of FeatureValues with `isInitial = true`,  e.g.,

**`attribute`**`count : Integer := 0;`

These result in the declared Feature having a nested FeatureWritePerformance Step (for KerML) or an AssignmentActionUsage (for SysML), unless `isDefault = true`, too, e.g.,

**`item`**`sensor`**`default`**`:= primarySensor;`

PR #305, opened 2021-12-07
ST6RI-486: Render RequirementConstraintMemberships in compartments

himi:
Render RequirementConstraintMemberships in compartments when COMPMOST is enabled or the constraint is anonymous

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VCompartment.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VCompartment.java
@@ -76,6 +76,10 @@ public class VCompartment extends VStructure {
 
     private final boolean compartmentMost;
 
+    protected boolean isCompartmentMost() {
+        return compartmentMost;
+    }
+
     protected boolean rec(Membership m, Element e, boolean force) {
         VTree subtree = parent.subtree(m, e, force);
         if (subtree == null) return false;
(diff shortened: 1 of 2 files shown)


seidewitz (changes requested):
The code

```
requirement <'2.1'> vehicleMass2: MassLimitationRequirement {
	doc /* The vehicle mass shall be less than or equal to 2500 kg when the fuel tank is empty. */
	subject vehicle : Vehicle;	
	attribute :>> massActual: MassValue = vehicle.mass;
	attribute :>> massReqd = 2500 [kg];
	assume constraint fuelConstraint {
		doc /* empty fuel tank */
		vehicle.fuelLevel == 0.0
	}
}
```
is visualized as follows:
image

The rendering of the assumed constraint has a few problems:

1. The `assumes` keyword is italic rather than bold (`requires` is correctly bold).
2. There is a spurious nested `constraint fuelConstraint` construct.
3. The line breaks between nested body elements are lost, making the text difficult to read.

himi:
Regarding 3, I made it unfold texts but still it contains intends.   Could you please check the results to see if it fits with your request?

seidewitz:
Your resolutions to points 1 and 3 are OK. But there are still problems with 2.

image

The line **`assume`**`fuelConstraint constraint fuelConstraint` should be just `**`assume`**`fuelConstraint`. Similarly, an unnamed required constraint appears as **`require`**`constraint`, when it should just be **`require`** (as it is already in a "constraints" compartment).

himi:
The corrensponding text of ConstraintUsage constains `constarint` keyword and it's tricky to exclude that.  I tried to construct text programatically but it requires a lot of cases including documentation.   So I decided to just exclude the keyword.

PR #307, opened 2021-12-10
ST6RI-487 Xtext grammar converter can't handle "use case" keyword

seidewitz:
Revised the grammar converter to handle usage keywords that have spaces in them.

Regenerated the highlighting files including the new **`assign`** keyword and `:=` symbol.


tool-support/syntax-highlighting/xtext_grammar_converter.py
```diff
@@ -255,8 +255,8 @@ def parse_xtext_grammar(self, xtext_source_paths: List[str]):
                 logging.info(f"{var:{max_name_len}s} = {keywords_dict[var]}")
                 if var.endswith("DefKeyword"):
                     def_keyword_var = keywords_dict[var][0].split()[0]
-                    def_keyword_value = [x for x in keywords_dict[def_keyword_var] if x.isalpha()][0]
-                    def_keywords_set.add(def_keyword_value)
+                    def_keyword_value = [x for x in keywords_dict[def_keyword_var] if x.replace(" ","").isalpha()][0]
+                    def_keywords_set.add(def_keyword_value.split()[-1])
```
hpdekoning, line 259: For "use case"  `def_keywords_set.add(def_keyword_value.split()[-1])` only adds "use". This works for now since "case" is already a self-standing reserved word. A more generic solution would be:
```
    for keyword in def_keyword_value.split():
        def_keywords_set.add(keyword)
```
so that all composed individual keywords get added to the set.

hpdekoning (commented):
Looks fine! 
I do suggest one improvement to ensure adding any keyword from a composed, space-separated sequence to the set of defined keywords.

tool-support/syntax-highlighting/xtext_grammar_converter.py
seidewitz: Note that the index in the argument expression to the `add` is `-1`. That is, it gets the _last_ word. For "use case", this would be "case", not "use". In fact, **`use`** is _not_ itself a definitional keyword, it is only when it is used in conjunction with **`case`** that it serves in this way. But it is only keyword **`case`** that actually comes right before the name of the element being defined. (It is true that **`case`** happens to be a self standing keyword, but the coded approach is intended to work even if it wasn't.)

hpdekoning: Ah, my bad. But then this assumes that a compound is always a pair of a definitional keyword prepended with a kind of qualifier, correct? I could add such rules / assumptions explicitly so that when there is a change in the syntax violations would be flagged.

seidewitz: I don't think I am assuming any more about this than you already were in your code. The code preceding the above that really identifies what the "def keywords" are hasn't changed. And it was already including "use case" in that set. It was just that the code in line 258 was stumbling over the space in "use case". However, if only line 258 is changed, then "use case", including the space, gets included in `def_keywords_set`, which causes problems later.

Now, I believe that the "def keywords" you are collecting are those that are "definitional" in the sense used in Jupyter highlighting, which really means they precede the first declaration of a name. (In this sense, both our usage  and definition declarations are "definitional".) If the keyword immediately before a name has been identified as definitional, then the name gets a special coloring to indicate that this is where it is being declared. The fact that "use" might come before "case" is not important for this -- it is only important that "case" can come immediately before the name being declared, which "use" never does.

hpdekoning: Thanks, and clear. I had forgotten what definitional exactly meant.

PR #309, opened 2021-12-20
Fix pattern matches in angle bracket

Rots:
Fixes an error reported by maven:
```
This Pom pom.xml is in the Wrong Format: Line 217, column 50: Illegal character in attribute value: '>' -> 
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -214,8 +214,8 @@
                   <fileset dir="${basedir}">
                     <include name="*/feature.xml"/>
                   </fileset>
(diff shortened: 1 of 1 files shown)


himi:
XML allows the use of `>` in attributes so that I do not think this fix is needed. (See https://www.w3.org/TR/2006/REC-xml11-20060816/.  Actually I checked that it is well-formed.). What version of maven are you using?

seidewitz:
@Rots 
I am going to close this PR. If you wish to follow up further on the comment made by @himi, please open an issue rather than submitting a PR. If we ultimately decide to make a change, we will handle the update through our internal task tracking and development process.

Issue #220, opened 2021-03-29
Eclipse freezes on autocomplete inside Packages (arguably belongs over here)


gSeaVelo:
https://github.com/Systems-Modeling/SysML-v2-Release/issues/6
Is really about the pilot implementation - don't want it to get missed

seidewitz:
Content assist has been fully disabled in the 2021-12 release. See PR #313 for details.

PR #313, opened 2022-01-04
ST6RI-484 Concrete syntax for control structures

seidewitz:
### New Concrete Syntax

This PR adds concrete textual syntax for conditional and looping control structures. This notation may only be used within the body of a containing ActionDefinition or ActionUsage.

1. _If actions (IfActionUsages)_
   - An if action evaluates a Boolean-valued condition. If the result is true, it performs a nested then-action, otherwise it performs a nested else-action.
   - An if action has the form
     **`action`**_`ifName`_**`if`**_`condition`_**`action`**_`thenActionName`_`{` ... `}`**`else action`**_`elseActionName`_`{` ... `}`
   - The **`action`** declaration part may be omitted for the entire if action, the then-action and/or the else-action: 
     **`if`**_`condition`_`{` ... `}`**`else`**`{` ... `}`
     The curly braces are required, even if the body of the then-action or else-action is empty, unless the else-action is itself an if action, in which case a simplified notation can be used:
     **`if`**_`condition`_`{` ... `}`**`else if`**`{` ... `}`**`else`**`{` ... `}`
     This is equivalent to
     **`if`**_`condition`_`{` ... `}`**`else`**`{`**`if`**`{` ... `}`**`else`**`{` ... `} }`
   - The else-action is optional. If it is omitted, then no action is performed if the _`condition`_ is false.
   - The implicit base action usage for an if action is `Actions::ifThenActions` (of type `IfThenAction`), if the if action does _not_ have an else-part, and `Actions::ifThenElseActions` (of type `IfThenElseAction`), if it _does_ have an else-part. The action definition `Actions::IfAction` specializes the Kernel behavior `IfThenPerformance`, while `Actions::IfThenElseAction` specializes `IfThenElsePerformance`.

2. _While loop actions (WhileLoopActionUsages)_
   - A while-loop action repeatedly performs a nested body action as long as a Boolean-valued while-condition is true and an until-condition is false. The while-condition is evaluated before each iteration, and the until-condition is evaluated after the iteration.
   - A while loop action has the form
     **`action`**_`whileLoopName`_**`while`**_`whileCondition`_**`action`**_`bodyName`_`{` ... `}`**`until`** _`untilCondition`_`;`
   - The **`action`** declaration part may be omitted for the entire while-loop action and/or the body action:
     **`while`**_`whileCondition`_`{` ... `}`**`until`** _`untilCondition`_`;`
     The curly braces are required, even if the body is empty.
   - The until-condition is optional. If it is omitted, then the default is **`until`**`false` (that is, no effect).
     **`while`**_`whileCondition`_`{` ... `}`
   - The while-condition may be replaced with the keyword **`loop`**, which is equivalent to **`while`**`true`.
     **`loop`**`{` ... `}`**`until`** _`untilCondition`_`;`
     If the until-condition is also omitted, then the action specifies a non-terminating loop.
     **`loop`**`{` ... `}`
   - The implicit base action usage for a while-loop action is `Actions::whileLoopActions` (of type `WhileLoopAction`). The action definition `Actions::WhileLoopAction` specializes the Kernel behavior `LoopPerformance`.

3. _For loop actions (ForLoopActionUsages)_
   - A for-loop action iterates over the values resulting from the evaluation of a sequence-expression, assigning each in turn to a for-variable feature and performing a nested body action for the assigned value.
   - A for-loop action has the form
     **`action`**_`forLoopName`_**`for`**_`loopVariableDecl`_**`in`**_`sequenceExpr`_**`action`**_`bodyActionName`_`{` ... `}`
     The for-variable is declared as a ReferenceUsage, but without the **`ref`** keyword.
   - The **`action`** declaration part may be omitted for the entire for-loop action and/or the body action:
     **`for`**_`loopVariableDecl`_**`in`**_`sequenceExpr`_`{` ... `}`
   - The implicit base action usage for a for-loop action is `Actions::forLoopActions` (of type `ForLoopAction`). The action definition `Actions::ForLoopAction` defines the behavior of a for-loop in terms of a nested while-loop.

### Custom Xtext ANTLR Parser Generator

**Problem**

The addition of the above new concrete syntax to the SysML grammar resulted in the Java error
```
The code for the static initializer is exceeding the 65535 bytes limit
```
in the class `org.omg.sysml.xtext.ide.contentassist.antlr.internal.InternalSysMLParser`. Using ANTLR class-splitting didn't help, because the problem is with one very large string array that can't be split, and no modifications to the grammar was found to prevent it.

It is already a known issue that content assist does not work with the SysML and KerML editors (see #220), so the generated content-assist parser is not even useful for our current purposes. However, even when `ContentAssistFragment2` is removed from the generation workflow, a content-assist parser is still unconditionally generated by `XtextAntlrGeneratorFragment2`.

**Resolution**

1. A new class `org.omg.kerml.expressions.xtext.NoContentAssistAntlrGeneratorFragment` was created as a subclass of `XtextAntlrGeneratorFragment2`, which eliminates the generation of the content-assist parser. (The class was placed in the `org.omg.kerml.expressions.xtext` project, because this is used by both `org.omg.kerml.xtext` and `org.omg.sysml.xtext`.) 
2. The generation workflows for all of `KerMLExpressions.xtext`, `KerML.xtext` and `SysML.xtext` were updated to replace `XtextAntlrGeneratorFragment2` with `NoContentAssistAntlrGeneratorFragment`, and remove the generation of anything else that depends on the content-assist parser.
3. All three grammars were regenerated without content-assist parsers, eliminating the problem. In addition, attempting to use content assist no longer causes the editors to hang, since it is now not implemented at all.


seidewitz:
@ivan-gomes
Branch ST6RI-484 includes an update to the `SysML.uml` that can be used to generate a new API release. It is expected that this branch will be merged into master in a day or so.

org.omg.kerml.xtext.ui/plugin.xml_gen
```diff
@@ -227,19 +227,6 @@
 			uriExtension="kerml">
 		</resourceServiceProvider>
 	</extension>
-	<extension
```
ujhelyiz, line 230: Please check whether these changes need to be copied to the plugin.xml file of the same project. The basic idea here is that Xtext is not capable of updating the plugin.xml files (as it could contain manual customizations as well), so an alternative version Is generated where the developers are expected to copy the changes.

org.omg.kerml.expressions.xtext.ui/plugin.xml_gen
```diff
@@ -330,68 +330,6 @@
 			</command>
 		</menuContribution>
 	</extension>
-	<!-- quickfix marker resolution generator for org.omg.kerml.expressions.xtext.KerMLExpressions -->
```
ujhelyiz, line 333: Please check whether these changes need to be copied to the plugin.xml file of the same project. The basic idea here is that Xtext is not capable of updating the plugin.xml files (as it could contain manual customizations as well), so an alternative version Is generated where the developers are expected to copy the changes.

org.omg.sysml.xtext.ui/plugin.xml_gen
```diff
@@ -330,68 +330,6 @@
 			</command>
 		</menuContribution>
 	</extension>
-	<!-- quickfix marker resolution generator for org.omg.sysml.xtext.SysML -->
```
ujhelyiz, line 333: Similar to the KerML extensions, these changes should be copied to the corresponding plugin.xml file.

ujhelyiz (commented):
Please be careful as some extension registrations were removed from the plugin.xml_gen files, but they are not removed from the plugin.xml files. This might cause issues when Xtext would try to access code not available anymore.

org.omg.kerml.xtext.ui/plugin.xml_gen
seidewitz: Thanks, Zoltán. Good catch.

I compared the `plugin.xml_gen` files to the existing `plugin.xml` to confirm that all the generated changes should be included in the production files. Then I updated the production files with all the generated changes (in all three projects).

PR #314, opened 2022-01-06
ST6RI-478 Add more time relation modeling to prevent HappensBefore from linking overlapping occurrences

seidewitz:
This PR updates the `HappensDuring` and `HappensBefore` associations in the `Occurrences` Kernel Library model with cross-relation and transitivity constraints, per ST6RI-478. It also revises `HappensJustBefore` as suggested in ST6RI-461.

**Note:** The KerML grammar had to be updated to properly allow feature chains at both ends of Subsetting and Disjoining relationships. The substantive change to the library model is in the final commit, e5f0d5a.


seidewitz:
Added your proposed documentation as a comment in `HappensDuring`.

PR #315, opened 2022-01-06
ST6RI-488: Render imported memberships having explicit names

himi:
Possibly due to the meta model change, imported memberships were not rendered and 13a-Model Containment was not visualized properly.  It can be tested with a simple example:
```
    package Pkg1 {
        part def P11;
        part def P12;
    }

    package Pkg2 {
        import Pkg1::P11;
        part p2 {
            ref part p11 : P11;
        }
    }
```

I would like to render them with dashed line style but it requires the update of PlantUML and it is deferred.


seidewitz (changes requested):
With `import Pkg1::P11;`, the `p11 : P11` relationship is visualized showing a second symbol for P11 within `Pkg2`. If the import is changed to `import Pkg1::*;`, then the visualization shows the `p11 : P11` relationship as pointing to the symbol for `P11` within `Pkg1`. However, in the abstract syntax, the result of either import is exactly the same as far as `P11` is concerned: the owning membership of `P11` in `Pkg1` becomes an imported membership of `Pkg2`. So the two cases should really be rendered in the same way.

Note that the dashed-boundary import notation does not actually apply here. So far, that notation has only been specified for `*` package import, indicating that the boundary of the package has been effectively "removed" by the import (see the release 2021-11 SysML Spec, 7.4.3, Table 8, and 8.2.3.4.1). It does not apply to element import.

himi:
I understand your point.  However, we need to treat import with explicit names for two reasons.
1) Even if the imported element is not referred, it should be shown.  If we have the example below, `P11` should be visualized, actually that is the case of 13a validation case.  If we use `import Pkg1::*`, it's not practical to show all of the elements.
```
    package Pkg2 {
        import Pkg1::P11;
    }
```

2) For the visualizer, we need to identify the corresponding memberships to properly show elements, but it is not easy for visualizers to identify them from the parsed results unless name resolvers give such information.

Let's consider the following example:

```
package Pkg1 {
   package Pkg11 {
      part def P1;
   }
   package Pkg12;
}
package Pkg2 {
   import Pkg1::Pkg11:*;
   package Pkg21 {
     import Pkg1::Pkg12:*;
     part p : P;
  }
}
```

In this example, visualizers should show part def `P` in `Pkg2` instead of `Pkg21` but we cannot get it from the parsed results.  Visualizer should get such information from name resolvers or need to compute the same result by itself, which is inefficient and impractical.

himi:
Regarding the dashed-boundary notation, I'd like to have a good way to distinguish directly owned elements from imported ones.  I think it would be an issue in the Graphical WG.

seidewitz:
The problem is, visualizing an imported elements inside a package symbol is entirely ambiguous with the visualization of an owned element. I think this would be confusing to most modelers as a default visualization. The expectation would be, I think, that a model like
```
    package Pkg2 {
        import Pkg1::P11;
    }
```
would be visualized using the dashed-arrow notation from the symbol for `Pkg2` to the symbol for `Pkg1::P11`, not by containment.

As to your point that "it is not easy for visualizers to identify them from the parsed results", I agree that this is an issue, and we may address it in the future. However, this is also an issue with the UML/SysML v1 abstract syntax. Indeed, UML/SysML v1 visualization always proceeds based on abstract syntax meta-references, regardless of naming or imports. That will make it particularly confusing if it is otherwise (at least by default) for SysML v2.

Finally, consider the case when a reference is fully qualified, but redundant with an import.
```
    package Pkg2 {
        import Pkg1::P11;
        part p2 {
            ref part p11 : Pkg1::P11;
        }
    }	
```
In the rendering as you propose it, the visualization is different whether or not there is an import, even though this is totally irrelevant to the model. I doubt this is what most modelers would expect.

So, over all, I do not believe your propose visualization for element imports is correct.

himi:
I'm quite busy these days and have much time to work on this but let me confirm that for the example below
```
package Pkg2 {
    import Pkg1::P11;
}
```
is this visualization

Screen Shot 2022-01-09 at 11 30 58 PM

ok with you?
I think this

Screen Shot 2022-01-09 at 11 32 32 PM

is what I would like to render first and Sandy also prefer this.  In my understanding, both notations should be ok and aligned with the discussions so far in Graphical WG.  Previously, I preferred the first `<<import>>` notation but in the actual rendering, this style gets busier than the second style, I found.  I think I can support both styles but it requires the modification of PlantUML and need to defer it to the next sprint if it is mandatory.

seidewitz:
The visualization
image
is OK with me.

Sandy has been using the dashed line notation with packages to mean package import, that is `import Pkg2::*`, _not_ `import Pkg2`. So, the visualization
image
would mean `import Pkg1::P11::*;`, _not_ `import Pkg1::P11;`. Further, if the dashed-line notation is extended to importing any kind of namespace, that would include usages, and the notation would then be ambiguous with the dashed-line notation for non-composite features on a nested interconnection diagram.

That is why, right now, the draft spec includes the dashed-line notation _only_ for `*` import from a package.

himi:
I've just noticed that only `<<import>>` notation covers all of the 3 types of import, element-level, wildcard (*), recursive(**) whilst dashed notation is applicable to wildcard only, at least in your statement.  To be honest, I think, it's quite confusing.  If we adopt dashed notation, it should cover all of the three types as `<<import>>` notation does, possibly annotated with (* or **) at the heading.   So I will support `<<import>>` notation first and then I'd like to discuss it in the Graphical WG.

seidewitz:
I am closing this pull request, since ST6RI-488 has been moved to the 2022-01 sprint, and the implementation  may change significantly. Please open a new PR once the implementation has been updated.

PR #316, opened 2022-01-10
ST6RI-416 Complete SI and USCustomaryUnits libraries

hpdekoning:
This is the result of the first attempt to add a complete set of US Customary Unit definitions to the Quantity and Units Library.
It is still not fully complete, but it should provide a good temporary version.
A small number of problematic definitions in `USCustomaryUnits.sysml` are commented out.
A full update will be finalized in the coming 2022-01 sprint, issue [ST6RI-491](https://openmbee.atlassian.net/browse/ST6RI-491) together with issue [ST6RI-415](https://openmbee.atlassian.net/browse/ST6RI-415).

### DIFF
--- a/org.omg.sysml.xpect.tests/library.domain/Quantities and Units/SI.sysml
+++ b/org.omg.sysml.xpect.tests/library.domain/Quantities and Units/SI.sysml
@@ -13,302 +13,301 @@ package SI {
     /*
      * SI simple units needed before creation of base units
      */
-    attribute g: MassUnit { :>> longName = "gram"; }
+    attribute <g> gram : MassUnit;
 
     /*
      * SI base units
      */
-    attribute m: LengthUnit { :>> longName = "metre"; }
-    attribute kg: MassUnit { :>> longName = "kilogram"; :>> unitConversion: ConversionByPrefix { :>> prefix = kilo; :>> referenceUnit = g; } }
-    attribute s: DurationUnit { :>> longName = "second"; }
-    attribute A: ElectricCurrentUnit { :>> longName = "ampere";  }
-    attribute K: TemperatureUnit {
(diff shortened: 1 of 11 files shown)


hpdekoning:
I added one small correction on the symbols defined for the Imperial and US gallon, `<gal_imp>` and `<gal>`.

seidewitz:
It is a bit problematic that the newly generated units in `USCustomaryUnits` use the new human ID/name convention, while the older unit declarations still use the `longName` attribute. I had thought that you had perhaps already  regenerated the `SI` package using the new approach based on the earlier discussion of it, which is why I hadn't expected the discrepancy. It is also not mentioned in the comment in [ST6RI-416](https://openmbee.atlassian.net/browse/ST6RI-416) on what has and has not been covered.

Ideally, of course, we would use the new convention for all units in this release. However, if that is not practical at this time, it would still be better to have the new US units use the old `longName` convention until we can change everything consistently. Is it possible to make this update with only a day or two of work? If so, you do that before I merge the PR. If not, then we can defer this to [ST6RI-491](https://openmbee.atlassian.net/browse/ST6RI-491).

Either way, though, it should be documented in ST6RI-416 as being covered or not.

hpdekoning:
I have updated the ISQ / SI generator to produce the `SI.sysml` package using the new `<unitsymbol> 'unit long name'` syntax in line with `USCustomaryUnits.sysml`. I checked that no parsing errors occurred.

So, should be ready for release 2021-12 now.

seidewitz (approved):

seidewitz:
I updated the remaining non-generated units in `USCustomaryUnits` to also use the new syntax, and then I deleted `TensorMeasurementReference::longName` (and a few further references to it). Please confirm that this is acceptable.

PR #317, opened 2022-01-14
ST6RI-416 Complete SI and USCustomaryUnits libraries - Update: Applied manual corrections (from ST6RI-414) in SI.sysml

hpdekoning:
Applied manual corrections (from ST6RI-414) in SI.sysml for Henry and newton metre

When waking up I realized I had overlooked the manual correction from bug fix [ST6RI-414](https://openmbee.atlassian.net/browse/ST6RI-414). This problem will be fully resolved via issue [ST6RI-415](https://openmbee.atlassian.net/browse/ST6RI-415).


seidewitz (approved):
Since we haven't announced the release yet, I can sneak this in.

PR #321, opened 2022-01-24
ST6RI-495 Library model for monitoring feature changes

seidewitz:
This PR adds library models for behaviors that monitor features for changes in their values, per the 2022-01 metamodel baseline.

1. The Kernel Library package `FeatureAccessPerformances`has been renamed to `FeatureReferencingPerformances`.
2. The following behaviors have been added to `FeatureReferencingPerformances`:
    - `FeatureMonitorPerformance`
    - `EvaluationResultMonitorPerformance`
    - `BooleanEvaluationResultMonitorPerformance`
    - `BooleanEvaluationResultToMonitorPerformance`
3. The Java class `org.omg.sysml.util.ImplicitGeneralizationMap` has been updated to replace references to `FeatureAccessPerformances` with `FeatureReferencingPerformances`.
4. The KerML and SysML Xpect tests have been updated to replace `FeatureAccessPerformances` with `FeatureReferencingPerformances`.


seidewitz:
@conradbock 

Besides syntactic corrections, I made one substantive change to what you had proposed. In `FeatureMonitorPerformance` (and, so, in `BooleanEvaluationResultMonitorPerformance`, too), I changed the order of `afterValues` and `beforeValues` so that `afterValues` comes first. This is because `afterValues` redefines the (otherwise) inherited parameter `values`, and all inherited parameter redefinitions currently need to come before the declaration of any new parameters (like `beforeValues`). (This has to do with still allowing behavior invocations with the positional argument notation.)

The only alternative would be to have `beforeValues` redefine `values` instead of `afterValues`, in which case it could come before `afterValues`. This would be semantically valid, too, but would perhaps less what would be expected as the "primary output" of the behavior.

conradbock (changes requested):
1) From https://openmbee.atlassian.net/browse/ST6RI-493 (was in comment, not the attached text file, sorry):

> [snip] FeatureReferencingPerformances.kerml  ... change the comment at the top to:

```
/**
 * This package defines Behaviors used to read, write, and monitor values of a Feature
 * of an Occurrence.
 */
```

> Change the comment above FeatureReferencingPerformance to:

```
/**  
 * <p>A FeatureReferencingPerformance is a Performance generalizing
 * other Behaviors relating to <code>values</code> of a Feature of
 * <code>onOccurrence</code>, as specified in the specialized
 * Behaviors.</p>
 */
```

2) About BooleanEvaluationResultMonitorPerformance, does "bool" mean "feature typed by Boolean", with the inherited name still applying (this convention isn't always used, just checking)?   Also the result nested under monitorOcc isn't redefined to Boolean, I guess that's OK bc it's bound to before/afterValues?

3) "default true" doesn't constraint M0, whereas "= true" does, is that right?

seidewitz:
1. I updated the comments (but not exactly as suggested for FeatureReferencingPerformance).
2. The keyword **`bool`** means "feature typed by BooleanEvaluation". BooleanEvaluation is an Evaluation of a BooleanExpression (an Expression typed by a Predicate returning a Boolean). BooleanEvaluation already redefines the result to be Boolean, so it does not have to be done explicitly for `monitorOccurrence`.
3. Yes.

conradbock (approved):

PR #326, opened 2022-02-01
ST6RI-497 Concrete syntax for change and time triggers

seidewitz:
**Abstract Syntax**

This PR adds makes the following changes to the abstract syntax, over the original 2022-01 baseline (PR #320).

1. Adds TriggerInvocationExpression metaclass.
2. Removes `triggerKind` property from AcceptActionUsage metaclass.
3. Removes `accept` literal from TriggerKind enumeration.

**Concrete Syntax**

This PR adds the following concrete syntax related to AcceptActionUsages.

1. _Accept actions._ The notation for AcceptActionUsages is updated to allow the payload parameter to be declared with a feature value or incoming flow. In this case, an accepted transfer must have a payload identical to the bound or flowed value.
   - **`accept`**_`payloadName`_`:`_`PayloadType`_`=`_`valueExpression`_`;`
   - **`accept`**_`payloadName`_`:`_`PayloadType`_**`flow from`**_`source`_`;`
   
   **Note.** The syntactic form **`accept`**_`PayloadType`_`;`, without a colon in the payload declaration, is still allowed and still represents a declaration of the payload _type_, not its name. A binding or flow _cannot_ be included in a declaration of this form.
2. _Change triggers._ A _change trigger_ of the form **`when`**_`changeCondition`_ can be used in place of a feature value in a payload declaration. The _`changeCondition`_ should be a Boolean-valued expression, and the trigger fires when the expression result changes from false to true (or if the result is true when first evaluated). A change trigger is parsed as a TriggerInvocationExpression with `kind = when`, invoking the Kernel Library `Triggers::triggerWhen` function, with the _`changeCondition`_ as its first argument (as a `BooleanEvaluation`) and the AcceptActionUsage `receiver` parameter as its second argument, returning a `ChangeSignal`.
   - **`accept`**_`payloadName`_`:`_`PayloadType`_**`when`**_`changeCondition`_`;`

   **Note.** The explicit declaration of the payload parameter is optional, allowing the simple form **`accept when`**_`changeCondition`_`;`. The form **`accept`**_`payloadName`_**`when`**_`changeCondition`_`;` is also allowed, but this represents a declaration of the payload _name_, _not_ its type.
3. _Time triggers._ A _time trigger_ of the form **`at`**_`timeExpression`_ or **`after`**_`durationExpression`_ can be used in place of a feature value in a payload declaration. The _`timeExpression`_ should result in a `TimeInstantValue`, and the trigger fires at that time (relative to the default clock). The _`durationExpression`_ should result in a `DurationValue`, and the trigger fires after the elapse of that duration (on the default clock). A time trigger is parsed as a TriggerInvocationExpression with `kind = at` or `kind = after`, invoking the Kernel Library `Triggers::triggerAt` or `Triggers::triggerAfter` function, respectively, with the result of the _`timeExpression`_  or _`durationExpression`_ as its first argument and the AcceptActionUsage `receiver` parameter as its second argument, returning a `TimeSignal`.
   - **`accept`**_`payloadName`_`:`_`PayloadType`_**`at`**_`timeExpression`_`;`
   - **`accept`**_`payloadName`_`:`_`PayloadType`_**`after`**_`durationExpression`_`;`

   **Note.** Simplified payload declarations are allowable for time triggers, similarly to change triggers.

**Model Library**

This PR includes the following model library changes, over those made in PR #321 and PR #322.

1. Replaced the use of `ScalarValue` for time and duration values in the `Clocks` package with `NumericalValue` (consistent with the usage in `Triggers`). Also changed the name of the `TimeOf::timeValue` result parameter to `timeInstant`.
2. _Quantities._ Made `ScalarQuantityValue` in the `Quantities` package a specialization of `NumericalValue` instead of `ScalarValue`.
3. _SI._ Removed **`import`**`QuantityCalculations::*` from the `SI` package. 
4. _Time._ Changed the name of the `timeValue` result parameter of the `TimeOf` calculation definition in the `Time` package to `timeInstant`. Also changed the order of the public owned features of the `Iso8601DateTime` and `Iso8601DateTimeStructure` calculation definitions so the non-redefining parameters come first (allowing a more natural use of positional argument notation when invoking these calculations).


seidewitz:
@ivan-gomes @manasbajaj 

The `SysML.uml` file on this branch is ready to be used for API generation.

@hpdekoning 

Notice the small changes in the Quantities and Units `Time` library model.

himi:
I quickly read `5-State-based Behavior` validation cases.  I hope we have examples of new syntaxes, 

 - **`accept`**_`payloadName`_`:`_`PayloadType`_`=`_`valueExpression`_`;`
 - **`accept`**_`payloadName`_`:`_`PayloadType`_**`flow from`**_`source`_`;`
 
as well as the existing syntax of **`accept`**_`PayloadType`_`;`

himi:
So I tried the following:
```
package TestAccept {
    item def I1;

    action a1 {
        out item o1 : I1;
    }

    state s1 {
        entry; then s11;
        state s11;
        accept i1: I1 flow from a1.o1 then s12;
        state s12;
    }
}
```
But it caused the warning of "Should be an accessible feature (use dot notation for nesting)."  What's wrong with this?

seidewitz:
This looks like a bug in the validation, and it is not specific to accept actions. The following also causes a warning, on the flow into `a3::i3` (but _not_ on the flow into `a2.i2`).
```
    action a1 {
        out item o1 : I1;
    }
    
    action a2 {
        in item i2 flow from a1.o1;
    	action a3 {
    	    in i3: I1 flow from a1.o1;
    	}
    }
```
I will look into it.

seidewitz:
Update:

The validation is failing on the source, rather than the target of the flow, which is implicit in the shorthand "flow from" form. However, the failure also happens in the following case, where the target `a3.i3` is specifically highlighted.
```
    action a1 {
        out item o1 : I1;
    }
    
    action a2 {
        in item i2 flow from a1.o1;
    	action a3 {
    	    in i3: I1;
            flow from a1.o1 to a3.i3;
    	}
    }
```
The warning above can be eliminated by replacing `a3.i3` with `a2.a3.i3`, but that isn't very helpful in the shorthand form when the target is implicit.

Since this problem is not specific to the changes being made in this PR, I will create a separate bug report for it.

himi (approved):
I tried 
```
package TestAccept {
    item def I1;
    import ScalarValues::Integer;

    action a1 {
        out item o1 : I1;
    }

	attribute time : Time::TimeInstantValue;
	
	attribute val : Integer;

    state s1 {
        entry; then s11;
        state s11;
        accept i1: I1 flow from a1.o1 then s12;
        state s12;
        accept i2: Integer = 5 * 3 then s13;
        state s13;
        accept after time then s14;
        state s14;
        accept when val > 5 then s15;
        state s15;
    }
}
``` and it seems to be parsed correctly except for the warning.

File: sysml.library/Kernel Library/Clocks.kerml [hunk 1]
```diff
@@ -44,7 +44,11 @@ package Clocks {
 	 * which is considered to be synchronized with the snapshot of the Clock 
 	 * with a currentTime equal to the returned timeValue.
 	 */
-	abstract function TimeOf(o : Occurrence[1], clock : Clock[1]) timeValue : ScalarValue[1] {
+	abstract function TimeOf {
+		in o : Occurrence[1];
+		in clock : Clock[1] default defaultClock;
+		return timeValue : ScalarValue[1];
+
```
hpdekoning, line 51: I would prefer using `timeInstantValue` rather than just `timeValue` to stress the difference with `duration`. Even in ISO 80000-3 `duration` is aliased to `time`, and many people do this. Humans mostly intuitively understand whether "time something takes" vs "the time at which something happens" is intended, but in the library I think we should be rigorous.

sysml.library/Kernel Library/Clocks.kerml
```diff
@@ -44,7 +44,11 @@ package Clocks {
 	 * which is considered to be synchronized with the snapshot of the Clock 
 	 * with a currentTime equal to the returned timeValue.
 	 */
-	abstract function TimeOf(o : Occurrence[1], clock : Clock[1]) timeValue : ScalarValue[1] {
+	abstract function TimeOf {
+		in o : Occurrence[1];
+		in clock : Clock[1] default defaultClock;
+		return timeValue : ScalarValue[1];
```
hpdekoning, line 50: I would prefer using `timeInstantValue` rather than just `timeValue` to stress the difference with duration. Even in ISO 80000-3 "duration" is aliased to "time", and many people do this. Humans mostly intuitively understand whether "time something takes" vs "the time at which something happens" is intended, but in the library I think we should be rigorous.

hpdekoning (commented):
I just have the two in-lined comments.

Main one is to add a change to `Quantities.sysml` as follows:
`ScalarQuantityValue` should specialize `NumericalValue`, and then `NumericalValue` can be used throughout `Clocks.kerml` and `Triggers.kerml`.

sysml.library/Kernel Library/Clocks.kerml
seidewitz: 1. I renamed the `timeValue` parameters to `timeInstant`.
2. Done.

sysml.library/Kernel Library/Triggers.kerml
```diff
@@ -5,7 +5,7 @@
  */
 package Triggers {
 	private import ScalarValues::Boolean;
-	private import ScalarValues::NumericalValue;
+	private import ScalarValues::ScalarValue;
```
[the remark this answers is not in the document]
seidewitz: Done. I also updated the PR description to reflect this.

seidewitz:
@himi 

See [ST6RI-502](https://openmbee.atlassian.net/browse/ST6RI-502) for an analysis of the bug you identified. I implemented a temporary workaround on this branch that disables the validation check for the case when the flow connection target is generated by the implementation. A full resolution will be addressed later.

himi:
@seidewitz I confirmed that disabled the validation and the warning disappeared.

PR #328, opened 2022-02-12
ST6RI-491 Complete SI and USCustomaryUnits libraries (continued)

hpdekoning:
The upgrade for generating the Q&U library packages from NIST SP811 is now reasonably complete. The generator splits the units defined in SP811 as follows:

- USCustomaryUnits.sysml (humanId `<USCU>`) only containing US Customary Units, This is the main package to be used by end-users.
- NIST_SP811_CGS (humanId `<CGS>`) 'Centimetre-Gram-Second System of Units' library
- NIST_SP811_Imp  (humanId `<Imp>`) 'Imperial (Canadian and UK) Units' library
- NIST_SP811_SI_recognized (humanId `<SI_recognized>`) 'International System of Units - Units recognized in SI' library
- NIST_SP811_SI_deprecated (humanId `<SI_deprecated>`) 'International System of Units - Deprecated units' library.

The last four provide complete coverage of all units and conversion factors defined in SP811, but are of limited value to end-users, with the possible exception of the `<Imp>` library, that could provide a basis for extension with more Canadian/UK Imperial units, if needed. Since the last four files are completely auto-generated they can also be left out of the 2022-01 release.

A number of units are still commented out, due the unavailability (in ISO/IEC 80000) of their corresponding quantity attribute defs. These quantities will need to be provided in a next sprint.

In addition to Thermodynamic Temperature now also Temperature Difference is properly modeled, which finally enabled proper specifications of the degree Celsius and Fahrenheit scales, as well as the degree Celsius and Fahrenheit units.

The remaining work will be explicitly documented in issue [ST6RI-491](https://openmbee.atlassian.net/browse/ST6RI-491).

### DIFF
--- a/sysml.library/Domain Libraries/Quantities and Units/ISQ.sysml
+++ b/sysml.library/Domain Libraries/Quantities and Units/ISQ.sysml
@@ -18,4 +18,24 @@ package ISQ {
     import ISQCharacteristicNumbers::*; // ISO 80000-11 "Characteristic numbers"
     import ISQCondensedMatter::*;       // ISO 80000-12 "Condensed matter physics"
     import ISQInformation::*;           // IEC 80000-13 "Information science and technology"
+
+    /* Additional quantity declarations */
+
+    /** 
+     * temperature difference
+     * 
+     * A separate temperature difference quantity and unit are needed in order to support °C, °F and centrigrade temperature differences
+     */
+    attribute def TemperatureDifferenceUnit :> SimpleUnit {
+        private attribute thermodynamicTemperaturePF: QuantityPowerFactor[1] { :>> quantity = isq.'Θ'; :>> exponent = 1; }
+        attribute :>> quantityDimension { :>> quantityPowerFactors = thermodynamicTemperaturePF; }
+    }
+
+    attribute def TemperatureDifferenceValue :> ScalarQuantityValue {
+        attribute :>> num: Real;
+        attribute :>> mRef: TemperatureDifferenceUnit[1];
+    }
+    
+    attribute temperatureDifference: TemperatureDifferenceValue [*] nonunique :> scalarQuantities;
+
 }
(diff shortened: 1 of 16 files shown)


seidewitz:
I would prefer not to make the backwards-incompatible change of removing the `QuantityValue` and `quantities` aliases with this PR. I have also made the following relatively minor revisions:
- `QuantityCalculations`: Restricted Quantities import to just `ScalarQuantityValue`.
- `SI`: Removed import of `QuantityCalculations`. (This change has already been made on master.)
- `USCustomaryUnits`: Restored documentation comment to a consistent style with other packages.
- `VectorCalculations`: Restricted Quantities import to just `VectorQuantityValue`.

hpdekoning:
I noticed that in my initial comment the humanId were not showing in rendered markdown, due to the angle brackets. Corrected now.

OK, with your changes.

- Removal of `QuantityValue` and `quantities` aliases: I propose to schedule this as an issue then.
- Documentation of `USCustomaryUnits` reverted to consistent library style: yeah, I agree. I was just playing with a different style similar to Python docstrings, that I like and find very readable. Now that the libraries are getting bigger and more serious to maintain we may actually want to conceive a (concise) documentation / coding style guide.

seidewitz (approved):

seidewitz:
I also made the import of `SI` by `USCustomaryUnits` private (this change had also already been made on master).

seidewitz:
After further consideration, I decided to remove the `NIST_SP811...` packages from the release. We can discuss separately whether it is worth including of these in future releases.

hpdekoning:
@seidewitz "decided to remove the `NIST_SP811...` packages"
Yes, I agree, they are of little value in the release, more for completeness of the SP811 transformation.
We may consider to add them to the sysml/src/examples or similar.

PR #332, opened 2022-02-22
ST6RI-514 Update implementation for 2022-02 baseline

seidewitz:
**Abstract Syntax**

_Metaclasses removed:_

1. PathStepExpression
2. PathSelectExpression

_Metaclasses added:_

1. FeatureChainExpression
2. CollectExpression (replaces PathStepExpression)
3. SelectExpression (replaces PathSelectExpression)

**Concrete Syntax**

1. *`expr`*`.`*`x`*`.`*`y`*`.`*`z`* parses as a FeatureChainExpression whose `operator` is `.`,  `argument` is *`expr`* and `targetFeature` is *`x`*`.`*`y`*`.`*`z`*. 

   **Note:** Dot notation in expressions previously parsed to a PathStepExpression.

2. *`expr`*`.{...}` parses as a CollectExpression whose `operator` is `collect` and `arguments` are *`expr`* and the body expression `{...}`. It is equivalent to *`expr`*`-> collect {...}`. 

   **Note:** This notation was previously used for a PathSelectExpression, that is, a `select` rather than a `collect`.

3. *`expr`*`.?{...}` parses as a SelectExpression whose `operator` is `select` and`arguments` are *`expr`* and the body expression `{...}`. It is equivalent to *`expr`*`-> select {...}`. 

   **Note:** This replaces the old notation for PathSelectExpressions.

**Kernel Library**

_Added to `ControlFunctions`:_

1. `function '.'` used as the implicit type of a FeatureChainExpression.


seidewitz:
@manasbajaj @ivan-gomes 

The `SysML.uml` file on this branch is ready to be used for API generation.

seidewitz:
@himi 

Commit [11d827e](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/332/commits/11d827e8c6dd3d3ed1e1f161da0b137cb1e9a9bc) updates the PlantUML code to avoid compilation errors. However, these changes have not been tested.

Also, it looks like the code still presumes that the dot notation with multiple features is parsed into a hierarchy of nested expressions, as it was originally with PathStepExpression. This is no longer the case, however, with such notation now parsed into a single FeatureChainExpression whose target feature is a feature chain. Perhaps the related PlantUML code can now be simplified.

himi:
I think the current implementation will not work properly because FeatureChainExpression is not compatible with PathStepExpression.  I would need to work to fix this but I'm now quite busy.   By when do you want to finish this PR?

seidewitz:
On the one hand, I would like to get this PR done soon, just to have the implementation updated to the current metamodel baseline. On the other hand, most of the other work for the 2022-02 sprint is on library models, which doesn't actually depend on the metamodel update. So keeping this PR open isn't actually holding anything else up at this point.

However, note the new parsing of FeatureChainExpressions is really not much different than the parsing of PathStepExpressions as of the 2021-12 release.  From the release notes for 2021-12:

>  _Path expressions._ Previously, a path expression of the form _`expr`_`.a.b.c` was parsed as a tree of path step expressions. That is, it was parsed like `(`_`expr`_`.a).b).c)`. Such expressions are now parsed so that a trailing path of two or more features parses as a feature chain. That is, _`expr`_`.a.b.c` parses as a single expression that applies the feature chain (e.g., `a.b.c`) to the result of the expression _`expr`_. 

The main difference is that a PathStepExpression still had two operands, the second of which was a FeatureReferenceExpression whose referent was the feature chain, while a FeatureChainExpression has the feature chain as a direct member. So, it isn't clear that the PlantUML-related code in question has even applied since 2021-12, and it doesn't seem to have caused any problem. Perhaps you could identify how it is still being used, if at all. It may actually be mostly dead code at this point.

himi:
So I'll work on it this (mid) night.  I guess e.g. Send/AcceptActionUsage with FeatureChainExpression do not work.

seidewitz:
As far as I can tell so far, Send/AcceptActionUsages seem to work using FeatureChainExpressions (to specify the receiver occurrence), being visualized just as they were in the previous release using PathStepExpressions. I am not sure what other cases need to be checked (since connector and flow ends use feature chains directly and SatisfyRequirementUsage uses a FeatureReferenceExpression). But if everything is still actually being visualized OK, then we can proceed with this PR and consider any potential refactoring of the PlantUML-related code as a separate issue.

himi:
I'm now testing the fixed code but I found FeatureChainExpression.getTargetFeature() does not seem to return a feature having FeatureChaining but actually returns the feature to be referred.   How can I get a list of FeatureChaining from FeatureChainExpression?

seidewitz:
A FeatureChainExpression has two parts: a source Expression, which is its single `argument`, and a `targetFeature`. The semantics of a FeatureChainExpression is to chain the `result` of the source Expression with the `targetFeature`. The `targetFeature` may or may not itself be a Feature chain. The semantic chaining done in a FeatureChainExpression is implicit and does not appear in the parse tree.

So, a FeatureChainExpression of the form `expr.a` will have `expr` as its `argument` and `a` as its `targetFeature`, which will be an _unowned_ member. But a FeatureChainExpression of the form `expr.a.b` will have a`targetFeature` that is an _owned_ member with a Feature chain of the features `a` and `b`. The chained Features of the `targetFeature`, if it has them, are available as usual using the `targetFeature.chainingFeature` metaproperty. 

If the initial `expr` is itself just a Feature name, e.g., `x.a.b`, then the `argument` of the FeatureChainExpression will be a FeatureReferenceExpression whose `referent` is `x`. That is, it is parsed as a special case of the general FeatureChainExpression notation `expr.a.b`, rather than as, say, a FeatureReferenceExpression for the Feature chain `x.a.b`.

This is very similar to how it worked with PathStepExpression in the last release. The only substantive difference is that, instead of having a `targetFeature` member itself, the PathStepExpression had a second `argument` that was a `FeatureReferenceExpression` whose `referent` was the effective target Feature (this was more complicated and didn't quite work right semantically, hence the change). As with `targetFeature`, this `referent` could either be an unowned member or an owned member with a feature chain. 

(The other difference is that the `operator` for a PathStepExpression was `collect` while the `operator` for a FeatureChainExpression is the new `'.'` Function. But this shouldn't matter for visualization.)

himi:
I thought `x.a.b` as a single feature chain but actually it's not. As you say, it will be a combination of FeatureReferenceExpression `x` and FeatureChain `.a.b`  I can adapt to it but does it look complicated?

himi:
I think FeatureReferenceExpression should be extended to cover FeatureChaining rather than introducing FeatureChainExpression.  Is it possible?

seidewitz:
> I think FeatureReferenceExpression should be extended to cover FeatureChaining rather than introducing FeatureChainExpression. Is it possible?

It is not possible for FeatureReferenceExpression to cover the general case of when the first segment of the chain is itself an expression, because you end up with a circularity. Something needs to represent the leaf node of a reference to a name that is not a chain. And trying to parse `x.a.b` to FeatureReferenceExpression as a special case, rather than as a limiting case of `expr.a.b` would make the grammar very complicated (or require difficult semantic disambiguation or backtracking in the parser).

In any case, the purpose of this PR is to implement the 2022-02 metamodel baseline, which has FeatureChainExpression in it. As I have previously noted, this should be no more complicated to handle than PathStepExpression was in the previous release. Any further simplification, if possible, can be discussed separately for inclusion in a future abstract syntax update.

himi:
I fixed the code to cover all of the three patterns.   I confirmed the following test case worked:
```
package SendAction {
    attribute def S1 { attribute arg; }
    part p1 {
        action acceptAction1;
        part p11 {
            action acceptAction2;
        }
    }
    action sendAction(in arg) {
		send S1(arg) to p1.acceptAction1;
		send S1(arg) to p1.p11.acceptAction2;
    }
}
```

seidewitz:
Looks good,.

himi:
BTW, I found SysML.xtext contains an inconsistent line terminator at Line 1269.  Is it ok to fix that?  Emacs is screwed to detect the line terminate due to this line:

modified   org.omg.sysml.xtext/src/org/omg/sysml/xtext/SysML.xtext
@@ -1266,7 +1266,7 @@ fragment ActionBodyItem returns SysML::Type :
 	  ownedRelationship += StructureUsageMember
 	| ownedRelationship += InitialNodeMember
 	  ( => ownedRelationship += TargetSuccessionMember )*
-	| ( ownedRelationship += EmptySuccessionMember )? 
+	| ( ownedRelationship += EmptySuccessionMember )? 
 	  ownedRelationship += ( BehaviorUsageMember | ActionNodeMember )
 	  ( => ownedRelationship += TargetSuccessionMember )*
 	| ownedRelationship += GuardedSuccessionMember

himi (approved):
I confirmed .{} and .?{} are parsed correctly.

PR #337, opened 2022-03-01
ST6RI-506 Reference frame Kernel Library model

seidewitz:
This pull request adds the following Kernel Library packages:

1. `VectorValues` – Provides a basic model of abstract vectors as well as concrete vectors whose components are numerical values.
2. `VectorFunctions` – Defines abstract functions on `VectorValue` corresponding to the algebraic operations provided by a vector space with inner product. It also includes concrete implementations of these functions specifically for `CartesianVectorValue`.
3. `TrigFunctions` – Defines basic trigonometric functions on real numbers and provides a constant feature for `pi`. (Included in this PR to provide the `arccos` function needed in the implementation of `VectorFunctions::CartesianAngle.`)
4. `SpatialFrame` – Models spatial frames of reference for quantifying the position vector of a point in a three-dimensional space.

It also makes the following updates to existing packages:

_Kernel Library_

1. `DataFunctions` and `ScalarFunctions`
    - The `sum` and `product` functions have been removed from `DataFunctions` and `ScalarFunctions`.
2. `NumericalFunctions`
    - The `sum` and `product` functions are now in `NumericalFunctions`.
    - The functions `isZero` and `isUnit` have been added to the package, to allow zero and unit values of a `NumericalValue` type to be abstractly identified without having to construct them concretely.
3. `ComplexFunctions`, `RealFunctions`, `RationalFunctions`, `IntegerFunctions` and `QuantityCalculations`
    - Updated as necessary due to the above changes.

_Quantities and Units Library_

4. `Quantities` – Makes `VectorQuantityValue` a kind of `NumericalVectorValue`.
5. `VectorCalculations` – Revised to provide calculation definitions for `VectorQuantityValue` that specialize the corresponding functions from `VectorFunctions`.

Finally, it makes updates across the mathematical function packages with a more consistent convention of starting the names of non-operator mathematical functions with a lower case letter.

**Note:** This PR does _not_ include a SysML-level version of`SpatialFrame` with functions specialized to take `TimeInstantValue` and return `VectorQuantityValue`. This may instead be included in the Geometry Domain Library update in PR #340, depending on the resolution of the discussion there.

### DIFF
--- a/sysml.library/Domain Libraries/Quantities and Units/QuantityCalculations.sysml
+++ b/sysml.library/Domain Libraries/Quantities and Units/QuantityCalculations.sysml
@@ -4,6 +4,13 @@ package QuantityCalculations {
     import UnitsAndScales::ScalarMeasurementReference;
     import UnitsAndScales::DimensionOneValue;
     
+    calc def isZero :> NumericalFunctions::isZero(x: ScalarQuantityValue): Boolean {
+    	NumericalFunctions::isZero(x.num)
+    }
+    calc def isUnit :> NumericalFunctions::isUnit(x: ScalarQuantityValue): Boolean {
+    	NumericalFunctions::isUnit(x.num)
+    }
+    
     calc def '[' specializes BaseFunctions::'[' (x: NumericalValue, y: ScalarMeasurementReference): ScalarQuantityValue;
 
 	calc def Abs specializes NumericalFunctions::Abs (x: ScalarQuantityValue): ScalarQuantityValue;
@@ -37,12 +44,16 @@ package QuantityCalculations {
 	calc def ToReal(x: ScalarQuantityValue): Real;
 	calc def ToDimensionOneValue(x: Real): DimensionOneValue;
 	
-	calc def sum specializes ScalarFunctions::sum (collection: ScalarQuantityValue[0..*]): ScalarQuantityValue {
-		ScalarFunctions::sum0(collection, 0.0)
+	calc def sum specializes NumericalFunctions::sum (collection: ScalarQuantityValue[0..*]): ScalarQuantityValue {
+		private attribute zero : ScalarQuantityValue;
+		assert constraint { isZero(zero) }
+		NumericalFunctions::sum0(collection, zero)
 	}
 	
-	calc def product specializes ScalarFunctions::product (collection: ScalarQuantityValue[0..*]): ScalarQuantityValue {
-		ScalarFunctions::product1(collection, 1.0)
+	calc def product specializes NumericalFunctions::product (collection: ScalarQuantityValue[0..*]): ScalarQuantityValue {
+		private attribute one : ScalarQuantityValue;
+		assert constraint { isUnit(one) }
+		NumericalFunctions::product1(collection, one)
 	}
 
     calc def ConvertQuantity(x: ScalarQuantityValue, targetMRef: ScalarMeasurementReference): ScalarQuantityValue;
--- a/sysml.library/Domain Libraries/Quantities and Units/QuantityCalculations.sysml
+++ b/sysml.library/Domain Libraries/Quantities and Units/QuantityCalculations.sysml
@@ -11,7 +11,7 @@ package QuantityCalculations {
     	NumericalFunctions::isUnit(x.num)
     }
     
-    calc def '[' specializes BaseFunctions::'[' (x: NumericalValue, y: ScalarMeasurementReference): ScalarQuantityValue;
+    calc def '[' specializes BaseFunctions::'[' (x: Number, y: ScalarMeasurementReference): ScalarQuantityValue;
 
 	calc def abs specializes NumericalFunctions::abs (x: ScalarQuantityValue): ScalarQuantityValue;
 
--- a/sysml.library/Domain Libraries/Quantities and Units/QuantityCalculations.sysml
+++ b/sysml.library/Domain Libraries/Quantities and Units/QuantityCalculations.sysml
@@ -4,16 +4,16 @@ package QuantityCalculations {
     import UnitsAndScales::ScalarMeasurementReference;
     import UnitsAndScales::DimensionOneValue;
     
-    calc def isZero :> NumericalFunctions::isZero(x: ScalarQuantityValue): Boolean {
+    calc def isZero specializes NumericalFunctions::isZero(x: ScalarQuantityValue): Boolean {
     	NumericalFunctions::isZero(x.num)
     }
-    calc def isUnit :> NumericalFunctions::isUnit(x: ScalarQuantityValue): Boolean {
+    calc def isUnit specializes NumericalFunctions::isUnit(x: ScalarQuantityValue): Boolean {
     	NumericalFunctions::isUnit(x.num)
     }
     
     calc def '[' specializes BaseFunctions::'[' (x: NumericalValue, y: ScalarMeasurementReference): ScalarQuantityValue;
 
-	calc def Abs specializes NumericalFunctions::Abs (x: ScalarQuantityValue): ScalarQuantityValue;
+	calc def abs specializes NumericalFunctions::abs (x: ScalarQuantityValue): ScalarQuantityValue;
 
 	calc def '+' specializes NumericalFunctions::'+' (x: ScalarQuantityValue, y: ScalarQuantityValue[0..1]): ScalarQuantityValue;
 	calc def '-' specializes NumericalFunctions::'-' (x: ScalarQuantityValue, y: ScalarQuantityValue[0..1]): ScalarQuantityValue;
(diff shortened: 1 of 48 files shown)


sysml.library/Domain Libraries/Quantities and Units/QuantityCalculations.sysml
```diff
@@ -4,6 +4,13 @@ package QuantityCalculations {
     import UnitsAndScales::ScalarMeasurementReference;
     import UnitsAndScales::DimensionOneValue;
     
+    calc def isZero :> NumericalFunctions::isZero(x: ScalarQuantityValue): Boolean {
+    	NumericalFunctions::isZero(x.num)
+    }
+    calc def isUnit :> NumericalFunctions::isUnit(x: ScalarQuantityValue): Boolean {
+    	NumericalFunctions::isUnit(x.num)
+    }
+
```
hpdekoning, line 13: W.r.t. naming convention we have a mix of calc defs / functions using PascalCase and camelCase. I guess in line with the rest of the definitions we should use PascalCase. Perhaps raise as a future maintenance issue.

sysml.library/Kernel Library/VectorValues.kerml
```diff
@@ -0,0 +1,50 @@
...
+		feature dimension[1] :>> dimensions;
+		feature :>> elements : NumericalValue;
+	}
+	
+	/**
+	 * CartesianVectorValue is a specialization Numerical VectorValue for which there are 
+	 * specific implementations in VectorFunctions of the abstract vector-space functions.
+	 * 
+	 * Note: The restriction of the element type to Real is to facilitate
+	 * complete of thes functions.
```
hpdekoning, line 32: Typo: completion of these functions. ?

sysml.library/Kernel Library/VectorFunctions.kerml
```diff
@@ -0,0 +1,192 @@
+/**
+ * This package defines abstraft functions on VectorValues corresponding to the algebraic operations
```
hpdekoning, line 2: typo: abstract

sysml.library/Kernel Library/VectorFunctions.kerml
```diff
@@ -0,0 +1,192 @@
...
+	 */
+	abstract function Inner specializes DataFunctions::'*'(v: NumericalVectorValue[1], w: NumericalVectorValue[1]) x: NumericalValue[1] {
+		inv commmutivity { x == Inner(w, v) }
+		inv zeroInner { isZeroVector(v) or isZeroVector(w) implies isZero(x)}
+	}
+	
+	/**
+	 * The length of a NumericalVectorValue, as a NumericalValue.
+	 */
+	abstract function Length(v: NumericalVectorValue[1]) l : NumericalValue[1] {
```
hpdekoning, line 101: Should we use Magnitude instead of Length? In general a NumericalVectorValue may represent any kind of physical quantity.

sysml.library/Kernel Library/VectorFunctions.kerml
```diff
@@ -0,0 +1,192 @@
...
+	function CartesianVectorScalarMult specializes VectorScalarMult (v: CartesianVectorValue[1], x: Real[1]): CartesianVectorValue[1] {
+		CartesianScalarVectorMult(x, v)
+	}
+	
+	function CartesianInner specializes Inner(v: CartesianVectorValue[1], w : CartesianVectorValue[1]): Real[1]{
+		inv precondition { v.dimension == w.dimension }
+		(1..v.dimension)->collect{in i : Positive; v[i] * w[i]}->reduce RealFunctions::'+'
+	}
+	
+	function CartesianLength specializes Length(v: CartesianVectorValue[1]) : NumericalValue[1] {
```
hpdekoning, line 179: CartesianMagnitude ?

sysml.library/Domain Libraries/Quantities and Units/QuantityCalculations.sysml
seidewitz: Yes, we have been very inconsistent about this, which becomes more noticeable as we add more function libraries. On the one hand, using initial-uppercase names is consistent with the usual convention for classifiers/definitions. On the other hand, it is inconsistent with usual practice for mathematical functions like `sin` and `cos` and for functions adapted from OCL, like `includes` and `select`.

I had been thinking that we need to resolve this all together in a future issue, as you suggest. However, there are now inconsistencies even within a number of packages (e.g., `NumericalFunctions` has `isZero`, `isUnit`, `sum` and `product`, but `Abs`, `Min` and `Max`). I don't think we can fix everything this time around, but maybe I can at least revised the packages touched by this PR to use a consistent naming convention. I would propose to use initial-_lowercase_, because the functions in question are then mostly mathematical.

sysml.library/Kernel Library/VectorValues.kerml
seidewitz: Should be "complete definition of these functions".

sysml.library/Kernel Library/VectorFunctions.kerml
seidewitz: Good point, though `Magnitude` is a bit long. An alternative would be `norm`. Do you think that would be familiar enough to people?

sysml.library/Kernel Library/SpatialFrames.kerml
```diff
@@ -0,0 +1,79 @@
...
+	function CartesianPositionOf :> PositionOf {
+		in point : Point[1];
+		in frame : CartesianSpatialFrame[1];
+		return positionVector : CartesianThreeVectorValue;
+	}
+	
+	/**
+	 * The DisplacementOf two points relative to a CartesianSpatialFrame is a CartesianThreeVectorValue.
+	 */
+	function Cartesian3DDisplacementOf :> DisplacementOf {
```
hpdekoning, line 72: I think if we use `3D` here, we should use it in `CartesianPositionOf `as well.
In the SysML libraries I abbreviate to `3d` (lowercase d) because it reads better in PascalCase/camelCase.

hpdekoning (commented):
Just some minor corrections and suggestions

sysml.library/Domain Libraries/Quantities and Units/QuantityCalculations.sysml
hpdekoning: Yes, agreed

sysml.library/Kernel Library/VectorFunctions.kerml
hpdekoning: Hey, Julia uses `norm` (https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.norm), what can I say? So good for me, clear and short.

sysml.library/Kernel Library/SpatialFrames.kerml
seidewitz: Actually, I think meant to take `3D` off both of them. It seemed redundant once `SpatialFrame` is limited to three dimensions anyway.

hpdekoning: Ok, agreed

sysml.library/Domain Libraries/Quantities and Units/QuantityCalculations.sysml
seidewitz: Updated in commit [23247b9](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/337/commits/23247b9ff208ee7b43e70f066ef41795c5ff5fd8).

sysml.library/Kernel Library/VectorValues.kerml
seidewitz: Fixed in commit [3505da2](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/337/commits/3505da240c3983885d275d65d636fde6adcf7ad6).

sysml.library/Kernel Library/VectorFunctions.kerml
seidewitz: Fixed in commit [3505da2](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/337/commits/3505da240c3983885d275d65d636fde6adcf7ad6).

sysml.library/Kernel Library/VectorFunctions.kerml
seidewitz: Updated in commit [3505da2](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/337/commits/3505da240c3983885d275d65d636fde6adcf7ad6).

sysml.library/Kernel Library/VectorFunctions.kerml
seidewitz: Updated in commit [3505da2](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/337/commits/3505da240c3983885d275d65d636fde6adcf7ad6).

sysml.library/Kernel Library/SpatialFrames.kerml
seidewitz: Updated in commit [3505da2](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/337/commits/3505da240c3983885d275d65d636fde6adcf7ad6).

sysml.library/Kernel Library/SpatialFrames.kerml
```diff
@@ -0,0 +1,160 @@
...
+	 * defaultFrame is a fixed SpatialFrame used as a universal default.
+	 */
+	readonly feature defaultFrame : SpatialFrame[1];
+	
+	/**
+	 * A SpatialFrame is a three-dimensional Body that provides a spatial extent that 
+	 * acts as a frame of reference for defining the phyiscal position and displacement 
+	 * vectors of Points over time.
+	 */
+	abstract struct SpatialFrame specializes Body;
```
hpdekoning, line 26: I find it slightly non-intuitive to make SpatialFrame which I would consider intangible (a mental abstraction) to specialize Body which I would intuitively consider tangible. I know Body is defined to be a 3D spatial extent (finite region), but then the label "body" is perhaps not appropriate. In geometry often "Solid" is used to denote a finite 3D extent (as indeed STEP and my current Shapes.sysml does), although that is misleading in a different way as it may be a volume containing a non-solid state of aggregation: liquid, gas, plasma, ...
Can SpatialFrame be a [1..1] feature of Body, rather than being a Body?

hpdekoning (approved):

sysml.library/Kernel Library/SpatialFrames.kerml
conradbock: Could have a different name for Body, but then Object has similar dissonance, bc it's just an Occurrence that (informally stated in the spec) occupies a connected region of space time.  This covers things you can pick up ("tanglible") and things you can't, eg, a piece of dark matter.

seidewitz: A SpatialFrame is intended to be an Object (as Clock is) with a three-dimensional spatial extent. Naming aside, this is just what a Body is, tangible or intangible (as Conrad notes) – and there probably is no perfect naming. So I think it makes sense for SpatialFrame to be a Body.

conradbock:
Only scanned it.  Started working on my mother's taxes.  Do you need it this evening?

seidewitz:
At this point, tomorrow is OK.

conradbock (approved):
SpatialFrame description:
  phyiscal -> physical

spacePositionConstraint:
    * If the given point occurs within the extent of a Point that is a
    * spaceShot of the given frame,
Are the frames unrestricted in size?  If so, every 4D point is within it, in particular will be within one of its space shots (ie the above "if" is always true).  Looking at the code, would suggest rewording:
    * The result positionVector is equal to the PositionOf the
    * frame's Point spaceShot that encloses the given point.

General: I've been meaning to write a generic performance wrapper for relationships that are supposed to be outside of time, like TimeOf, PositionOf.  These give the same result for the same inputs whenever they're evaluated, so could be associations that happened to be "queried" by a performance.

conradbock:
Did you fix the typo in SpatialFrame descripttion (see above, also spacePositionConstraint)?

PR #340, opened 2022-03-08
ST6RI-507 Initial Basic Geometry Domain Library model for shapes

hpdekoning:
Initial version of Shapes.sysml and ShapeCalculations.sysml libraries plus 3 simple examples in sysml/src/Geometry Examples

The approach and examples are documented in a Monthly Track 6 presentation "Progress Basic Geometry Shapes".

### DIFF
--- a/sysml.library/Systems Library/Items.sysml
+++ b/sysml.library/Systems Library/Items.sysml
@@ -37,14 +37,14 @@ package Items {
 		
 		/**
 		 * This Item enveloped by all its envelopingShapes. That is, for each envelopingShape, there
-		 * is an Item with that shape for which this Item is a spaceSlice.
+		 * is an Item with that shape that encloses this Item in space for its entire lifetime.
 		 */
 		assert constraint envelopingShapeConstraint { 
 			envelopingShapes->forAll { in envelopingShape : Item;
 				item envelopingItem {
 					item :>> shape = envelopingShape; 
 				}
-				envelopingItem.spaceSlices->includes(Item::self)
+				envelopingItem.spaceTimeEnclosedOccurrences->includes(Item::self)
 			}
 		}
 		


conradbock (commented):
I gather Shape would specialize Object (to support curves, etc)?  Do you have a UML(ish) graphic?

conradbock (commented):
What is the purpose of the duplicate CSG model?  We (will have) Occurrence::unionsOf, intersectionsOf, and differencesOf.

sysml.library/Domain Libraries/Geometry/ShapeItems.sysml
```diff
@@ -0,0 +1,249 @@
...
+		attribute :>> area;
+		attribute :>> outerSpaceDimension = 2;
+	}
+
+	item def Circle :> Shell {
+		attribute radius :>> radius;
+
+		item :>> faces [0];
+		item :>> edges [1] {
+			attribute circumference :>> length = Circle::radius * 3.1415926535898;
```
seidewitz, line 44: Should use TrigFunctions::pi.

sysml.library/Domain Libraries/Geometry/ShapeItems.sysml
```diff
@@ -0,0 +1,249 @@
...
+		item :>> vertices [0];
+	}
+
+	/**
+	 * Surface bounded by a circle.
+	 */
+	item def Disc :> Shell {
+		attribute radius :>> radius;
+
+		item :>> spaceBoundary : Circle [1]
```
seidewitz: If a Circle and a Disc are both Shells (Surfaces), then what is the difference? How can a Disc, as a Surface, be bounded by a Circle, if that is also a Surface? Shouldn't a Circle be a Curve without a boundary, not a Surface?

sysml.library/Domain Libraries/Geometry/ShapeItems.sysml
```diff
@@ -0,0 +1,249 @@
...
+	private import UnitsAndScales::CoordinateTransformation;
+	private import ISQ::*;
+	private import SI::*;
+	private import Collections::KeyValuePair;
+	private import Collections::Map;
+	private import Objects::*;
+	private import Items::*;
+	private import SequenceFunctions::union;
+	private import CollectionFunctions::contains;
+	private import SequenceFunctions::isEmpty;
```
seidewitz, line 13: The following imports are unnecessary:
```
private import ScalarValues::String;
private import UnitsAndScales::CoordinateTransformation;
private import Collections::KeyValuePair;
private import Collections::Map;
private import SequenceFunctions::union;
private import CollectionFunctions::contains;
```
The imports from `SI` and `Items` don't need to be `*`:
```
private import SI::m;
private import Items::Item;
```

sysml.library/Domain Libraries/Geometry/ShapeItems.sysml
conradbock: Shells can be innerDim 1 by not having faces, as in Circle (Shell doesn't restrict StructuredSpaceObject, just renames it).  Perhaps we have a naming or factoring problem.  @hpdekoning, is there a common name for 1-dimensional "shells", like polygons, ie, curves that are potentially segmented into edges and vertices?

sysml.library/Domain Libraries/Geometry/ShapeItems.sysml
conradbock: Thx, I fixed this in my local copy.  Figured I'd wait for HP/your reply on the next comment to push.

sysml.library/Domain Libraries/Geometry/ShapeItems.sysml
conradbock: Also fixed locally, awaiting push.

sysml.library/Domain Libraries/Geometry/ShapeItems.sysml
seidewitz: > Oops, just noticed Shells are Surfaces, that's not right, or at least not consistent.

Yes, that was my point. 

Colloquially, at last, it makes sense to me to have a Shell be a surface. But then it doesn't make sense to have a Circle be a Shell. Per your previous comment, it would be nice to have a separate name for one-dimensional "shells". In the shape model, are these curves always going to be closed (like circles and polygons)? Could you repurpose "Line" for that, and rename the current "Line" to "StraightLine"? Or maybe "Path"?

conradbock: See proposal at HP and I just posted at [ST6RI-507](https://openmbee.atlassian.net/browse/ST6RI-507?focusedCommentId=13373) would be Shell of 2 dim and Path for 1 dim.  These could be open or closed.  Would that be good?

seidewitz: Fine with me.

sysml.library/Domain Libraries/Geometry/ShapeItems.sysml
conradbock: Done.

sysml.library/Domain Libraries/Geometry/ShapeItems.sysml
conradbock: Done.

sysml.library/Domain Libraries/Geometry/ShapeItems.sysml
conradbock: Done.

sysml.library/Domain Libraries/Geometry/ShapeItems.sysml
```diff
@@ -35,7 +23,20 @@ package ShapeItems {
...
-	item def Circle :> Shell {
+
+	/**
+	 * Most general surface shape.	Solids are objects with these as space boundary.	
+	 */
+	abstract item def Shell :> Item, StructuredSpaceObject, Surface	 {
+	}
+
+	/**
+	 * Most general curve shape.	Solids are objects with these as space boundary.
```
seidewitz, line 34: Don't solids have 2D Shells as space boundary, not 1D Paths?

conradbock: Cr*p, thx.  Will fix in push.  Item::isSolid=true requires them to be dim 3 (I checked the definition with HP).

sysml/src/examples/Geometry Examples/CarWithEnvelopingShape.sysml
```diff
@@ -0,0 +1,29 @@
+package CarWithEnvelopingShape {
```
conradbock, line 1: These depend on the removed files.  I assume the latest updates are on 507-1.  Could either delete them, or I could replace with examples, if needed.

hpdekoning:
Added updates to existing validation case 15-03 and Geometry Examples.

👍 OK for me for Release 2022-02.
This seems to provide the basics. Will formulate remaining tasks on JIRA for release 2022-03 tomorrow morning.

sysml/src/examples/Geometry Examples/CarWithEnvelopingShape.sysml
hpdekoning: We could add this one as a basic CSG example to sysml/src/examples/Geometry Examples

conradbock: Added this to the Geometry examples as CarWithShapeAndCSG.  Pushed.

conradbock (approved):

seidewitz:
@conradbock @hpdekoning 
Please hold off making any further pushes for the moment.

seidewitz:
- Added `SpatialItems` to the Geometry domain library.
- Added `envelopingShapes` feature to `Item`.
- Updated the Geometry Examples per the above.
- Removed the `WithinBoth` connection def from `ShapeItems` (and made other editorial updates). 

(The connection def is not necessary because connection usages can be explicitly typed by associations. However, they then are still also implicitly typed by `Connection`, which is an association structure. So the connections are semantically the same as if typed by the `WithinBoth` connection def.)

sysml.library/Domain Libraries/Geometry/ShapeItems.sysml
```diff
@@ -0,0 +1,266 @@
...
+	 */
+	abstract item def Shell :> Item, StructuredSpaceObject, Surface;
+
+	/**
+	 * Path is the most general Curve shape.
+	 */
+	abstract item def Path :> Item, StructuredSpaceObject, Curve;
+
+	/**
+	 * A Circle is a Path in the shape of a circle of a given radius.
```
conradbock, line 40: Wouldn't say "shape", a circle has no boundary so has no Item::shape.  Definition is circular (which actually might be appropriate here!), but I guess avoiding it means saying more than is modeled (eg, points a radius distance from one point).

seidewitz: Well, maybe it doesn’t _have_ a shape, but it _is_ a shape — the shape of a disc.

conradbock:
Comment on Item::enclosingShapeConstraint 
on spaceInterior is [0..1] (like boundary), ie, it's a single occurrence (one that unionsOf with the boundary to the entire occurrence). I gather enclosingShapes is a subset of Occurrences this one is Within, but Occurrence doesn't have an inverse of spaceTimeEnclosedOccurrences. But this would give shapes that don't exist for the entire duration of this one. Maybe this is a kind of spaceShot.
What is enclosingShapes used for? If not for anything yet, maybe we should look at it in the next sprint. I don't stay up as late as you do. :) Starting for bed in a half hour or so.
Also the term "enclosing" might not be so good for SysML, it might be taken in the sense of "a watermelon rind encloses the fruit", which isn't 4D Inside (Sandy' comment about this term).

seidewitz:
> Comment on Item::enclosingShapeConstraint

It is supposed to be "enveloping". I didn't make all the changes for this I meant to.

What I want to say is that the Item is inside any of the enveloping shapes, which may be larger than the space boundary. This was easy when the shapes were "solid". It seems to be harder when the shape is a "shell".

conradbock:
Actually we can't right now, neither can RCC.  The rind and fruit are outside each other (ie, disjoint 3/4D points).  We might need to identify a curve that has one end in the fruit, the other end somehow on the other side of the rind, which must intersect the rind. Maybe orientation might help, HP says it can be used to identify the "direction" the outside is on a boundary.

seidewitz:
I think the following essentially works, but it seems harder than necessary.
```
		assert constraint envelopingShapeConstraint { 
			envelopingShapes->forAll { in envelopingShape : Item;
				item envelopingItem {
					item :>> shape = envelopingShape; 
				}
				envelopingItem.spaceSlices->includes(Item::self)
			}
		}
```
That is, for each `envelopingShape`, there is an Item with that shape for which this Item is a `spaceSlice`.

conradbock:
That looks right for solids, assuming the enveloping shapes are supposed to HappensWhile the enveloped ones.  Other shapes that the enveloping ones HappenDuring won't qualify, eg, an engine that in a car, but only for some of the car's life.

conradbock:
Is envelopingShapes needed for something?

seidewitz:
> Is envelopingShapes needed for something?

It is what Hans Peter is try to do in both of his examples. And it is really the only way the have "other shapes" for an Item makes sense once you define "the" shape as the boundary of the Item (for which there is only one). It was also more or less also the way I was defining "shape" in 507-1, for the same reason.

conradbock:
`   envelopingItem.spaceTimeEnclosedOccurrences->includes(Item::self)`
would handle the engine in car only some of the car's life.

seidewitz:
No, the idea would be to, say, have a shape for the car for its entire life that the car would always be "inside" (in the 3D spatial sense), but might be "bigger" than the actual exact boundary of the car.

conradbock:
envelopingItem.spaceTimeEnclosedOccurrences implies InsideOf.

PR #342, opened 2022-03-10
ST6RI-506 Reference frame Kernel Library model

seidewitz:
This pull request adds the following Kernel Library models.

1. `VectorValues` – Provides a basic model of abstract vectors as well as concrete vectors whose components are numerical values.
2. `VectorFunctions` – Defines abstract functions on `VectorValue` corresponding to the algebraic operations provided by a vector space with inner product. It also includes concrete implementations of these functions specifically for `CartesianVectorValue`.
3. `TrigFunctions` – Defines basic trigonometric functions on real numbers and provides a constant feature for `pi`.
4. `SpatialFrame` – Models spatial frames of reference for quantifying the position vector of a point in a three-dimensional space.

It also updates the following Domain Library models.

1. `Quantities` – Makes `VectorQuantityValue` a kind of `NumericalVectorValue`.
2. `VectorCalculations` – Revised to provide calculation definitions for `VectorQuantityValue` that specialize the corresponding functions from `VectorFunctions`.

**Note:** This PR does _not_ include a SysML-level version of`SpatialFrame` with functions specialized to take `TimeInstantValue` and return `VectorQuantityValue`. This may instead be included in the Geometry Domain Library update in PR #507, depending on the resolution of the discussion there.

### DIFF
--- a/kerml/src/examples/Mass Roll-up Example/MassRollup_1.kerml
+++ b/kerml/src/examples/Mass Roll-up Example/MassRollup_1.kerml
@@ -1,5 +1,5 @@
 package MassRollup_1 {
-	import ScalarFunctions::*;
+	import NumericalFunctions::*;
 
 	class MassedThing {
 		feature mass subsets ISQ::mass;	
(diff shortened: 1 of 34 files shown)


seidewitz:
Accidental duplicate of PR #337.

PR #343, opened 2022-03-11
ST6RI-507-1 Initial Basic Geometry Domain Library model for shapes (revision)

seidewitz:
This PR is a proposed revision to the draft geometry model on branch ST6RI-507 as of PR #340. It makes the following changes relative to that branch.

_Adds package CSGShapes_

- Moves `CsgSolid` and its specialization out of the `Shapes` package into a new `CSGShapes` package.
- Removes the `operator` attribute of `CsgCompound` and, instead, defines new `CsgUnion`, `CsgDifference` and `CsgIntersection` specializations. (This makes it easier to specify corresponding specializations of `ReifyShape`; see below.)

_Updates package ShapeCalculations_

- Removes `csgUnion`, `csgDifference` and `csgIntersection` (attribute definition constructors can be used instead).
- Adds the abstract calculation definition `ReifyShape`, which returns an `Object` whose spatial extent is determined by a given `Shape`.
- Adds concrete specializations of `ReifyShape` for `CsgUnion`, `CsgDifference` and `CsgIntersection`.
- Adds placeholder specializations of `ReifyShape` for various specializations of `CsgPrimitive`.

_Adds package SpatialItems_

- A `SpatialItem `is an `Item` with a three dimensional spatial extent that also acts as a `SpacialFrame` of reference. It also specifies a local `Clock` (defaults to the `defaultClock`), a local coordinate system (`VectorMeasurementReference`) and an optional bounding `Shape`.
- A `CompoundSpatialItem` is a `SpatialItem` that is a union of a collection of component `SpatialItems`. Each of its component `SpatialItems` has a local coordinate system that is nested in the coordinate system of the enclosing `SpatialItem` and a local `Clock` that defaults to the `localClock` of the enclosing `SpatialItem`. If all the component `SpacialItems` have `boundingShapes`, then the `boundingShape` of the enclosing `SpatialItem` defaults to the union of the `boundingShapes` of the component `SpatialItems`.
- `PositionOf`, `CurrentPositionOf`, `DisplacementOf` and `CurrentDisplacementOf` are specialized for `SpatialItem` to return a `VectorQuantityValue` in the coordinate system of the `SpatialItem` and with the default `Clock` being the `localClock` of the `SpatialItem`.

**Notes**

1. The specializations of `ReifyShape` in `ShapeCalculations` for the `CsgPrimitives` need to be tied to an eventual model of Objects corresponding to these primitive shapes, perhaps as `SpatialItems` whose `boundingShape` is the corresponding `CsgPrimitive`.
2. Something like `SpatialItem` is needed whether or not we keep the attribute-definition `Shape` models, to tie back to the `SpatialFrame` semantics.
3. An alternative to using `Shape` specifications might be to just have `CompoundSpatialItems`, for all of union, difference and intersection, and `PrimitiveSpatialItems` for the primitive shapes, forgoing the infix operators for now. This would depend on how convenient it would be for a modeler to easily construct complicated shapes.


sysml.library/Domain Libraries/Geometry/ShapeCalculations.sysml
```diff
@@ -1,33 +1,79 @@
...
+		return unionObject : Object {
+			:>> unionsOf[1] {
+				:>> elements = shape.operands->collect{in s; ReifyShape(s)};
+			}
+		}
+	}
+	
+	/**
+	 * Return an Object whose special extent is determined by a CsgDifference shape.
+	 */
```
hpdekoning, line 51: typo: spatial extent

sysml.library/Domain Libraries/Geometry/ShapeCalculations.sysml
```diff
@@ -1,33 +1,79 @@
...
+			:>> differencesOf[1] {
+				:>> elements = shape.operands->collect{in s; ReifyShape(s)};
+			}
+		}
+	}
+	
+	/**
+	 * Return an Object whose special extent is determined by a CsgIntersection shape.
+	 */
+	calc def ReifyCsgIntersection :> ReifyShape (shape : CsgIntersection) {
```
hpdekoning, line 63: typo: spatial extent

sysml.library/Domain Libraries/Geometry/SpacialItems.sysml
```diff
@@ -0,0 +1,152 @@
+/**
```
hpdekoning, line 1: Filename typo: SpatialItems.sysml

sysml.library/Domain Libraries/Geometry/SpacialItems.sysml
```diff
@@ -0,0 +1,152 @@
...
+	private import UnitsAndScales::VectorMeasurementReference;
+	private import Time::Clock;
+	private import Time::TimeInstantValue;
+	private import ISQ::Cartesian3dSpatialCoordinateSystem;
+	private import ControlFunctions::forAll;
+	private import ControlFunctions::reduce;
+	
+	/**
+	 * A SpatialItem is an Item with a three dimensional spatial extent that also acts as a SpacialFrame of reference.
+	 */
```
hpdekoning, line 23: typo: SpatialFrame

sysml.library/Domain Libraries/Geometry/SpacialItems.sysml
```diff
@@ -0,0 +1,152 @@
...
+		 * and any componentItems. By default this is the global defaultClock.
+		 */
+		item localClock : Clock[1] default Time::defaultClock;
+		
+		/**
+		 * The VectorMeasurementReference to be used as the measurement reference for position and displacement
+		 * vector values relative to this SpatialItem.
+		 */
+		attribute coordinateSystem : VectorMeasurementReference[1] default Cartesian3dSpatialCoordinateSystem() {
+			attribute :>> dimensions = 3;
```
hpdekoning, line 36: On new usages we should use `coordinateFrame`. I will raise a JIRA issue to rename all uses of CoordinateSystem in the libraries.

sysml.library/Domain Libraries/Geometry/SpacialItems.sysml
```diff
@@ -0,0 +1,152 @@
...
+	private import Time::Clock;
+	private import Time::TimeInstantValue;
+	private import ISQ::Cartesian3dSpatialCoordinateSystem;
+	private import ControlFunctions::forAll;
+	private import ControlFunctions::reduce;
+	
+	/**
+	 * A SpatialItem is an Item with a three dimensional spatial extent that also acts as a SpacialFrame of reference.
+	 */
+	item def SpatialItem :> SpatialFrame {
```
hpdekoning, line 24: Should probably be an `abstract item def`

hpdekoning:
I like the replacement of `BooleanOperatorKind` with the attribute defs for each operation, much cleaner.

For the Reify... functions I think it would be possible and very interesting to have a _lazy_ and a _full_ variant where the lazy variant just stores the recipe and the full variant (or an additional function) computes - on demand - the evaluated final shapes / boundaries. This may keep the whole capability lightweight while having rigorous semantic / logical representation when needed.

In branch ST6RI-507-2 I was adding a generic Boundary Representation shape, which is what STEP-42 models as a connected faceset that is both a geometric and topological representation. This can also be used - and is used in many 3D editors, visualizers and standards - to represent the fully evaluated end result of an arbitrarily complex CSG recipe.

As an implementation strategy we could at first allow planar faces only - what STEP-42 calls a faceted_brep - and later allow also any curved faces, typically defined through NURBS.

seidewitz:
> For the Reify... functions I think it would be possible and very interesting to have a _lazy_ and a _full_ variant ..

That is an interesting implementation idea, but I don't think we handle the complexity of that in our model. But we should try not to do something that might prevent a tool from using a lazy-evaluation implementation.

sysml.library/Domain Libraries/Geometry/ShapeCalculations.sysml
seidewitz: Fixed.

sysml.library/Domain Libraries/Geometry/ShapeCalculations.sysml
seidewitz: Fixed.

sysml.library/Domain Libraries/Geometry/SpacialItems.sysml
seidewitz: Fixed.

sysml.library/Domain Libraries/Geometry/SpacialItems.sysml
seidewitz: Updated.

seidewitz:
I have updated `ShapeDefinitions` and `ShapeCalculations` on the ST6RI-507-1 branch to work directly on `Items` and `ShapeItems` as merged from the ST6RI-507 branch (PR #340). I am closing this pull request, but keeping the branch for possible future use.

PR #345, opened 2022-03-29
ST6RI-510: Improved Interconnection View

himi:
By this PR, interconnection view makes use of the directions of connections and flow connections to give better rendering results.  For example:

```
package TestConnection0 {
    part a1 {
        port p1;
    }
    part a2 {
        port p2;
    }
    flow from a1.p1 to a2.p2;
}
```
gives
Screen Shot 2022-03-29 at 2 22 31 PM

This PR also renders specializations in interconnection views.

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/SysML2PlantUMLText.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/SysML2PlantUMLText.java
@@ -422,6 +422,7 @@ public class SysML2PlantUMLText {
     private int idCounter;
     private IDMap idMap;
 
+    // element 1<-* id 1<-* path(feature, featureChain, featureChainExpression, ItemFlowEnd)
     private class IDMap {    	
         private final Map<Element, Integer> idMap = new HashMap<Element, Integer>();
         private final IDMap prev;
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VActionMembers.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VActionMembers.java
@@ -36,7 +36,7 @@ import org.omg.sysml.lang.sysml.Type;
 public class VActionMembers extends VBehavior {
 
     private void addNode(Feature f, String type) {
-        String name = getNameAnyway(f, true);
+        String name = getNameAnyway(f);
         append(type);
         append(' ');
         addNameWithId(f, name, true);
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VBehavior.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VBehavior.java
@@ -139,7 +139,7 @@ public abstract class VBehavior extends VDefault {
         	}
         }
         if (text.length() == 0) {
-            String name = getNameAnyway(au, true);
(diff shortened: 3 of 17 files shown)


himi:
Previously, the example above is rendered as:
Screen Shot 2022-03-29 at 2 24 39 PM

seidewitz:
> Screen Shot 2022-03-29 at 2 22 31 PM

Is there any way to get PlantUML to render `a1` and `a2` further apart, so `p1` and `p2` don't overlap?

himi:
I'm now tuning a layout parameter but I'm not so sure it works.

seidewitz:
Actually, I realized a bigger problem. This is not the correct notation for flows on an interconnection diagram. The source output and target input of a flow should not be ports, they should be features of ports:
```
package TestConnection0 {
    part a1 {
        port p1 {
            out x;
        }
    }
    part a2 {
        port p2 {
            in y;
        }
    }
    flow from a1.p1.x to a2.p2.y;
}
```
The flow connection is actually between the ports `a1.p1` and `a2.p2`, but the transfer happens between `x` and `y`. The expected graphical notation for a flow connection on an interconnection diagram is like an item flow in SysML v1: an undirected line between the ports (like a regular connection), with a triangle on it to show the transfer. 

The arrow notation for flows is only used on action diagrams when drawn between action parameters, which are like features on ports, _not_ ports themselves. I suppose you could also use an arrow for flows similarly on an interconnection diagram, but only if you could show the nested features on the ports and draw the arrow between them.

himi:
I investigated why ports having parameter were not properly rendered and found FeatureMembership.memberFeature is not a derived feature of Membership.memberElement.  Rather, in the current implementation, memberFeature seems to be a distinct feature.

Also if targets of flow connection must not point to ports, those should be invalidated, I believe.

Anyway, can we deal with this issue in the separate JIRA issue?  I think we need to discuss it in Graphical Specification WG as well.

seidewitz:
> Also if targets of flow connection must not point to ports, those should be invalidated, I believe.

Yes, there are a lot of validations that have not been implemented yet!

> Anyway, can we deal with this issue in the separate JIRA issue? I think we need to discuss it in Graphical Specification WG as well.

I am not sure it is worth "improving the interconnection view" for flow connections when the notation is wrong, anyway. The PlantUML visualization can already show features nested within ports (see the visualization of `training/13. Flow Connections/Streaming Example.sysml`, for example). It would be great if the arrow notation could at least be shown between the features in such cases.

How does this PR affect the visualization of non-flow connections?

himi:
Unfortunately, in the current situation, any changes related to feature reference are problematic because they would make a big impact to ST6RI-529, dealing with inheritance.

Regarding the second question, yes, heuristically this change regards the first end as input and the rest as output.  In many cases, this improvement works.  For example, `02-Parts Interconnection` validation cases will give better results.

If it's not acceptable, please close this PR.  I'll include these change to ST6RI-529.

himi:
Maybe I can do a small modification to change the style of flow connection.  I'm now trying.

himi:
I thought I could change the style of flow connection but it would make another side effect to action view.  So I decided to work on the further development after ST6RI-529 because it will be a fundamental change in the infrastructure of the visualizer.

seidewitz:
This update is now included in PR #348 with ST6RI-529.

PR #347, opened 2022-04-01
ST6RI-520 Update implementation for 2022-03 baseline

seidewitz:
This pull requests implements significant updates to the abstract syntax for annotations and metadata, per the 2022-03 baseline metamodel. 

**Not Implemented**
1. The proposed semantic metadata/language extension mechanism. (This will be address in a future PR for ST6RI-521.)
2. Setting the value of the `annotatedElement` value(s) of a MetadataFeature instance. (This is potentially deferred until the full implementation of reflection for ST6RI-417.)

**Abstract Syntax**

_Metaclasses removed:_

1. Documentation (relationship)
3. AnnotatingFeature (renamed to MetadataFeature)
4. MetadataFeature
5. MetadataFeatureValue

_Metaclasses added:_

1. Documentation (as a subclass of Comment)
2. MetadataFeature (replaces AnnotatingFeature)
3. Metaclass
4. MetadataDefinition
6. MetadataUsage

_Metaclasses updated:_

1. AnnotatingElement – the implementation of `AnnotatingElement::annotedElement` has been updated so that, if the AnnotatingElement has not explicit Annotation relationships, the `owner` of the AnnotatingElement is returned as the `annotatedElement` (e.g., the `owningNamespace`, if the AnnotatingElement is an `ownedMember` of a Namespace).
2. Definition – no longer abstract. (This is necessary for the planned language extension mechanism. However, there is no concrete syntax yet for a Definition that is not of some more specialized type.)
3. Usage – no longer abstract. (This is necessary for the planned language extension mechanism. However, there is no concrete syntax yet for a Usage that is not of some more specialized type.)

_Validation:_

1. A MetadataFeature (formerly AnnotatingFeature) is equired to have a concrete Metaclass as its type. A MetadataUsage (subclass of MetadataFeature) is required to have a concrete MetadataDefinition (subclass of Metaclass) as its type.
2. If a MetadataFeature (or MetadataUsage) (concretely) subsets/redefines the `Metaobjects::annotatedElement` feature, then each of the `annotatedElements` of the MetadataFeature must conform to the types of one of the subsetting/redefining features.
3. Each owned Feature of a MetadataFeature (or MetadataUsage) must pass the following validation checks:
    - It must redefine a Feature (owned or inherited) of a supertype of its owning Feature.
    - If it has a FeatureValue, the `value` Expression must be model-level-evaluable.
    - Each of its owned Features (if any) must also pass these validation checks.

**Concrete Syntax**

[...]

PR #349, opened 2022-04-07
ST6RI-521 Keyword metadata for language extension

seidewitz:
This pull request adds a mechanism that allows "semantic metadata" annotations to be used to specify implicit specialization of declared types. It also provides a new keyword notation for annotations in the SysML grammar.

1. _Semantic metadata._ The `SemanticMetadata` metaclass was included in the Kernel `Metaobjects` package added in PR #347. It is intended to be specialized with its `baseType` feature bound to a specific reflective `Type` metaobject value. When a specialization of `SemanticMetadata` is used to annotate an appropriate type declaration, the specified `baseType` is used to implicitly specialize the annotated type.
   - If the annotated type is neither a Classifier nor a Feature, then the annotated type implicitly specializes the `baseType`.
   - If the annotated type is a Classifier and the `baseType` is a Classifier, then annotated classifier implicitly subclassifies the `baseType`.
   - If the annotated type is a Classifier and the `baseType` is a Feature, then the annotated classifier implicitly subclassifies each type of the `baseType`.
   - If the annotated type is a Feature and the `baseType` is a Feature, then the annotated feature implicitly subsets the `baseType`.
   - In all other cases, no implicit specialization is added.

2. _Casting._ The cast operator **`as`** has been added to the implemented operators for model-level evaluable expressions. In addition, if the target type of the case is a metaclass, and the argument expression evaluates to a feature whose abstract syntax metaclass conforms to the given metaclass, then the result is a "meta-upcast" to the metaobject representation of the feature. (E.g., if `vehicle` is a part usage, then `vehicle`**`as`**`KerML::Feature` evaluates, as a model-level evaluable expression, to a MetadataFeature for `vehicle` with type `SysML::PartUsage`.)

   **Note:** The 2022-02 baseline did not add an expression for obtaining the metaclass of an element. Using the cast operator is a workaround at this time. It means that, currently, it is only possible to specify a feature as a `SemanticMetadata::baseType`. However, this is sufficient to also define implicit specializations for classifiers, as described above.

3. _Keywords._ A _user-defined keyword_ is a (possibly qualified) metaclass/metadata definition name or human ID preceded by the symbol `#`. Such a keyword can be used in the SysML textual notation in package, dependency, definition and usage declarations. The user-defined keyword is placed immediately before the language-defined (reserved) keyword for the declaration and specifies a metadata annotation of the declared element. For example, if `SafetyCritical` is a visible metadata definition, then `#SafetyCritical`**`part`**`brakes;` is equivalent to **`part`**`brakes { @SafetyCritical; }`

    **Notes:**
    -  User-defined keywords are not currently available in the KerML notation.
    -  If the given metaclass or metadata definition is a kind of `SemanticMetadata`, then the implicit specialization rules given above apply.
    - It is not possible to specify nested features for a metadata feature annotation in the keyword notation.

4. _Extended definitions and usages._ The metaclasses Definition and Usage are no longer abstract in the 2022-03 baseline abstract syntax. This allows the possibility of declaring a definition or usage without a language keyword identifying a more specialized kind, as long as a user-defined keyword is given to specify `SemanticMetadata` to be used for implicit specialization. For example, if `provider` is the human ID of `SemanticMetadata` whose `baseType` is the feature `serviceProviders : ServiceProvider`, then `#provider`**`def`**`BankingService;` parses to a Definition with an implicit subclassification of `ServiceProvider` and `#provider bankingService;` parses to a Usage with an implicit subsetting of `serviceProviders`.


tool-support/syntax-highlighting/vscode/sysml/syntaxes/sysml.tmLanguage.json
```diff
@@ -18,11 +18,11 @@
...
         { "match": "(%|&|&&|\\?|\\?\\?|\\^|\\^\\^|\\||\\|\\|)", "name": "keyword.operator.logical.sysml" },
         { "match": "(!|!=|\\<|\\<=|=|==|\\>|\\>=)", "name": "keyword.operator.comparison.sysml" },
         { "match": "(\\*|\\*\\*|\\+|\\-|/)", "name": "keyword.operator.arithmetic.sysml" },
-        { "match": "(\\-\\>|\\.\\.|:|::|:=|:\\>|:\\>\\>|=\\>|@|~)", "name": "keyword.operator.other.sysml" }
+        { "match": "(\\-\\>|\\.$|\\.\\.|\\.\\?|:|::|:=|:\\>|:\\>\\>|=\\>|@|~)", "name": "keyword.operator.other.sysml" }
```
hpdekoning, line 25: The `$` char should be escaped as `\\$`. Small adaptation needed in `tool-support/syntax-highlighting/xtext_grammar_converter.py` line 409, to properly recognize the `.$` operator, as `$` is end of test string symbol in regex.

hpdekoning (approved):
One little fix in syntax-highlighting. Further looks great to me.

himi:
I'm confused that the explanation of `as` operator.  If we write `vehicle as KerML::Feature`, I think it should be upcast to `KerML::Feature` rather than `SysML::PartUsage`. Or am I misunderstand it?   In my understanding, `as` operator is needed to resolve diamond problems in multiple inheritance. Is that right?

himi (approved):
I tested this version and so far I do not find any problems in the implementation while the visualization is not appropriate.  I would like to update the visualizer soon in another PR soon but I think it will not be in this sprint.

seidewitz:
> I'm confused that the explanation of as operator. 

The static type of an expression like `x``**`as`**`KerML::Feature` is KerML::Feature, and that is what is used for static type checking. However, dynamically the expression evaluates to the _actual_ meta object for `x` (as long as that is a subtype of KerML::Feature). Thus, if `x` is a part usage, then the result will be an instance of SysML::PartUsage, which is an (indirect) subtype of KerML::Feature. 

Other than the meta-shift, this is similar to normal up-casting, in which the dynamic runtime type of the result may be a subtype of the static type given in the cast. The important thing is that this is consistent with the static type checking.

himi:
Oh, I see. `as` actually allows incompatible cast to pass static type checking.  Thanks.

tool-support/syntax-highlighting/vscode/sysml/syntaxes/sysml.tmLanguage.json
seidewitz: We actually decided use `.?` rather than `.$` in the expression syntax. The appearance of `.$` here is do to this not being properly updated in the SysML.xtext grammar for `TargetExpression`. Good that we found this! I will fix it on this branch.

PR #352, opened 2022-04-08
ST6RI-531: Non-ASCII characters may not be parsed properly in Jupyter

himi:
This PR intends to fix the bug that the SysML kernel may wrongly load model library files in UTF-8 encoding.

We have two approaches to fix this bug: one is that explicitly specifying UTF-8 encoding to load SysML files by using something like `XtextResource.setEncoding`; the other is that system's default file encodings with System property.

However, the first approach  would require to modify more codes related to Xtext so that the fix and test should be more complicated.  So I decided to take the second approach in this PR because we confirmed this problem did not occur under the environment where the default encoding is UTF-8.

In this fix, I propose to extract SysML kernel initialization part to the separate method `initialize()` because the most part of `main()` comes from IJava implementation (see https://github.com/SpencerPark/IJava/blob/master/src/main/java/io/github/spencerpark/ijava/IJava.java for details).  I also made it public to allow other modules to do `initialize()` part with different Jupyter connections.

I also made the getter method as `SysMLKernel.getMagics()` for the similar purpose to test and extend the kernel while it is optional in this PR.

### DIFF
--- a/org.omg.sysml.jupyter.kernel/src/main/java/org/omg/sysml/jupyter/kernel/ISysML.java
+++ b/org.omg.sysml.jupyter.kernel/src/main/java/org/omg/sysml/jupyter/kernel/ISysML.java
@@ -35,10 +35,17 @@ public class ISysML {
 
     private static volatile SysMLKernel kernel;
 
+    public static void initialize() {
+        System.setProperty("file.encoding", "UTF-8");
+        kernel = new SysMLKernel();
+    }
+
     public static void main(String[] args) throws Exception {
         if (args.length < 1) {
             throw new IllegalArgumentException("Missing connection file argument");
         }
+
+        initialize();
 
         Path connectionFile = Paths.get(args[0]);
 
@@ -52,8 +59,6 @@ public class ISysML {
         KernelConnectionProperties connProps = KernelConnectionProperties.parse(contents);
         JupyterConnection connection = new JupyterConnection(connProps);
 
-
-        kernel = new SysMLKernel();
         kernel.becomeHandlerForConnection(connection);
 
         connection.connect();
--- a/org.omg.sysml.jupyter.kernel/src/main/java/org/omg/sysml/jupyter/kernel/ISysML.java
+++ b/org.omg.sysml.jupyter.kernel/src/main/java/org/omg/sysml/jupyter/kernel/ISysML.java
@@ -36,7 +36,6 @@ public class ISysML {
     private static volatile SysMLKernel kernel;
 
     public static void initialize() {
-        System.setProperty("file.encoding", "UTF-8");
         kernel = new SysMLKernel();
     }
 
--- a/org.omg.sysml.jupyter.kernel/src/main/java/org/omg/sysml/jupyter/kernel/SysMLKernel.java
+++ b/org.omg.sysml.jupyter.kernel/src/main/java/org/omg/sysml/jupyter/kernel/SysMLKernel.java
@@ -45,6 +45,10 @@ public class SysMLKernel extends BaseKernel {
     private final Magics magics;
     private final MyMagicParser magicParser;
 
+    public Magics getMagics() {
+        return magics;
+    }
+
     public SysMLKernel() {
         this.interactive = SysMLInteractive.getInstance();
         Optional<String> libraryPath = Optional.ofNullable(System.getenv(ISysML.LIBRARY_PATH_KEY));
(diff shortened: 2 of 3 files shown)


seidewitz:
I considered the alternative of putting `System.setProperty("file.encoding", "UTF-8");` in `SysMLInteractive::createInstance` or `SysMLInteractive::loadLibrary`, rather than in `ISysML`. The advantage of this is that it would deal with the issue for potential uses of `SysMLInteractive` for other integrations in the future. The disadvantage is that creating a `SysMLInteractive` instance would, "behind the scenes", possibly change the file encoding for the running JVM, which might be unexpected.

Also, have you confirmed that the proposed change actually resolves the problem? I have seen it noted in some on-line discussions that the InputStreamReader class caches the file encoding on JVM start up, and that changing the system property after that does not necessarily have any effect on reading input streams.

himi:
At least I confirmed it worked on Windows 11, OpenAdoptJDK11, Japanese Locale (more specifically CP932 local encoding), which had not passed the test.
On MacOS 12.2.1 and aarch64 JDK11, no problem had occurred.
I'm now testing it on Ubuntu Linux...

himi:
I have no objection to improve `SysMLUtil` by explicitly specifying `ENCODING`.  Still it will be safer to set the default file.encoding to "UTF-8" because most  of our environment is actually using that that default encoding.

seidewitz:
I updated `SysMLUtil` to set the encoding for Xtext resources, but also left the setting of `file.encoding` in `ISysML`, as you suggested. However, I haven't tested it. Since you are already set up to test the encoding, could you test that this update to `SysMLUtil` would actually solve the problem, even _without_ the change to `ISysML`? This could be important for other future users.

himi:
I see. I'll test it without setting `file.encoding` property this night.

seidewitz (approved):

seidewitz:
@ivan-gomes 
Please review this PR as soon as possible.

org.omg.sysml.jupyter.kernel/src/main/java/org/omg/sysml/jupyter/kernel/ISysML.java
```diff
@@ -35,11 +35,18 @@
 
     private static volatile SysMLKernel kernel;
 
+    public static void initialize() {
+        System.setProperty("file.encoding", "UTF-8");
```
ivan-gomes, line 39: > fixed this problem even without `file.encoding` fix in Windows 11, CP932 environment

Given this, I would ask that we consider reverting the changes in `ISysML` as I agree with the general sentiment that the Jupyter adaptation should be the thinnest layer required to interoperate with the real work being done on the Eclipse side.

org.omg.sysml.jupyter.kernel/src/main/java/org/omg/sysml/jupyter/kernel/SysMLKernel.java
```diff
@@ -45,6 +45,10 @@
     private final Magics magics;
     private final MyMagicParser magicParser;
 
+    public Magics getMagics() {
```
ivan-gomes, line 48: I don't believe this new public accessor is being used anywhere.

ivan-gomes (changes requested):

org.omg.sysml.jupyter.kernel/src/main/java/org/omg/sysml/jupyter/kernel/SysMLKernel.java
himi: @ivan-gomes, actually, I'm using these changes for other modules to test and use the original SysMLv2 kernel, and I showed these to you at Long beech meeting while we can revert these changes because the issue is solved without these changes.  But I'm not sure other problem may occur in non UTF-8 default encoding environment.

ivan-gomes: Gotcha, in that case I will withdraw my review comment. I double checked that the `Magics` class only allows registering new magics, but not removing already registered ones, including the ones we specify, which was my initial concern with exposing it if there wasn't a use case for it.

org.omg.sysml.jupyter.kernel/src/main/java/org/omg/sysml/jupyter/kernel/ISysML.java
seidewitz: Let’s remove the setting of `file.encoding`, but leave the `initialization` method.

himi: I see.  It's done.

ivan-gomes (approved):

PR #354, opened 2022-04-09
ST6RI-508 Nested coordinate frames for composite structure

hpdekoning:
This PR adds `CoordinateFrame` as a specialization of `VectorMeasurementReference`, an abstract CoordinateTransformation (attribute def) and three concrete specializations of `CoordinateTransformation`, namely `CoordinateFramePlacement`, `TranslationRotationSequence`, `AffineTransformationMatrix`, to the `UnitAndScales` domain library.

A fourth transformation by quaternions will be captured in a JIRA issue for future implementation after `Quaternion` is added to `ScalarValues.kerml`.

An example on use of the coordinate frame transformations and nesting is given in:
`VehicleGeometryAndCoordinateFrames.sysml` in `sysml/src/examples/Geometry Examples`. Also example `CarWithShapeAndCSG.sysml` has been updated with quantified shapes for a simplistic geometric model of the engine as a rectangular box with two cylindrical holes, using the currently available CSG construct via `differenceOf`.

There remain a number of points to be resolved in a next issue:

- Currently the spatial extent of a composite structure of Items and/or Parts is modeled via `SpatialItem` and `CompoundSpatialItem`. This is inconvenient as during modeling it is not known a priori whether an Item or Part will be a leaf node in a composite structure. The modeler should not need to change the type of an item/part when adding more detail / depth to a model. So there should be a single concept of `SpatialItem` or `Item` with shape and coordinate frame that can be used for both leaf and non-leaf nodes in a composite structure.
- The shape of an Item is currently its `spaceBoundary`, which is a 2D surface. However CSG is really defined on 3D body operands (or "solids"  in CAD / 3D modeling vernacular). The boundary representation (B-rep solid in CAD / 3D modeling vernacular) is an efficient and topologically well-defined way to specify the point set that makes up a body (i.e. 3D spatial extent). In my opinion the best would be to first align the KerML concept of spatial extent to B-rep body with outer shell is its space boundary as well as with optional (inner) voids, and then elaborate / update the SysML shape / spatial item implementation.


seidewitz (changes requested):
It is probably too late to fully review this for inclusion in 2022-03, but, on first look, at least the following need to be fixed.

1. The change to `SpatialItem` doesn't work. See [comment on ST4MD-382](https://openmbee.atlassian.net/browse/ST4MD-382?focusedCommentId=13422).
2. Regarding 

   `assert constraint { source.dimensions = origin.dimensions } // Problem: source does not resolve`

   The problem is a syntax error, not a name resoluton error. Change `=` to `==` and it will parse.
4. Do not include `TranslationRotationQuaternion` if it is not complete. We should strive not to put any more "TODOs" into the library models and keep them in a form that could be potentially included in the submission without further changes. Proposed additional work should be tracked in Jira, not with "TODOs" at this point.

himi:
It looks it takes an approach to explicitly specify a source `coordinateFrame` for nested coordinate transformation.  However, in `VehicleGeometryAndCoordinateFrames`, `vehicle.chassis.coordinateFrame.transformation` does not specify it but the other parts such as *wheels do that.  I guess it should specify `nestedCoodinateFrame` by subsetting `datum` if I understand you intention correctly.

himi:
And may I ask to merge the latest master for further review?  This version is too old and painful for me to evaluate.

hpdekoning:
"It is probably too late to fully review this for inclusion in 2022-03"

Yes, agreed. In my assessment the current library is really complicated to use and explain to normal users. I really think we need to find ways to simplify, at least the user experience.

"The change to SpatialItem doesn't work. See [comment on ST4MD-382](https://openmbee.atlassian.net/browse/ST4MD-382?focusedCommentId=13422)."

I meant to add `self` to the union but could not find expression syntax to do it. Since it is a `Set` I am assuming that adding an item more than once to the set is idempotent. I think then `unionsOf` would represent the full spatial extent defined by a complete assembly tree. The redefinition of `unionsOf` could also be named `compositeSpatialExtent` or something similar (instead of `componentUnion`). Could we do something like:
```
    attribute compositeSpatialExtent[1] :> unionsOf {
        item :>> elements = self + spatialSubitems;
    }
```
or
```
    attribute compositeSpatialExtent[1] :> unionsOf {
        item :>> elements = self, spatialSubitems;
    }
```

seidewitz:
> I really think we need to find ways to simplify, at least the user experience.

This is beyond the scope of ST6RI-508, and this PR is not the best place to discuss changes beyond the implementation of 508. Let’s focus first on getting the coordinate frame update completed, and then we can separately discuss and prioritize further changes.

hpdekoning:
> It looks it takes an approach to explicitly specify a source `coordinateFrame` for nested coordinate transformation. However, in `VehicleGeometryAndCoordinateFrames`, `vehicle.chassis.coordinateFrame.transformation` does not specify it but the other parts such as *wheels do that. I guess it should specify `nestedCoodinateFrame` by subsetting `datum` if I understand you intention correctly.

In the updates the `source` coordinate frame was added in all example usages. In the future I hope to add a default value to `source` being the `coordinateFrame` of the _owning item usage_ of an _owned composite item usage_ (i.e. one level down in the composite structure). For this we need something like `owningItem : Item[0..1]` the inverse of `subitems : Item[*]` in the specification of `Items::Item`.

hpdekoning:
> And may I ask to merge the latest master for further review? This version is too old and painful for me to evaluate.

One of the last master commits before the 2022-03 release was merged into this branch to check consistency and ease review.

sysml.library/Domain Libraries/Quantities and Units/UnitsAndScales.sysml
```diff
@@ -304,7 +426,7 @@ package UnitsAndScales {
 	}
 
 	/**
-	 * Declare a constraint to verify that the given unit power factors comply with the required quantity dimension
+	 * Constraint definition to verify that the given unit power factors comply with the required quantity dimension
 	 *
 	 * TODO: The implementation of this constraint requires reflection.
```
conradbock, line 431: Probly should remove TODOs.

conradbock:
In UnitsAndScales, probly should remove the TODO constraint at the bottom.

hpdekoning:
The TODO for `TranslationRotationQuaternion` has been removed. Issues https://openmbee.atlassian.net/browse/ST6RI-535 and https://openmbee.atlassian.net/browse/ST6RI-536 have been created to capture the need.

conradbock (approved):

sysml.library/Domain Libraries/Geometry/SpatialItems.sysml
```diff
@@ -81,18 +59,15 @@ package SpatialItems {
...
+		item componentItems : SpatialItem[0..*] ordered :> subitems {
 			item nestedLocalClock :>> localClock default localClock;
 			attribute nestedBasisDirections : VectorQuantityValue[3];
 			attribute nestedOrigin : VectorQuantityValue[0..1];
 			attribute nestedCoordinateFrame :>> coordinateFrame {
 				attribute :>> transformation[1] {
 				 	attribute :>> source = coordinateFrame;
 				 	attribute :>> target = nestedCoordinateFrame;
-				 	//attribute :>> basisDirections = nestedBasisDirections;
-				 	//attribute :>> origin = nestedOrigin;
```
seidewitz: If `basisDirections` and `origin` are removed from `nestedCoordinateFrame`, then `nestedBasisDirections` and `nestedOrigin` need to be removed from `componentItems`. The latter were only there to make it easier to set the former.

sysml.library/Domain Libraries/Quantities and Units/UnitsAndScales.sysml
```diff
@@ -1,37 +1,52 @@
 /**
  * This package defines the representations for measurement references.
- * TODO: Propose to rename to MeasurementReferences
+ * TODO: Rename package to MeasurementReferences as this name captures the content much better
```
seidewitz: Please do not put "TODOs" in library packages. Any issues or proposals should be recording in Jira for potential scheduling in a sprint.

sysml.library/Domain Libraries/Quantities and Units/UnitsAndScales.sysml
seidewitz: I suppose we can leave this one "TODO" in here for now, since it marks an incomplete definition that should be completable before we are done. But if it does not get done before submission, then the whole definition will need to be removed.

sysml/src/examples/Geometry Examples/VehicleGeometryAndCoordinateFrames.sysml
```diff
@@ -1,33 +1,77 @@
...
 
-    part def Wheel :> SpatialItem {
+	/**
+	 * Generic wheel with lugbolts
+	 * 
+	 * The radius is estimated for a 22 inch hub plus 110 mm tire height.
+	 * The wheel width is equal to the cylinder height.
+	 * The wheel has 5 lugbolts that are evenly distributed along a circle centered at the wheel's center.
+	 */
+    part def Wheel :> CompoundSpatialItem {
```
seidewitz: Something should not be a `CompoundSpatialItem` unless it actually has at least one `componentItem`, because a `CompoundSpatialItem` is _made up_ of its `componentItems`. I now think `componentItems` really needs to have multiplicity 1..* to make this clear.

sysml/src/examples/Geometry Examples/VehicleGeometryAndCoordinateFrames.sysml
```diff
@@ -42,45 +86,46 @@ package VehicleGeometryAndCoordinateFrames {
         part chassis : Chassis[1] :> componentItems {
 			attribute :>> coordinateFrame {
```
seidewitz: Better to use `nestedCoordinateFrame`, as `coordinateFrame` is redefined for `componentItems`. This is a diamond inheritance issue and it may not otherwise pass validation in the future.

sysml/src/examples/Geometry Examples/VehicleGeometryAndCoordinateFrames.sysml
```diff
@@ -42,45 +86,46 @@ package VehicleGeometryAndCoordinateFrames {
...
         private attribute plusXAxis : Array { :>> dimensions = 3; :>> elements : Real[3] = (1, 0, 0); }
-        private attribute xFrontWheelShift : Real = 1670;
-        private attribute xRearWheelShift : Real = -1820;
-        private attribute yWheelShift : Real = 720;
+        private attribute frontWheelXShift : Real = 1670;
+        private attribute rearWheelXShift : Real = -1820;
+        private attribute wheelYShift : Real = 720;
 
         part leftFrontWheel : Wheel[1] {
             attribute :>> coordinateFrame {
```
seidewitz: Missing `:> componentItems`.

seidewitz:
I have pushed a commit that resolves the additional comments I posted above. I have also copied the updated library models to the library directories used by the KerML and SysML Xpect tests (these unfortunately still need to be copied by hand).

With these changes, I can approve this PR.

seidewitz (approved):

sysml/src/examples/Geometry Examples/VehicleGeometryAndCoordinateFrames.sysml
hpdekoning: I see why it has to be like this with the current SpatialItem and CompoundSpatialItem setup, and I think we should keep it like this for now. But it is a quite complicated structure that most end-users will find extremely difficult to grok and use. 

So, I feel we need to raise a new issue to make a concerted effort to simplify how to model composite structure with spatial extent in SysML in a way that is similar / analogous to a CAD / 3D modeling assembly tree with leaf parts, without the necessity to change item/part type and coordinateFrame attribute when a leaf of the tree is changed into a (sub)assembly with more detailed lower level parts. This is needed to evolve a physical architecture e.g. from conceptual through preliminary to detailed design or to 'defeature' a 3D model in the other direction from detailed to approximate / lower fidelity. Or indeed to enable a single model supporting different levels of detail / fidelity, perhaps using variants.

hpdekoning:
Upon rereading definitions of `innerSpaceDimension` and `outerSpaceDimension` corrected minor typos.

OK, to merge for me.

sysml/src/examples/Geometry Examples/VehicleGeometryAndCoordinateFrames.sysml
seidewitz: Agreed that there should be a new issue, probably in Track 4, if we don't already have an issue to cover this there.

PR #355, opened 2022-04-12
ST6RI-532 Minor fixes to space modeling & shape library

conradbock:
**Occurrences.kerml**

- `spaceShotOf` multiplicity: [1..] to [0..*].  Some occurrences aren't snapshots.

- invariants:
-- `innerSpaceDimension`: removed redundancy with feature type.
-- `spaceBoundaryOf`: isClosed to !isClosed
-- `spaceBoundary`: moved to after spaceBoundaryOf.
-- `SpaceShotOf`: Added that innerSpaceDimenion must be strictly lower (compare to SpaceSliceOf), not sure how it got left out.

- `spaceBoundary` description: reworded to avoid "of which".

- Typo in multiple places ("occcurrence" -> "occurrence").

**Objects.kerml**

`genus`: changed type Integer to Natural, simplified invariant.

**ShapeItems.sysml**

- Added `PlanarCurve`, to match PlanarSurface and used it below.

- Added multiplicities to quantification attributes (length, etc).

- `PlanarSurface::shape` typed by PlanarCurve.

- `Line` and `Circle` specialize PlanarCurve (moved Line after PlanarCurve/Surface).

- `Disc` specializes PlanarSurface.

- `Circle::edges::circumference`: Added "* 2".

- `Cone`: Restored radius (was being redefined), bound baseRadius to it.

- Corrected some cell multiplicities in Cone and Cuboid (along with invariant on Cone:ae).

- Added `genus = 0` to Sphere, Cone, and Cuboid.

- Removed redundant feature names when redefining as the same name.

- Moved multiplicities to just after subsetting feature names.

**CarWithShapeAndCSG.sysml**

Changed `Engine::c1` and `c2` to be cylindrically shaped, rather than specialize Cylinder, so they are bodies for intersection to each engine occurrence. The shape (space boundary) of engines is calculated by CAD tools.

### DIFF
--- a/sysml.library/Domain Libraries/Geometry/ShapeItems.sysml
+++ b/sysml.library/Domain Libraries/Geometry/ShapeItems.sysml
@@ -14,10 +14,10 @@ package ShapeItems {
 	private import ControlFunctions::'?';
 
 	/**
-	 * A PlanarCurve is a flat Surface with a given area.
+	 * A PlanarCurve is a Curve with a given length embeddable in a plane.
 	 */
 	item def PlanarCurve :> Curve {
-		attribute :>> area [1] ;
+		attribute :>> length [1] ;
 
 		constraint { notEmpty(outerSpaceDimension) &  outerSpaceDimension <= 2 }
 	}
--- a/sysml.library/Domain Libraries/Geometry/ShapeItems.sysml
+++ b/sysml.library/Domain Libraries/Geometry/ShapeItems.sysml
@@ -9,21 +9,35 @@ package ShapeItems {
 	private import Objects::*;
 	private import Items::Item;
 	private import SequenceFunctions::isEmpty;
-	
+	private import SequenceFunctions::notEmpty;
+	private import SequenceFunctions::size;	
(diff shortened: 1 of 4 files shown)


seidewitz (changes requested):
```
/**
 * A PlanarCurve is a flat Surface with a given area.
 */
item def PlanarCurve :> Curve {
	attribute :>> area [1] ;

	constraint { notEmpty(outerSpaceDimension) &  outerSpaceDimension <= 2 }
}
```
The comment doesn't seem right. A PlanarCurve is a Curve, not a Surface. Perhaps it is a cut-and-paste error.

And shouldn't the PlanarCurve have a length, not an area? An area only might make sense if the curve was closed, which isn't a requirement here.

conradbock:
Oops, copy/paste error, will fix.

conradbock:
Pushed PlanarCurve fix.  Also Objects.kerml, missed pushing it before.

hpdekoning (approved):

hpdekoning:
All looks OK for me.

PR #358, opened 2022-05-06
ST6RI-539: SysMLLibraryUtil.isModelLibrary does not work in Jupyter on Windows

himi:
In Jupyter environment, SysML kernel indirectly calls setModelLibraryDirectory with the kernel directory string, which contains OS-specific path syntax such as backslash and drive letter and volume separator (`c:`).  This makes inconsistencies in isModelLibrary path matching.

In this fix, setModelLibrary normalize the directory string to EMF URI form and use drivePath() if it's not null; otherwise use path().  Also I changed the name of `modelLibraryDirectory` to `modelLibraryPath` because it's normalized to URI path syntax.

isModelLibray() also checks if path() is not null because some EMF Resource URI does not have path() part.

This fix also adds getModelLibraryPath() getter method for other modules to obtain the library path.


himi:
I checked this fix on Windows 11 (JDK11) and MacOS 12.2.1 (JDK11).

seidewitz (approved):

PR #364, opened 2022-06-06
ST6RI-551 Specializations to feature chains are not rendered (PlantUML)

himi:
When we subset or redefine features with feature chains, they were not shown because VPath did not care about such feature chains.  This fix renders such relationships.

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VActionMembers.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VActionMembers.java
@@ -46,6 +46,11 @@ public class VActionMembers extends VBehavior {
     }
 
     public String caseFeature(Feature f) {
+        if (VSSRMembers.isStateSpaceMember(f)) {
+            if (!addRecLine(f, true)) return null;
+            VSSRMembers vs = new VSSRMembers(this);
+            return vs.start(f);
+        }
         FeatureDirectionKind fdk = f.getDirection();
         if (fdk == null) return null;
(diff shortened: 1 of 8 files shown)


himi:
In the following example, 

```
package FeatureChainTest {
    part p1 {
        part p2 {
        	part p211 {
        		attribute a;	
        	}
        }
    }
    part p3 :> p1.p2;
    part p4 :> p3.p211;   
}
```

Still we have an issue in tree view without `SHOWINHERITED` style:

Screen Shot 2022-06-06 at 12 09 24 PM
(With `SHOWINHERITED`, it is rendered as:
Screen Shot 2022-06-06 at 12 09 54 PM
)

It is because inherited features to be referred are always rendered in compartment because the current visualizer cannot identify the relevant membership for such inherited features.  It requires more complex modifications to VTree and VCompartment.  I decided to defer it.    Interconnection view does not have such a problem because it does not explicitly render memberships.
Screen Shot 2022-06-06 at 12 12 04 PM

himi:
I could come up with the way to get a membership from an element by looking up inherited memberships.  So the issue above was solved.

seidewitz (approved):
The result is a little odd, because you end up with two symbols for `P211`. But I think it is acceptable for now, and better than not having it rendered at all.

himi:
@seidewitz previously, it was rendered as one symbol. But in this case, we could not distinguish p4 subsets `p3.p211` instead of `p2.p211`.  So the visualizer renders it twice but with `^` to show it is inherited.

PR #365, opened 2022-06-08
ST6RI-545 "Context" features for nesting and for performances/actions

seidewitz:
This PR adds two new features to Kernel Library models: `that`, which is inherited in all features, and `this`, which is inherited by all occurrences.

**`that`: context for nesting**

The `that` feature is declared as a nested feature of `Base::things`, the base for all features:
```
abstract feature things: Anything [0..*] nonunique {
    feature that : Anything[1];
}
```

It is constrained in `Base::Anything` (the base type of everything) such that the `that` reference for all values reached by navigating through any nested feature of anything is the nesting thing.
```
abstract classifier Anything {
    feature self: Anything[1] subsets things chains things.that;
}
```

For example, given
```
classifier Example {
    feature f;
}
feature x : Example[1];
```
the value of `x.f.that` is `x`.

**`this`: context for performances**

The `this` feature provides a context for performances similar to the "this" pointer in object-oriented programming languages such as C++, Java and JavaScript. For occurrences that are not performances, `this` defaults to `self`:
```
abstract class Occurrence specializes Anything {
    ...
    feature this : Occurrence[1] default self;
    ...
}
```

For performances that are owned via composite features of an object, `this` defaults to the owning object:
```
abstract class Object specializes Occurrence {
    ...
    composite step ownedPerformances: Performance[0..*] subsets enactedPerformances {
        feature redefines this default that;
    }
    ...
}
```
(All composite steps nested in a class or object feature directly or indirectly subset `ownedPerformances`.)

For performances that are owned via composite features of a performance, `this` defaults to the same value as for the owning performance:
```
abstract class Performance specializes Occurrence {
    ...
    composite step subperformances: Performance[0..*] subsets enclosedPerformances {
        feature redefines this default (that as Performance).this;
    }
    ...
}
```
(All composite steps nested in a behavior or step directly or indirectly subset `subperformances`.)

So, for example, a step declared at "package level" will be the `this` reference for all steps in the composition tree it roots:
```
package Example1 {
    behavior A {
        composite step b {
            composite step c;
        }
    }

    step a : A;
    // a.b.c.this == a
}
```
On the other hand, a composite step declared within a class will have an instance of that class as the `this` reference for itself and all composite steps within it:
```
package Example2 {
    class C {
        composite step a {
            composite step b;
        }
    }

    feature c : C;
    // c.a.b.this == c
}
```

**`AcceptAction`: default receiver**

`AcceptAction` has a `receiver` parameter that references the occurrence via which an `incomingTransfer` may be accepted. The declaration for `AcceptAction` has been updated so that the default `receiver` is `this`:
```
action def AcceptAction {
    inout payload[0..*];
    in receiver: Occurrence[1] default this;
    ...
}
```
In the concrete syntax for accept actions, the `receiver` is specified using a **`via`** clause. The default is used if no **`via`** clause is given. For example, in
```
part def Device {
    state Device_Behavior {
        state off;
        transition
            first off
            accept On_Signal
            then on;
        state on;
    }
}
```
the default receiver for `accept On_Signal` is the instance of `Device` owning the `Device_Behavior` performance. (Note that this is similar to the concept of the "context object" for event pooling in UML/SysML v1.)


conradbock:
Maybe in Kernel the `this` bindings in Object and Performance (and elsewhere if any) should be `default`, since choosing a “c” thing seems potentially language-dependent, and some built on Kernel in the future might make other choices, even tho we UMLers don't have experience with them yet.  SysML could redefine them to be bindings.

conradbock:
About redefining feature names that aren't needed with `that` available, in addition to `nestedCoorinateFrame` I found
        `Item::envelopingShapeConstraint {... Item::self ...}`
which I can fix, but there was another one I wasn't sure what it parsed to:
         `View::viewpointConformance by View::self`
Couldn't tell if it should be `that`'d.

A search on "nest" didn't come up with anything more, but would need to search all redefinitions to really know, eg, wasn't sure about Part::`portsOnPart`, which gave me unpleasant UML flashbacks.

conradbock:
Couple parse problems (or user errors) with `that`/casting, when removing the renaming of `localClock` and `coordinateFrame` in SpatialItems.  Changed:
```
item def CompoundSpatialItem :> SpatialItem {
  item componentItems : SpatialItem[1..*] ordered :> subitems {
    item nestedLocalClock :>> localClock default localClock;
    attribute nestedCoordinateFrame :>> coordinateFrame {
      attribute :>> transformation[1] {
        attribute :>> source = coordinateFrame;
	attribute :>> target = nestedCoordinateFrame; } } } ...}
```
to
```
item def CompoundSpatialItem :> SpatialItem {
  item componentItems : SpatialItem[1..*] ordered :> subitems {
    item :>> localClock default (that as CompoundSpatialItem).localClock;
    attribute coordinateFrame :>> coordinateFrame { /* Parse error 1 above ↑↑↑ */
      attribute :>> transformation[1] {
        attribute :>> source = /* Parse error 2 below ↓↓↓ */
          (that as CoordinateFrame).(that as SpatialItem).(that as CompoundSpatialItem).coordinateFrame;
	attribute :>> target = (that as CoordinateFrame);  } } } ...}
```

The first error (localClock default) is:

> Multiple markers at this line
> Couldn't resolve reference to Element 'that'.
> Must be a valid feature

It goes away if I don't use that (which is the way it was, not what it parses to, tho).

The second error (coordinateFrame binding) is:

> Multiple markers at this line
> no viable alternative at input ')'
> no viable alternative at input '('
> no viable alternative at input 'as'
> Duplicate of inherited member name
> Duplicate of other element name
> no viable alternative at input 'that'

BTW, seems like coordinateFrame should be defaulted since localClock is.

conradbock:
Couple parse problems (or user errors) with `that`/casting, when removing the renaming of `localClock` and `coordinateFrame` in SpatialItems.  Changed:
```
item def CompoundSpatialItem :> SpatialItem {
  item componentItems : SpatialItem[1..*] ordered :> subitems {
    item nestedLocalClock :>> localClock default localClock;
    attribute nestedCoordinateFrame :>> coordinateFrame {
      attribute :>> transformation[1] {
        attribute :>> source = coordinateFrame;
	attribute :>> target = nestedCoordinateFrame; } } } ...}
```
to
```
item def CompoundSpatialItem :> SpatialItem {
  item componentItems : SpatialItem[1..*] ordered :> subitems {
    item :>> localClock default (that as CompoundSpatialItem).localClock;
    attribute coordinateFrame :>> coordinateFrame { /* Parse error 1 above ↑↑↑ */
      attribute :>> transformation[1] {
        attribute :>> source = /* Parse error 2 below ↓↓↓ */
          (that as CoordinateFrame).(that as SpatialItem).(that as CompoundSpatialItem).coordinateFrame;
	attribute :>> target = (that as CoordinateFrame);  } } } ...}
```

The first error (localClock default) is:

> Multiple markers at this line
> Couldn't resolve reference to Element 'that'.
> Must be a valid feature

It goes away if I don't use that (which is the way it was, not what it parses to, tho).

The second error (coordinateFrame binding) is:

> Multiple markers at this line
> no viable alternative at input ')'
> no viable alternative at input '('
> no viable alternative at input 'as'
> Duplicate of inherited member name
> Duplicate of other element name
> no viable alternative at input 'that'

BTW, seems like coordinateFrame should be defaulted since localClock is.

conradbock:
After your fixes the only change left was to under CompoundSpatialItem::CompoundSpatialItem, to the transformation source and target `default`, like` localClock` is.  I guess not quite this issue, if that's alright.  HP and I were looking at it this morning.

sysml.library/Systems Library/Actions.sysml
```diff
@@ -117,7 +119,10 @@ package Actions {
 	 * A SendAction is an Action used to type a SendActionUsage. It initiates an outgoingTransferFromSelf 
 	 * to a designated receiver Occurrence with a given payload.
 	 */
-	action def SendAction(in payload[0..*], in receiver: Occurrence[1]) {
+	action def SendAction {
+		in payload[0..*];
+		in receiver: Occurrence[1];
```
hpdekoning, line 124: Just for my understanding: what is the rationale for moving away from programming language style, positional formal parameter declarations to plain feature declarations? Is it to support alternative constructor signatures with positional parameters in the future?

hpdekoning (approved):

conradbock (approved):

sysml.library/Systems Library/Actions.sysml
seidewitz: Since parameters are now just directed features on behaviors, I think it is, in general, clearer to use the directed feature notation consistently. Further, the syntax for the programming-style parameters is more limited. In this particular case, it was necessary to use the directed feature notation in order to put a default on the `receiver` parameter of `AcceptAction`. It therefore seemed more parallel to also use that notation on `SendAction`.

PR #371, opened 2022-06-24
Issue #370 API packages and jar export

### LINKED ISSUE #370: API client inaccessible from dependent plugin project
**SysMLv2 Pilot Implementation version**: 2022-05 (commit: [ST6RI-545 Updated constraint in RiskMetadata::Level to use "that".](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/commit/3307d64e3ee9ca9e2f265a4297cd7826d090fdf1) on June 15th, 2022)

**Steps to reproduce:**
- Create an Eclipse plugin project that depends on the `org.omg.sysml` plugin.
- Try to import and use any of the classes from the `org.omg.sysml.api` package.

The following project reproduces the problem: [sysml.client.zip](https://github.com/IncQueryLabs/sysmlv2-benchmarking/files/8958157/sysml.client.zip). The relevant part:
```java
import org.omg.sysml.ApiClient;
import org.omg.sysml.ApiException;
import org.omg.sysml.Configuration;
import org.omg.sysml.api.CommitApi;
import org.omg.sysml.api.ProjectApi;
import org.omg.sysml.model.Commit;
import org.omg.sysml.model.Project;

//...

ApiClient defaultClient = Configuration.getDefaultApiClient();
defaultClient.setBasePath("http://localhost:9000"); // the SysML v2 API and Services pilot implementation is used
ProjectApi projectApi = new ProjectApi(defaultClient);
CommitApi commitApi = new CommitApi(defaultClient);
try {
	Project project = projectApi.postProject(new Project());
	Commit commit = commitApi.postCommitByProject(project.getAtId(), new Commit(), null);
} catch (ApiException e) {
	// ...
}
```

**Expected output**: The code compiles and runs without an exception, the appropiate requests are sent.

**Actual output**: The code does not compile, instead the following compile errors arise:
```
ApiClient cannot be resolved to a type
ApiException cannot be resolved to a type
CommitApi cannot be resolved to a type
CommitApi cannot be resolved to a type
Commit cannot be resolved to a type
Commit cannot be resolved to a type
Configuration cannot be resolved
ProjectApi cannot be resolved to a type
ProjectApi cannot be resolved to a type
Project cannot be resolved to a type
Project cannot be resolved to a type
The import org.omg.sysml.api cannot be resolved
The import org.omg.sysml.api cannot be resolved
The import org.omg.sysml.ApiClient cannot be resolved
The import org.omg.sysml.ApiException cannot be resolved
The import org.omg.sysml.Configuration cannot be resolved
The import org.omg.sysml.model cannot be resolved
The import org.omg.sysml.model cannot be resolved
```

**Additional info**: Adding the [API jar](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/blob/master/org.omg.sysml/lib/sysml-v2-api-client-all.jar) directly as a dependency will remove compile errors, but instead causes a `LinkageError` in runtime, revealing that the API classes are already loaded. This suggests that the API package isn't properly exported from the `org.omg.sysml` plugin. 

We believe that the jar file should not be added as an extra classpath entry, and the api packages should be exported.

s0mark:
Fixes #370.

Removes the [API jar](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/blob/master/org.omg.sysml/lib/sysml-v2-api-client-all.jar) from the extra classpath, and exports it along with the `org.omg.sysml.api` and `org.omg.sysml.model` packages.


seidewitz:
We will need to run this through our normal issue tracking and sprint scheduling process before making any changes. We will take your proposal under advisement.

PR #380, opened 2022-08-01
ST6RI-500 Implement delegate mechanism for derived properties

seidewitz:
This pull request moves the implementation of derived properties from non-generated in-line method code in metamodel `Impl` classes to separate delegate classes using the Eclipse "setting delegate" mechanism (as described [here](http://wiki.eclipse.org/EMF/New_and_Noteworthy/Helios#Support_for_Feature_Setting_Delegates)). 

**Custom UML Importer**

1. The `org.omg.sysml.uml.ecore.importer.CustomUMLImporter` class is a customization of the `UMLImporter` class from the Eclipse UML2 project, which is used to load the `SysML.uml` file into the `SysML.genmodel`, generating athecorresponding `SysML.ecore` file. This class has been updated to add an `ECore` annotation to the `sysml` Epackage in `SysML.ecore`, with `settingDelegates->http://www.omg.org/spec/SysML`. This declares `http://www.omg.org/spec/SysML` to be the annotation to be used on structural features to be delegates.
2. The `CustomUML2EcoreConverter` class is a customization of the `UML2EcoreConverter` from the Eclipse UML2 project, used by the `CustomUMLImporter` to actually convert the content of `SysML.uml` into the content of `SysML.ecore`. This class as been updated to add a `http://www.omg.org/spec/SysML` annotation to a structural feature in the Ecore file if one of the following conditions holds:
   - The corresponding UML property is derived, but it is _not_ a derived union or a property to be made into a containment feature as part of the custom conversion.
   - The corresponding UML property is not derived, but the structural feature is marked as derived as part of the custom conversion.
   As a result, every structural feature in`SysML.ecore` that is not annotated as a `union` and is not a containment property is annotated for delegation.

**Setting Delegates**

1. Setting delegates are created using the `org.omg.sysml.delegate.DerivedPropertySettingDelegateFactory`, which is registered in the `MANIFEST.MF` file for `org.omg.sysml`. The setting delegate for a given structural feature is instantiated using reflection, based on the name of setting delegate class, which must have the form _`ContainingClassName`_\__`structuralFeatureName`_\_`SettingDelegate`. If no such class exists in the same package as the factory, then the `DefaultDerivedPropertySettingDelegate` is instantiated instead.
2. The `DefaultDerivedPropertySettingDelegate` can be used for a derived structural for which the following are all true:
   - The derived structural is annotated as subsetting at least one other structural feature.
   - The type of the derived feature is a subtype of (and not the same as) the type of the first subsetted feature.
   - The derived feature has a "many" multiplicity.
   - The derivation rule is that the values of the derived feature are all the values of the first subsetted feature that are instances of the type of the derived feature.
3. All derived property setting delegate are implemented using the following base classes:
   - `BasicDerivedPropertySettingDelegate`. This abstract class extends the Eclipse `BasicSettingDelegate.Stateless` class and implements the `get` and `isSet` methods using the abstract `basicGet` method. It is directly extended for properties with UML primitive types, overriding the abstract `basicGet` operation to implement the derivation computation.
   - `BasicDerivedObjectSettingDelegate`. This abstract class extends `BasicDerivedPropertySettingDelegate`, specializing the type of `basicGet` to `EObject` and overriding `get` to provide proxy object resolution if requested. It is used as the base class for setting delegates that implement derivations that result in a non-collection object of a non-primitive type.
   - `BasicDerivedListSettingDelegate`. This abstract class extends `BasicDerivedPropertySettingDelegate`, specializing the type of `basicGet` to `EList<?>` and overriding `isSet` to check for an empty list. It is used as the base class for setting delegates that implement derivations resulting in a list object.

**`Impl` Classes**

1. Previously, getter and setter methods for derived properties were marked as `@generated NOT`, since they contained in-line handwritten code for the derivation computations. The `NOT` has now been removed from the annotation for these methods, so they are automatically generated as calling to the appropriate setting delegate object.
2. The getter and setter methods for certain non-derived properties are still marked as `@generated NOT` in order to implement workarounds for limitations on what parse-tree actions can be specified in an Xtext grammar. Comments have been added to described the functioning of the workaround code.
3. Other methods used to implement operations from the UML abstract syntax metaclasses are still marked as `@generated NOT` and implemented in-line. (The intent is to eventually re-implement these methods using "operation invocation delegates".)


seidewitz:
As part of this pull request, the following abstract syntax model corrections were made in `SysML.uml`:

1. `Relationship::relatedElement` - made not union (still derived)
2. `Association::relatedType` - made not union (still derived)
3. `Connector::relatedFeature` - made not union (still derived)
4. `Function::expression` - changed from redefinition to subset
5. 'FeatureValue::featureWithValue' - made derived
6. `StakeholderMembership::ownedStakeholderParameter` - changed from subset to redefinition

TiloSchreiber (approved):

seidewitz:
I ran the `SysMLInteractiveParsingProfiler` on the 2022-06 release, and, for comparison, on the head of this branch. Ther results indicate that there is some performance degradation, which is not entirely surprising. For parsing and validating a large model, the time seemingly can be up to about 30% longer (e.g., 43 seconds vs. 34 seconds). This is a bit unfortunately, but probably still acceptable.

PR #381, opened 2022-08-05
ST6RI-572 Redundant textual notation simplification

seidewitz:
This pull request removes a number of parts of the textual notation that are now considered redundant or unnecessary as the notation has evolved. These are entirely changes to the KerML and SysML grammars and their processing, with no changes required to the abstract syntax models. (However, one temporary change has been made in the in the implementation of the abstract syntax to help in the transition; see below.)

**Notation Updates**

1. _Logical Operators._ For certain Boolean operators, the expression notation currently includes both operator symbols (adapted from C/Java expression syntax) and equivalent operator keywords (introduced later). This update removes the operator symbols in favor of the keywords.

   | Removed | Retained |
   | :---: | :---: |
   | `! `  |  **`not`**   |
   | `&&` | **`and`** |
   | `\|\|` | **`or`** |
   | `^^` | **`xor`** |
   | `=>` | **`implies`** |
   | `... ? ... : ...` | **`if`**`... ? ...`**`else`**`...` |

   Note that the pure logical operators `&` and `|` have also been retained.

2. _Parameter Declaration._ The textual syntax currently allows for a parenthesized functional notation for declaring the parameters of a behavioral element (e.g., behaviors and steps in KerML and action definitions and usages in SysML), similarly to how parameters are declared in many programming languages. However, in KerML and SysML, parameters are simply directed features of behavioral elements which can also be declared in the body of a behavioral element. This update removes the parenthesized notation in favor of consistently using the same directed feature declaration syntax used for structural elements (e.g., ports).

    **Currently allowed**
    ```
    action def A (
       x : T1,
       out y : T2
    );
    calc def B (
        x : T1;
    ) return : T2;
    ```

    **Alternative using directed features**
    ```
    action def A {
        in x : T1,
        out y : T2
    }
    calc def C {
         in x : T1;
         return : T2;
    }
    ```

     Note that the direction _must_ be provided when declaring a parameter as a directed feature (unlike the special parameter notation, in which **`in`** was the default), otherwise in will not be directed and not considered a parameter. Parenthesized functional notation is can still be used for invoking a function or calculation in an expression, e.g., `C(t1)`.

3. _**`flow from`** Shorthand._ The textual notation currently allows for a **`flow from`** shorthand on a usage declaration for a flow connection into that usage, similar to the feature value shorthand for a binding. This is being eliminated.
4. _Prefix Comments._ Currently, a comment starting with `/**` (i.e., with two stars instead of one) is parsed as a prefix comment that is automatically about the lexically following element in a namespace body. This notation is being eliminated in favor of using a documentation (**`doc`**) comment nested in the body of the element being annotated. In order to allow this more generally, the textual syntax has also been extended to allow bodies on elements that where not able to have them before (e.g., imports, aliases, dependencies, control nodes and transitions).

**Temporary Abstract Syntax Change**

A `PrefixCommit` metaclass has temporarily been added to the `SysML.uml` and `SysML.ecore` abstract syntax metamodels, as a subclass of `Comment`. The previous prefix comment notation is now parsed into this metaclass, although this is now just the same as a regular comment, no longer automatically about the lexically next element. Instead, a validation rule has been added to generate a warning that prefix comments are not longer supported. The intent is to include this in the 2022-07 release to allow aid user transition away from prefix comments, and then remove it in the 2022-08 release (at which time the prefix comment syntax will still parse as a regular comment without warning or error).

**Other Changes**

1. All library, example, training and validation models have been updated to remove prefix comments, so they parse without the prefix comment warning.
2. The Kermel **`relationship`** notation has be extended to allow any kind of element to be declared as an owned related element of a relationship, not just **`element`** declarations and annotations.


seidewitz:
@hpdekoning
Please regenerate the ISO packages in the Quantities and Units Domain Library in order to clean up the documentation comments.

@conradbock 
Please use this branch (ST6RI-572) as the base for your updates for ST6RI-568/569.

conradbock:
I gather folks wanted to keep "?" in the if-then-else, but it seems consistent with the rest of the changes.

conradbock:
Not that it matters much necessarily, but Occurrences.kerml seems to use a different end-of-line character.  Emacs shows it as control-M, whereas the other files use control-J.

conradbock:
Trying the updates to ST6RI-568/56 branching off this 572, this test line parses:
    `feature afNum: Natural [1] = 1;`
but this doesn't:
  `feature afNum: Natural [1] = (if true ? 1 else 0);`     (with or with parens)
says"Features must have at least one type."  I gather it's referring to a feature created by the parser?

conradbock:
Multiplicities don't seem to inherit anymore, eg, this used to parse:
  `in feature transitionLinkSource: StatePerformance redefines TransitionPerformance::transitionLinkSource`
but now says
- Redefining feature should not have smaller multiplicity lower bound
- Subsetting/redefining feature should not have larger multiplicity upper bound

Restating the inherited multiplicity parses:
  `in feature transitionLinkSource: StatePerformance[1] redefines TransitionPerformance::transitionLinkSource`

conradbock:
Not sure if it's related, but the feature above has a nested redef that used to parse (adding the multiplicity workaround above):
  ```
in feature transitionLinkSource: StatePerformance[1] redefines TransitionPerformance::transitionLinkSource {
			feature redefines accepted[1]; }
```
but now give the error:
- Couldn't resolve reference to Feature 'accepted'.

Qualifying with `StatePerformance::accepted` doesn't fix it, even tho StatePerformance defines the feature in the same file.

**Comment by seidewitz** [comment]
Did you do a clean rebuild of system.library? That is always necessary when there are grammar changes.

Sent from my iPhone

On Aug 6, 2022, at 3:19 PM, conradbock ***@***.***> wrote:

﻿

Not sure if it's related, but the feature above has a nested redef that used to parse (adding the multiplicity workaround above):

in feature transitionLinkSource: StatePerformance[1] redefines TransitionPerformance::transitionLinkSource {
                feature redefines accepted[1]; }


but now give the error:

  *   Couldn't resolve reference to Feature 'accepted'.

Qualifying with StatePerformance::accepted doesn't fix it, even tho StatePerformance defines the feature in the same file.

**Comment by conradbock** [comment]
I did a clean and rebuilt all, and just now sysml.library.  Also saved the file and reloaded, that sometimes helps.  The other errirs I reported here eventually went away, but not this one. Happens in two places the same way.

**Comment by seidewitz** [comment]
> I gather folks wanted to keep "?" in the if-then-else, but it seems inconsistent with the rest of the changes.

The keyword **`then`** is used for succession. It can't be used in the syntax for a conditional expression without introducing and ambiguity with the notation for a conditional succession/transition.

**Comment by seidewitz** [comment]
> Not that it matters much necessarily, but Occurrences.kerml seems to use a different end-of-line character. Emacs shows it as control-M, whereas the other files use control-J.

Windows vs. Mac/Unix conventions. It doesn't matter, as long as the end-of-line marker is consistent for all lines in the file.

**Comment by seidewitz** [comment]
> I did a clean and rebuilt all, and just now sysml.library. Also saved the file and reloaded, that sometimes helps. The other errors reported here eventually went away, but not this one. Happens in two places the same way in my files.
> 
> I see a line like this in SequenceFunctions::size: `return : Natural[1] = if isEmpty(seq)? 0 else size(tail(seq)) + 1;` which gives the same error for me. A number of other files similarly (eg, Objects, VectorFunctions).

I checked again, and I have no errors for the branch in my workspace. 

The name of the underlying functions for a conditional expression changed from `?` to `if`, so it may be that it is not resolving the new name for some reason. Make sure you have the very latest commit pulled from the ST6RI-572, restart your runtime Eclipse instance, and try a clean build again.

**Comment by hpdekoning** [comment]
@seidewitz I have now a complete generated update of all ISQ modules with the new doc syntax, on top of the latest ST6RI-572 commit. I used the doc style with two spaces indent, which I found most readable. E.g.:
```
    attribute def LengthValue :> ScalarQuantityValue {
        doc
        /*
          source: item 3-1.1 length
          symbol(s): `l`, `L`
          application domain: generic
          name: Length
          quantity dimension: L^1
          measurement unit(s): m
          tensor order: 0
          definition: linear extent in space between any two points
          remarks: Length does not need to be measured along a straight line. Length is one of the seven base quantities in the International System of Units (ISO 80000-1).
        */
        attribute :>> num: Real;
        attribute :>> mRef: LengthUnit[1];
    }

```
If you prefer another style, then that is easy to change ... 
I also changed the order of declaring the QuantityValue and MeasurementReference, as the doc comment pertains more directly to the quantity than to the XxxUnit (or other MeasurementReference).

The commit with the updated files follows. I have checked that there no errors occur with my Eclipse installation.

seidewitz:
Actually, it is important that you put an asterisk at the start of each line. Otherwise, the initial white space gets included in the comment text in the model, which looks wrong, e.g., when the comment is displayed graphically. (Comment body processing is specified in the KerML spec.)

Sent from my iPhone

On Aug 6, 2022, at 7:36 PM, Hans Peter de Koning ***@***.***> wrote:

﻿

@seidewitz<https://url.emailprotection.link/?buEq4eyNl9NRYRhKUfm4VcPR8zcu3zjFsda3eD_bX5NVmlAJGq5dieBvtZ3foM-YGChCVFfA0HvAdIyOXxXREKg~~> I have now a complete generated update of all ISQ modules with the new doc syntax, on top of the latest ST6RI-572 commit. I used the doc style with two spaces indent, which I found most readable. E.g.:

    attribute def LengthValue :> ScalarQuantityValue {
        doc
        /*
          source: item 3-1.1 length
          symbol(s): `l`, `L`
          application domain: generic
          name: Length
          quantity dimension: L^1
          measurement unit(s): m
          tensor order: 0
          definition: linear extent in space between any two points
          remarks: Length does not need to be measured along a straight line. Length is one of the seven base quantities in the International System of Units (ISO 80000-1).
        */
        attribute :>> num: Real;
        attribute :>> mRef: LengthUnit[1];
    }



If you prefer another style, then that is easy to change ...
I also changed the order of declaring the QuantityValue and MeasurementReference, as the doc comment pertains more directly to the quantity than to the XxxUnit (or other MeasurementReference).

The commit with the updated files follows. I have checked that there no errors occur with my Eclipse installation.

conradbock (approved):

hpdekoning:
@seidewitz Just updated the ISQ generator to generate doc comments that include a leading space-asterisk on each documentation line, as suggested.

For a future improvement I think we should consider adding triple double-quote strings to the syntax, and allowing them for doc comments / bodies, like in Python and Julia. My favorite is to use Julia style (https://docs.julialang.org/en/v1/manual/strings/#Triple-Quoted-String-Literals) that includes auto-dedent behavior. This is a very nice way to also enable asciidoc rich text content inside KerML and SysML code, including for doc strings, which will be very useful for e.g. requirement bodies, but also documentation in general. Using C / Java comment style with an asterisk on each line conflicts with bulleted lists in asciidoc and most markdown syntaxes. In the documentation of the ISQ packages I have anticipated such a capability by using the ASCIIMath notation for mathematical content, which is supported with the asciidoc stem attribute (see https://docs.asciidoctor.org/asciidoc/latest/stem/ ).

If you agree I can create a low priority issue on the Backlog for this.

hpdekoning:
@seidewitz If you want I can fix doc indentation on all non-ISQ files in the Domain Libraries.

seidewitz:
> For a future improvement I think we should consider adding triple double-quote strings to the syntax...

The current comment syntax has known issues. But it is a Track 4 task to resolve this, not a Track 6 task. And it is not a priority before submission, since it should be possible to handle it in the FTF.

seidewitz:
@hpdekoning 

> If you want I can fix doc indentation on all non-ISQ files in the Domain Libraries.

The doc indentation should have been fixed already in all the Domain Library models outside of Quantities and Units. I checked, and found a couple of small remaining errors, which I fixed. So I went ahead and also fixed the indentation in all the non-ISQ models under Quantities and Units. If you find anything else remaining, feel free to fix that, too.

seidewitz:
@hpdekoning 

The file `ISQCharacteristicNumbers.sysml` is generating two warnings due to duplicate names.

1. The name `bodensteinNumber` is declared as an attribute on line 331 and as an alias on line 1114.
2. The name `jFactor` is declared as an attribute on line 868 and as an alias on line 1201.

hpdekoning:
> > For a future improvement I think we should consider adding triple double-quote strings to the syntax...
> 
> The current comment syntax has known issues. But it is a Track 4 task to resolve this, not a Track 6 task. And it is not a priority before submission, since it should be possible to handle it in the FTF.

Fully agreed, it's not a priority before submission, just did not want to forget. Should I record an issue on support of rich text in documentation for Track 4 then?

hpdekoning (approved):

PR #382, opened 2022-08-06
ST6RI-573 Added localClock to Occurrence, used as trigger clock default.

seidewitz:
This PR updates the model library to allow each occurrence to have a "local clock", which is used as the default clock for triggers within an occurrence. This update allows, for example, the following:
```
part context {
  // This defines a local clock used by default within the “context” part.
  part :>> localClock = Time::Clock();

  state behavior {
    entry; then S1;
    
    state S1;
    
    transition
      first S1
      // The time instant in the trigger is, by default, 
      // relative to context::localClock.
      accept at Iso8601DateTime("22-05-12T00:00:00")
      then S2;
    
    state S2;
  }
}
```

**Kernel Model Library**

1. _Clocks._ 
   - Rename `defaultClock` to `universalClock`.
   - Change the default for the `clock` parameter of `TimeOf` and `DurationOf` from `defaultClock` to `localClock`.
2. _Occurrences._
   - Add a `localClock` feature to `Occurrence` with `Clocks::universalClock` as its default.
   - Add a `suboccurrences` feature to `Occurrence` to include all composite suboccurrences of an occurrence and pass the `localClock` of an occurrence down as the default for the `localClock` of its `suboccurrences`.
3. _Objects._
   - Make `Object::subobjects` and `Object::ownedPerformances` subsets of `suboccurrences`.
4. _Performances._
   - Make `Performance::subperformances` a subset of `suboccurrences`.
6. _SpatialFrames._
   - Change the default for the `clock` parameter of various functions from `defaultClock` to `frame.localClock`.
7. _Triggers._
   - Change the default for the `clock` parameter of `TriggerAt` and `TriggerWhen` to their `localClock` (which, by default, will be bound to the `localClock` of some containing "context" of the usages of these functions).

**Systems Model Library**

1. _Items._
   - Make `Item::subitems` a subset of `subobjects`.
   - Make `Item::checkedConstraints` composite and a subset of `ownedPerformances`.

**Quantities and Units Domain Library**

1. _Time._
   - Rename `defaultClock` to `universalClock`, defaulting to `Clocks::universalClock`.
   - Change the default for the `clock` parameter of `TimeOf` and `DurationOf` from `defaultClock` to `localClock`.

**Geometry Domain Library**

1. _SpatialItem._
   - Update the existing declaration of `localClock` for `SpatialItem`  to redefine` Occurrences::localClock` and default to `Time::universalClock`. (Note that this cannot be done in `Item`, because the `Time` package is in the Quantities and Units Domain library.)

**Implicit Specializations**

The following are just the changes to the implicit specialization implementation. Unchanged rules are not included.

1. _FeatureAdapter._
   - A composite feature with a class type and an owning type that is a class or a feature with a class type specializes `Occurrence::suboccurrences`.
   - A composite feature with a structure type and an owning type that is a structure or a feature with a structure type specializes `Object::subobjects`.
3. _UsageAdapter._
   - By default, a usage has the same implicit specialization as for a feature. (This only applies for reference usages and bare usages with user-defined keywords.)
4. _OccurrenceUsageAdapter._
   - If an occurrence usage fits the requirements of being a suboccurrence (as generically for a feature), then it specializes `Occurrence::suboccurrences`. 
5. _ItemUsageAdapter._
   - If an item usage is a suboccurrence but not a subitem, then it specializes `Occurrence::suboccurrences`. (`Item::subitems` subsets `subobjects` which already subsets `suboccurrences`.)
6. _ActionUsageAdapter._
   - If an action usage is a suboccurrence but not a subaction, then it specializes `Occurrence::suboccurrences`. (`Action::subactions` subsets `subperformances` which already subsets `suboccurrences`.)
7. _ConstraintUsageAdapter._
   - If a constraint usage is composite and its owning type is an item definition or usage, then it specializes `Item::checkedConstraints`. (This fixes the required specialization for constraints, which was broken.)

### DIFF
--- a/sysml.library/Kernel Library/Triggers.kerml
+++ b/sysml.library/Kernel Library/Triggers.kerml
@@ -113,7 +113,7 @@ package Triggers {
 			doc
 			/*
 			 * The Clock to be used as the reference for the timeInstant. The default is
-			 * the localClock at the point of invocation of the function. 
+			 * the localClock, which will be bound when the function is invoked. 
 			 */
 		}
 		
@@ -163,7 +163,7 @@ package Triggers {
 			doc
 			/*
 			 * The Clock to be used as the reference for the time delay. The default is
-			 * the localClock at the point of invocation of the function. 
+			 * the localClock, which will be bound when the function is invoked. 
 			 */
 		}
 		
--- a/sysml.library/Kernel Library/Triggers.kerml
+++ b/sysml.library/Kernel Library/Triggers.kerml
@@ -109,11 +109,11 @@ package Triggers {
 			 */
 		}
 		
-		in feature clock : Clock[1] default defaultClock {
+		in feature clock : Clock[1] default localClock {
 			doc
 			/*
 			 * The Clock to be used as the reference for the timeInstant. The default is
-			 * the Clocks::defaultClock. 
+			 * the localClock at the point of invocation of the function. 
 			 */
 		}
 		
@@ -159,11 +159,11 @@ package Triggers {
 			 */
 		}
 		
-		in feature clock : Clock[1] default defaultClock {
+		in feature clock : Clock[1] default localClock {
 			doc
 			/*
 			 * The Clock to be used as the reference for the time delay. The default is
-			 * the Clocks::defaultClock.
+			 * the localClock at the point of invocation of the function. 
 			 */
 		}
 		
--- a/kerml/src/examples/Simple Tests/Classes.kerml
+++ b/kerml/src/examples/Simple Tests/Classes.kerml
@@ -8,13 +8,22 @@ package Classes {
 	}
 	
 	abstract class <'2'> B {
-		public abstract feature a: A;
+		public abstract feature a: A {
+			composite feature aa: A;
+		}
+		public composite feature a1: A;
+		feature x {
+			composite feature a: A;
+		}
 		package P { }
 	}
 	
-	private class C specializes Classes::'2' {
+	private struct C specializes Classes::'2' {
 		private y: A, '2'[0..*];
 		alias z for y;
+		composite feature c : C {
+			composite feature cc : C;
+		}
 	}
 	
 }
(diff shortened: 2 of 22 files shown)


conradbock:
defaultClock->localClock a good idea.  In Objects:

`The suboccurrences of this Object that are themselves Objects.`

would be less geeky as

`The suboccurrences of this Object that are also Objects.`

seidewitz:
I don't have any problem sounding geeky! But I don't seem to have used the "that are themselves" wording elsewhere (at least that I can find on a quick look). I can change it.

conradbock:
My comments in code are showing as Pending again, and Outlook isn't letting me search emails right now to find the one where you figured out the problem.  Can you remember?

seidewitz:
You need to create a review and ask for changes.

conradbock:
That seems to want me to finish all code comments and submit.  Is there a way to do incrementally, like my comment above about Objects?

seidewitz:
I don't think so.

sysml.library/Kernel Library/Triggers.kerml
```diff
@@ -159,11 +159,11 @@ package Triggers {
 			 */
 		}
 		
-		in feature clock : Clock[1] default defaultClock {
+		in feature clock : Clock[1] default localClock {
 			doc
 			/*
 			 * The Clock to be used as the reference for the time delay. The default is
-			 * the Clocks::defaultClock.
+			 * the localClock at the point of invocation of the function.
```
conradbock, line 166: What is "the point of invocation"?

conradbock:
Found an option to make a single comment after starting review.

sysml.library/Kernel Library/Triggers.kerml
seidewitz: A place where the function is invoked.

conradbock (approved):

conradbock:
BTW, Richard was really happy about this one. :)

PR #384, opened 2022-08-09
ST6RI-563 Organize Kernel Model Library text files into subdirectories

seidewitz:
The Kernel Libraries directory in the `sysml.library` Eclipse project has been reorganized into three subdirectories:
   - _Kernel Semantic Library_ – Library models that provide the base types for required implicit specializations in KerML user model constructs, plus additional semantic models for more advanced capabilities (such as state-based behavior, etc.).
   - _Kernel Data Type Library_ – Library models of standard data types that can be used in user models.
   - _Kernel Function Library_ – Library models of standard functions that can be used in user models, including functions corresponding to all the operators in the KerML expression syntax.

Even though this actually has no effect on the scoping or packaging of the models, it provides a useful grouping of the model files that can also be reflected in the subclause organization of the documentation of the libraries in the KerML specification. it can also be considered a precursor to packaging the libraries into interchange projects corresponding to the subdirectories (once that capability has been implemented).


seidewitz:
@conradbock 
If you would like to comment on this, please do so as soon as possible. Thanks.

conradbock (approved):
Much better, thx.

PR #385, opened 2022-08-12
ST6RI-565: Visualize Language Extension (PlantUML)

himi:
Visualize language extensions.

This PR add two new styles:

- HIDEMETADATA: Do not render callouts to show metadata
- SHOWMETACLASS: Always show metaclasses for types as well as MetadataUsage (with `«#...»`)


himi:
Notice that it does not support n-ary connections.   Thus, neither n-ary derived requirements nor cause-effect is properly rendered.

seidewitz (changes requested):
Small change: In the menu in Eclipse, "hide metadata" should be capitalized "Hide metadata", to be consistent with the other options.

himi:
Done.

seidewitz (approved):

PR #386, opened 2022-08-22
ST6RI-578 SysML2XMI error from ItemFlow::itemType setting delegate

seidewitz:
This pull request removes a spurious line from the `ItemFlow_itemType_SettingDelegate::basicGet` method that caused an `IndexOutOfBoundsException` during the derivation of the `itemType` property of an `ItemFlow`/`FlowConnectionUsage` with no item feature.


conradbock:
ServerSequenceRealization-3.syml works now, but the other two (ServerSequenceRealization-2 and ServerSequenceModel) give the error that -3 used to (with and without -g).  This is with a runnable jar (with implicit) exported from Eclipse after checking out this commit.

conradbock:
Works now, forgot to rebuild after checkout.

conradbock (approved):

PR #398, opened 2022-09-15
ST6RI-594: Cannot resolve names in the lastly loaded resource in Jupyter

himi:
The approach I took to resolve ST6RI-594 is that using a dummy resource as a context to get the global scope.  A dummy resource is created with a bogus URI (`new ResourceImpl(URI.createURI("dummy"))`) to prevent local file lookups.  SysMLInteractive.resolve() uses it for scopeProvider.getScope() to avoid exclusions from the local context.

I made a change how to create a Xtext resource by injecting xtextResourceFactory (createNonPersistentXtextResource).  The previous code relies on suffix matching and might make a problem in resource creations.

Also I created a getter for resouceSet, getResourceSet() in SysMLUtil.  So I made resourceSet private and I needed to modify KerML2XMI.createOutputResources().

I also made SysMLUtil.addResourceToIndex() public for other programs to put resources into the index.

### DIFF
--- a/org.omg.kerml.xtext/src/org/omg/kerml/xtext/util/KerML2XMI.java
+++ b/org.omg.kerml.xtext/src/org/omg/kerml/xtext/util/KerML2XMI.java
@@ -78,7 +78,7 @@ public class KerML2XMI extends SysMLUtil {
 			}
 		};
 		
-		this.resourceSet.getResources().add(resource);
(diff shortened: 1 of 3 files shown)


seidewitz:
The bug proposed to be addressed by this PR has not yet been confirmed.

PR #401, opened 2022-09-23
ST6RI-596:  Fixed to show flow connection usages

himi:
This PR fixes the problem that does not render FlowConnectionUsages.   This is partly caused by the metamodel change and the visualizer was confused them as ActionUsage.  So we capture them with caseConnectionUsage() in VCompartment and VDefault.


himi:
This PR also fix the connection rendering problem of 
```
package ConnectorInherited3 {
  part def B {
  	  part g;
  	  part h1;
	  connection c1 connect g to h1;
  }
  
  part b : B {
  	 // connection :>> c1 connect g to h2;
  	 // part h2;
  	 connect g to h1;
  }
}
``` was not rendered correctly.  After the fix, it is correctly rendered as:
![image](https://user-images.githubusercontent.com/10537/191872444-f4b348be-fccb-4a12-9cf7-462a5fd16cbd.png)

PR #402, opened 2022-09-23
ST6RI-594 Cannot resolve names in the lastly loaded resource in Jupyter

seidewitz:
Immediately after opening a Jupyter notebook, doing a `%show DerivationConnections` command results in an error because `DerivationConnections` cannot be resolved. This is because `DerivationConnections.syml` is the last library model loaded, so the resolution search starts with that resource in memory. However, the global scope provider is used for name resolution, and the starting resource for resolution is considered to be a local scope, not part of the global scope. As a result `DerivationConnections` is not resolved.

This pull request fixes the error by creating a dummy resource if the resolution of a magic command argument is attempted immediately after loading the libraries. The last library model is then no longer the last resource in the resource set, and, therefore is properly in global scope.

In addition, the encapsulation of `SysMLUtil` is improved by making its instance fields private (instead of public) and adding public access methods as appropriate. The `addResourceToIndex` method is also made public.

### DIFF
--- a/org.omg.sysml.interactive/src/org/omg/sysml/interactive/SysMLInteractive.java
+++ b/org.omg.sysml.interactive/src/org/omg/sysml/interactive/SysMLInteractive.java
@@ -219,8 +219,7 @@ public class SysMLInteractive extends SysMLUtil {
 		this.counter++;
 		try {
 			List<Membership> globalMemberships = 
-					resourceSet.getResources().stream().
-					filter(r->!inputResources.contains(r)).
+					this.getLibraryResources().stream().
 					flatMap(r->r.getContents().stream()).
 					filter(Namespace.class::isInstance).
 					flatMap(n->((Namespace)n).visibleMemberships(new BasicEList<>(), false, false).stream()).
@@ -324,7 +323,7 @@ public class SysMLInteractive extends SysMLUtil {
 			Element element = this.resolve(name);
 			if (element == null) {
 				return "ERROR:Couldn't resolve reference to Element '" + name + "'\n";
-			} else if (!inputResources.contains(element.eResource())) {
+			} else if (!this.isInputResource(element.eResource())) {
 				return "ERROR:'" + name + "' is a library element\n";
 			} else {
 				String modelName = element.getName() + " " + new Date();
--- a/org.omg.sysml.interactive/src/org/omg/sysml/interactive/SysMLInteractive.java
+++ b/org.omg.sysml.interactive/src/org/omg/sysml/interactive/SysMLInteractive.java
@@ -39,6 +39,7 @@ import java.util.stream.Collectors;
 import org.eclipse.emf.common.util.BasicEList;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.EPackage;
+import org.eclipse.emf.ecore.resource.Resource;
 import org.eclipse.xtext.nodemodel.ICompositeNode;
 import org.eclipse.xtext.nodemodel.util.NodeModelUtils;
 import org.eclipse.xtext.parser.IParseResult;
@@ -93,6 +94,8 @@ public class SysMLInteractive extends SysMLUtil {
 	protected Traversal traversal;
 	
     protected SysML2PlantUMLSvc sysml2PlantUMLSvc;
+    
+    private Resource dummyResource;
 
     @Inject
 	private IGlobalScopeProvider scopeProvider;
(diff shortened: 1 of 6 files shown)


himi:
Since `scope` is injected as `IGlobalScopeProvider`, the first argument of `scope.getScope` will be just extracted from the scope in my understanding, but fortunately, the resource to be created without URIs and it may not be affected in the result. I'll check it for sure when I get back.

seidewitz:
@himi 
I don't really understand your comment. As far as I have been able to test, the proposed changed fixes the identified bug. Do you mean that the description of the original problem in terms of global scope is incorrect?

himi:
The code you changed uses this.resource is used to reparse() the inputs and it may make an issue but I’m not so sure.  I’d like to test it but I cannot do it now in the limited internet connection.  I’m writing this comment on my iPhone ;(

seidewitz:
Except on startup, this.resource will be the last resource in the resource set, so resolve will work exactly as before. On startup, this.resource will be a dummy, but that is replaced with a new resource in eval before the reparse.

seidewitz:
What I am saying is that, as far as reparse is concerned, the new code is the same as the old code, and the old code worked (other than the bug being fixed, which didn’t have to do with eval/reparse). If you think there might be a different bug, that’s a different issue.

himi:
I’m not 100% sure that the old code works for all the inputs to be repeated.  In my view, it looks just fortunate due to the current FilterUriContainer of Xtext implementation.  I want to confirm for sure because this is really intricate problem to identify.

himi:
But if we read the intention of the current Xtext code, both of the old and new codes looks dangerous, I can say.  Please read DefaultGlobalScopeProvider.java in Xtext and in getScope() is implemented as:
```
protected IScope getScope(IScope parent, final Resource context, boolean ignoreCase, EClass type, Predicate<IEObjectDescription> filter) {
		IScope result = parent;
		if (context == null || context.getResourceSet() == null)
			return result;
		List<IContainer> containers = Lists.newArrayList(getVisibleContainers(context));
		Collections.reverse(containers);
		Iterator<IContainer> iter = containers.iterator();
		while (iter.hasNext()) {
			IContainer container = iter.next();
			result = createContainerScopeWithContext(context, result, container, filter, type, ignoreCase);
		}
		return result;
	}
...
protected IScope createContainerScopeWithContext(Resource eResource, IScope parent, IContainer container,
			Predicate<IEObjectDescription> filter, EClass type, boolean ignoreCase) {
		if (eResource != null) {
			URI uriToFilter = eResource.getURI();
			if (container.hasResourceDescription(uriToFilter))
				container = new FilterUriContainer(uriToFilter, container);
		}
		return createContainerScope(parent, container, filter, type, ignoreCase);
	}
```
This code means, return the scope with filtering the resouce in the first argument in `getScope()`.    That is, simply, the result scope may NOT include the names (more specifically IEObjectDescriptor) in the resource in the first argument.  However, the filtering is based on URIs, and there are generated from IResourceDescriptor, depending on the implementation of ResourceDescriptionsData, index of EObjects.  IMHO, we should not give the resource to be looked up into the first argument of `getScope()` of a global scope provider.

seidewitz:
OK, I think I see your point. We should always use a dummy resource for `resolve`, as you had previously proposed.

org.omg.sysml.interactive/src/org/omg/sysml/interactive/SysMLInteractive.java
```diff
@@ -191,29 +194,34 @@ public SysMLInteractiveResult eval(String input, boolean isAddResource) {
 		}
 	}
 	
+	private Resource getDummyResource() {
+		if (this.dummyResource == null) {
+			this.dummyResource = this.createResource("dummy" + SYSML_EXTENSION);
+			this.dummyResource.getContents().add(SysMLFactory.eINSTANCE.createNamespace());
+		}
```
himi, line 201: Why do we need to add a SysML namespace to the dummy resource?  It will not affect anything but I do not see your intention to add it.

himi (approved):
I tested this in command line and jupyter environment including the editor.  This works properly.

org.omg.sysml.interactive/src/org/omg/sysml/interactive/SysMLInteractive.java
seidewitz: It is probably not strictly necessary, but it makes the resource a legal KerML/SysML resource with an empty root namespace. It also prevents the system from trying to autoload content for the resource.

himi: Thank you.  That helps me understand it.

PR #403, opened 2022-09-27
ST6RI-46 Library function implementation and expression evaluation

seidewitz:
**Note:** The previous branch in this repository named ST6RI-46 has been deleted. The branch now named ST6RI-46 is a new branch, with all the same commits as the old branch except the last two.

This pull request revises the implementation of model-level expression evaluation in the `org.omg.sysml` project.

1.  Refactors `ExpressionEvaluator` to use the singleton pattern and improves the mechanism for library function implementation.
2. Adds evaluation of feature chains.
3. Adds implementations for `NumericalFunctions::sum` and `product` as primitive library functions.
4. Adds a Jupyter `%eval` magic command that evaluates an expression against previously parsed model resources.
5. Adds an `EVAL` style to the PlantUML-based visualization, which shows the evaluation of feature-value expressions on diagrams.

This PR does _not_ include the implementation of any non-model-level evaluable expressions. The implementation of the evaluation of invocation of user-defined functions (which is not model-level evaluable) is to be handled in ST6RI-599.


himi:
Let me confirm the commit of https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/commit/715f0b5aaae77abadc991d760a2ef8df58d5ee90 was not included in this PR.  Are you ok with this? I think that means invocations of user-defined functions are dropped.

himi:
Sorry, I missed that.  I did `git reset --hard` to cancel the previous ST6RI-46.  I'll test it.

himi:
I'm testing it with CLI interface.   First I input:
`package EvalTest4 { part p2 { attribute a2 = 2; part p21 { attribute a21 = a2 + 2; } } }`

And trying to evaluate `EvalTest4::p2.p21.a21`.  But none of the below works:
`%eval EvalTest4::p2.p21.a21`
`%eval --target EvalTest4::p2 EvalTest4::p2.p21.a21`
`%eval --target EvalTest4::p2::p21 EvalTest4::p2.p21.a21`
How can we successfully evaluate it?

I'm trying to improv the visualizer but I could not deal with such a case, that is, expressions referring to parent's attributes.

FYI, we call evaluate it by manually calling `EvaluationUtil.evaluate(a21, null)` but I'm puzzled why it works with setting `null` to the target.

seidewitz:
This is not working because when you specify a non-null target, the system is trying to find the feature in _just_ that target, and it isn't considering outer scopes. So, when you evaluate with the target `EvalTest4::p2::p21`, it is not able to get the value of `a2` in `a2 + 2`, because that is defined in the outer scope `EvalTest3::p2`. If you use a null target, however, the system simply follows the feature reference and checks if it is bound.

Handling nested target contexts, including properly dealing with redefinition, is the tricky bit that I have been addressing for ST6RI-599. Your example works on my local ST6RI-599 branch, but I haven't pushed this update. I started dealing with this to handle chained-feature targets in invocation expressions, but then realized I also had to handle it for evaluating features in general. The latter part also applies for model-level evaluable expressions, so it should really be included here as part of ST6RI-46.

seidewitz:
Try it now.

himi:
@seidewitz thank you. I confirmed it worked.  I updated the visualizer to properly pick up `target` in the visualization context.  So I think it gives better visualizations.  I'll show them in the following comments.

himi (approved):

himi:
To test inherits and outer scopes,
```
package EvalTest2 {
	part p1 {
		attribute a1 = 1;
	}
	part p2 {
		attribute a2 = 2;
		part p21 :> p1 {
			attribute a21 = a1 + a2;
		}	
	}
}
```
gives:

Screen Shot 2022-09-28 at 3 10 00 PM

himi:
To test the different inherit contexts,
```
package EvalTest3 {
	part p1 {
		attribute a1;
		attribute a2 = a1 + 3;
		
	}
	part p11 :> p1 {
		attribute :>> a1 = 11;
	}
	part p12 :> p1 {
		attribute :>> a1 = 12;
	}
}
```
gives:

Screen Shot 2022-09-28 at 3 11 05 PM

himi:
To evaluate inherited context with a redefined (bound) outer attribute:
```
package EvalTest4 {

part p1 {
    attribute a1;
    attribute a2 = 3;
    constraint c11 { attribute b1 = a1 < 2; a1 < 2 }
    constraint c12 { a2 < 5 }
}

part p2 :> p1 {
	attribute :>> a1 = 1;
	attribute :>> a2 = 7;
	constraint :>> c12 { a2 < 5}
}
}
```
gives:
Screen Shot 2022-09-28 at 3 12 12 PM

himi:
Finally, I tested it with more practical example:
```
package EvalTest {
    requirement req {
        require constraint { 1< 3 }
    }
    part def Vehicle {
        attribute mass;

        attribute length = 7.2;
        attribute width = 2.8;

        part engine : Engine;

        constraint weightConstraint {
            mass < 1300
        }
        constraint lengthConstraint {
        	length < 8.0
        }
    }

    part def Engine {
        attribute mass = 200;
    }

    part vehicle1 : Vehicle {
        attribute :>> mass = 1000 + engine.mass;
        //attribute :>> mass = 1200;
    }

    part vehicle2 : Vehicle {
        attribute :>> mass = 1500 + engine.mass;
        attribute :>> length = 9.0;
    }

    constraint constraintTest {
        vehicle1.mass == 1200
    }

    requirement def WeightRequirement {
        subject v : Vehicle;
        attribute totalMaxMass = 1400;
        require constraint { v.mass < totalMaxMass }
    }

    requirement vehicle1Requirement : WeightRequirement {
        subject :>> v = vehicle1;
    }

    requirement vehicle2Requirement : WeightRequirement {
        subject :>> v = vehicle2;
    }
}
```
gives:

Screen Shot 2022-09-28 at 3 14 18 PM

This also has issues.  The first one is the same with `EvalTest4`, it cannot evaluate `weightConstraint`.  The other issue is not related to this PR.  But `typeOf` relationships from `vehicle1.engine` to `Engine` are redundant.  I guess it's related to inheritance and I'll look into it.

seidewitz:
@himi 
I did some clean up and refactoring on the ModelLevelExpressionEvaluator code. I made sure that all your visualizer evaluation tests still worked. I also took your `EvalTest1` through `EvalTest4` models and adapted them for use in a Junit `ExpressionEvaluationTest` that is now part of the build.

The difficulty with evaluating the constraints is that when you use a syntax like `constraint c12 { a2 < 5 }`, the result expression is connected to the result parameter just by a binding connector, not by a feature value, and the evaluation algorithm on this branch does not handle this (I presume you are doing something special to evaluate this expression). This will work better after ST6RI-599 (since a constraint is a predicate, which is really a kind of user-defined function).

For the Junit test, I used the following adaption of `EvalTest4`:
```
package EvalTest4 {
	part p1 {
	    attribute a1;
	    attribute a2 = 3;
	    constraint c11 { return result = a1 < 2; }
	    constraint c12 { return result = a2 < 5; }
	}
	
	part p2 :> p1 {
		attribute :>> a1 = 1;
		attribute :>> a2 = 7;
	}
}
```
and tested on, e.g., `p1.c11.result`. But your visualization doesn't seem to show return result parameters on constraints.

himi:
I had wrongly filtered `ReturnParameterMembership` because it's automatically attached calculations, which are frequently used in SSRs.  I changed to filter empty `result` only.

PR #404, opened 2022-09-30
ST6RI-601: Specialization relationships are not rendered properly with SHOWINHERITED style (PlantUML)

himi:
In SHOWINHERITED style, inherited elements can be multiple and we cannot identify sources only with a element.  So I used node id to identify a source.


himi:
By this fix, with SHOWINHERITED style

```
package TestInherit {
    part def Vehicle {
        part engine : Engine;
    }
    part def Engine;

    part vehicle1 : Vehicle;
}
```
is rendered as:
Screen Shot 2022-09-29 at 9 32 25 PM

seidewitz (approved):

PR #405, opened 2022-09-30
ST6RI-599 Basic evaluation of user-defined functions

seidewitz:
This pull request extends expression evaluation to include invocation of user-defined functions, in addition to the model-level evaluation capability from ST6RI-46 (PR #403).

1. Updates `org.omg.sysml.expressions.ModelLevelExpressionEvaluator` is updated to allow the library function implementations to work with extensions to the base `ModelLevelExpressionEvaluator`.
2. Adds `org.omg.sysml.execution.expressions.ExpressionEvaluator` in a new project `org.omg.sysml.execution`, which overrides the `evaluateInvocation` method to implement invocation of user-defined functions. (The `org.omg.sysml.execution` project is intended to eventually include a complete implementation for model execution, but, for now, it only has the expression evaluation extension.)
3. Updates `org.omg.sysml.interactive.SysMLInteractive` and `org.omg.sysml.plantuml.VStructure` to use the new `ExpressionEvaluator`.
4. Updates `org.omg.sysml.interactive.tests.ExpressionEvaluationTest` to test the new capability.

**Capabilities**

1. The primitive functions implemented for model-level evaluable expressions are available.
2. User-defined KerML functions and steps and SysML calculation definitions and calculations can be invoked. There result can be specified using either a result expression, e.g., 
   ```
   calc def Incr {
      in x; 
      x + 1
   }
   ```
   or a return parameter with a bound feature value, e.g.,
   ```
   calc def Incr {
      in x; 
      return result = x + 1;
   }
   ```
3. The capability for handling nested context in ST6RI-46 is used to properly handle scoping and redefinition of features and to allow the invocation of steps/calculations using feature chains. For example, given
   ```
   part p1 {
       attribute x default 1;
       calc c { in y; x + y }
   }
   part p2 :> p1 {
       attribute :>> x = 2;
   }
   ```
   the expression `p1.c(3)` evaluates to `LiteralInteger 4`, while `p2.c(3)` evaluates to `LiteralInteger 5`.
4. User-defined function invocation is supported in expression evaluation using the Jupyter `%eval` command and in PlantUML-based visualization with the `EVAL` diagram style.

**Limitations**

1. As in ST6RI-46, only feature values are recognized for binding the value of a feature. A separately declared binding connector/binding connection usage cannot be used, and subsetting requirements are not automatically enforced.
2. Evaluation is based only on the statically bound values of features in the model structure. Model instantiation and dynamic execution over time are not implemented.


seidewitz:
@himi 
For this update, I have revised `EvalTest4` to:
```
package EvalTest4 {
	part p1 {
	    attribute a1;
	    attribute a2 = 3;
	    constraint c11 { a1 < 2 }
	    constraint c12 { a2 < 5 }
	}
	
	part p2 :> p1 {
		attribute :>> a1 = 1;
		attribute :>> a2 = 7;
	}
}
```
The following evaluations pass in the Junit test:

- `p1.c12()` -> true
- `p2.c11()` -> true
- `p2.c12()` -> false

However, in the visualization:
image
`p2.c11()` does not have a value, and the value for `p2.c12()` is true. This seems to be because the containing constraint usages are being used as the evaluation targets for their result expressions, even when inherited. This puts the evaluation implicitly in the context of `p1`, so the redefinitions in `p2` are not seen.

seidewitz:
New test:
```
package EvalTest5 {
	constraint def MaxValue {
		in max;
		in value;
		value <= max
	}
	
	attribute max1 = 3;
	
	part p {
		attribute value1 = 2;
		attribute value2 = 4;
		
		constraint c1 : MaxValue {
			in max = max1;
			in value = value1;
		}
		
		constraint c2 { MaxValue(max = max1, value = value2) }
	}
}
```
This is visualized correctly:
image
(though it would be nice if an evaluation was shown for the package-level attribute usage `max1`).

seidewitz:
Another new test:
```
package EvalTest6 {
	calc def A {
		in x;
		in y;
		
		x + y
	}
	
	part p1 {
		attribute a = 10;

		calc b {
			in u;
			in v;
			in w;
			
			return r = A(u, v * w) + a;
		}
		
		ref z1 = b(1, 2, 3);
	}
	
	part p2 :> p1 {
		attribute :>> a = 20;

		ref z2 = b(1, 2, 3);
	}
}
```
Also visualized correctly (as a tree diagram):
image

himi:
I fixed the issue in `EvalTest4`.  It wrongly selected the enclosing type when rendering compartments.
Screen Shot 2022-09-30 at 5 03 12 PM

himi:
Regarding the issue in `EvalTest5`, it does not render feature values in attribute usages.   I would like to refactor the current code to deal with this.  So I'll make a Jira issue for it.

himi (approved):

seidewitz:
`EvalTest5` now seems to have a problem:
image
The expression in `p.c1` is no longer showing a value (it should be `true`, as was shown previously).

himi:
I changed the strategy and checked all of the EvalTest?.  However, to check the deep inheritances, I slightly extended `EvalTest5` and could not evaluate `p2.c1` but `p2.c2` was successfully evaluated.   Can you come up with anything?
```
package EvalTest5 {
	constraint def MaxValue {
		in max;
		in value;
		value <= max
	}
	
	attribute max1 = 3;
	
	part p {
		attribute value1 = 2;
		attribute value2 = 4;
		
		constraint c1 : MaxValue {
			in max = max1;
			in value = value1;
		}
		
		constraint c2 { MaxValue(max = max1, value = value2) }
	}
	
	part p2 :> p {
		attribute :>> value1 = 4;
		attribute :>> value2 = 2;
	}
}
```
Screen Shot 2022-10-01 at 12 01 37 AM

himi:
I tested it with CLI interface as well and I now understand why `c1` is not properly evaluated.  The current visualizer do not properly invoke user-defined function rather it directly evaluate operator expressions.  In other words, it cannot evaluate something like `%eval EvalTest5::p2.c1()`.  So I'd need to investigate how to invoke such functions.

seidewitz:
Invocation of a function with a result expression essentially _is_ just evaluating that expression, but with the right kind of target. For nested features, the target needs to be a chained feature that records the path by which the function was invoked. So, for `p2.c1()`, the result expression needs to be evaluated with a target that is the feature chain `p2.c1` (as opposed to `p1.c1`, which would be the "ownership" feature chain for `c1`).

I updated `SysML2PlantUMLText::eval` in your refactored code to construct the proper eval target for a nested invocation (if I understand your code correctly!). With this change, the updated `EvalTest5` visualizes correctly:
image
All the other tests work, too.

I need to go back and better document the use of feature chain targets in the PR #403 for ST6RI-46. This was functionality was not on the ST6RI-46 branch when I originally created the PR (it is part of the code that I later moved from this branch), so it is not mentioned in the original description!

himi:
@seidewitz thank you.  I'd though feature chain would be needed to evaluate `deep` inheritances.  The previous code was something like a compromise.  I confirmed it worked for all the test cases.

PR #410, opened 2022-10-12
ST6RI-579/580 Kernel capabilities for prioritized dispatch, match, and RTC; SysML capabilities for message transfers

conradbock:
This pull request adds Kernel and SysML capabilities for event handling (in the UML sense).

- Occurrences

  - `isDispatch` determines whether `StatePerformances` composed under an `Occurrence `can accept the same incoming transfer to `dispatchScope` more than once.

  - `isRunToCompletion` determines whether `TransitionPerformances` composed under an `Occurrence` can happen during `StatePerformance` `entry` Performances that are composed under `runToCompletionScope`.

  - `incomingTransferSort` determines which transfer to accept when multiple are available and which of the unaccepted transfers are never to be accepted (dispatched). It is a predicate of two transfers that is true when the first should be accepted instead of the other.  It defaults to `earlierFirstIncomingTransferSort`, which is true if the first transfer ends (arrives) before the other (replaces StatePerformance::`isAcceptFirst`).

- Performances

  - `thisPerformance` defaults to the root of the performance composition tree.  It is the default `dispatchScope` for Performances.

- Transfers

  - The following aspects of `Transfers` are moved to `FlowTransfer`s:

      - Features of the `source` and `target` to pickup and drop off items.

      - Optionally initiating a transfer when items become available at the `source`, and removing picked up items from the `source`.

    `MessageTransfers` are disjoint with `FlowTransfers` and do not support the above.

  - `SendPerformance` requires a transfer carrying a `sentItem` in the `outgoingTransfersFromSelf` of its `sender`, with the same transfer in the `incomingTransfersFromSelf` of its `receiver`, if given.

  - `AcceptPerformance` requires a transfer carrying an `acceptedItem` in the i`ncomingTransfersFromSelf` of its `receiver`, 

- TransitionPeformances

  - `triggerTarget` identifies the `target` of the accepted incoming transfer, if any.

  - `accept` is an AcceptAction for the accepted incoming transfer, if any.

- StatePerformances

  - Updated to support the added capabilities in `Occurrences`.

- Actions

  - `SendAction` specializes `SendPerformance` with redefinitions for SysML.

  - `AcceptAction` renamed to `AcceptMessageAction`, specializing `AcceptPerformanc`e, used as `TransitionAction::accepter`, and specialized to new `AcceptAction` that is made analogous to SysML v1 `AcceptEventAction` by adding a state transition between `start` and `done`.

  - `TransitionAction` updated for changes in `TransitionPerformance`.

- Connections

  - `MessageConnection` added as a specialization of `BinaryConnection` and `Transfer`, replacing `FlowConnection` as the base type of message declarations. `FlowConnection` is now a specialization of `MessageConnection`. 
 
  - `MessageConnection::sourceEvent` and `targetEvent` added to identify event occurrences of `source` and `target` of a `MessageConnection` which must happen during its `source` and `target`, respectively, and be its `start` and `done` or occurrences happening before or after them, respectively.

- Interaction Sequence Examples

  - Updated `ServerSequenceRealization-2` (state machine) to

      - Realize `ServerSequenceModel` (sequence specification) messages as `MessageTransfers`, with ports as their `source`/`target`.

      - bind sent/accepted messages to the ones in `ServerSequenceModel.`

  - Added an example of sequence specification that orders
    `start`/`done` of messages, and realizations of it:

      - `ServerSequenceModelOutside` specializing `ServerSequenceModel` redefines events as the `start`/`done` of messages in `ServerSequenceModel`.

      - `ServerSequenceOutsideRealization-2` specializing S`erverSequenceModelOutside`, similar to `ServerSequenceRealization-2`, requiring (rather than defaulting to) `earlierFirstIncomingTransferSort`, ensuring its send/accept order conforms to ServerSequenceModelOutside `start`/`done` order.

      - `ServerSequenceOutsideRealization-3` `ServerSequenceModelOutside`, similar to `ServerSequenceRealization-3,` with instantaneous internal flows to ensure they happen in the same order as `ServerSequenceModelOutside` `start`/`done`.

Other fixes:

  - Changed **step** to **feature** on
    - `TransitionPerformance::transitionLinkSource`
    - `Object::`nvolvingPerfomances`

  - Added `HappensDuring` connector in Occurrence from `self` to `this`.


seidewitz (changes requested):
Please review proposed changes to Transfers model. (Commit [69f881e](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/410/commits/69f881e3fa14b756ed3d2bb01ddeb8cc697c1e1b).)

seidewitz (approved):

PR #411, opened 2022-10-13
ST6RI-605: Feature values in attribute usages is not rendered (PlantUML)

himi:
Render feature values in a compartment of attribute usages.

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/CompartmentEntry.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/CompartmentEntry.java
@@ -1,0 +1,200 @@
+/*****************************************************************************
+ * SysML 2 Pilot Implementation, PlantUML Visualization
+ * Copyright (c) 2022 Mgnite Inc.
+ *    
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of theGNU Lesser General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * @license LGPL-3.0-or-later <http://spdx.org/licenses/LGPL-3.0-or-later>
+ * 
+ * Contributors:
+ *  Hisashi Miyashita, Mgnite Inc.
+ *
+ *****************************************************************************/
+
+package org.omg.sysml.plantuml;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.omg.sysml.lang.sysml.AttributeUsage;
+import org.omg.sysml.lang.sysml.BindingConnector;
+import org.omg.sysml.lang.sysml.Element;
+import org.omg.sysml.lang.sysml.EnumerationUsage;
+import org.omg.sysml.lang.sysml.Feature;
+import org.omg.sysml.lang.sysml.FeatureDirectionKind;
+import org.omg.sysml.lang.sysml.FlowConnectionUsage;
+import org.omg.sysml.lang.sysml.Membership;
+import org.omg.sysml.lang.sysml.OwningMembership;
+import org.omg.sysml.lang.sysml.SuccessionFlowConnectionUsage;
+import org.omg.sysml.lang.sysml.Usage;
+import org.omg.sysml.lang.sysml.util.SysMLSwitch;
+
+class CompartmentEntry implements Comparable<CompartmentEntry> {
+    public final Feature f;
+    public final OwningMembership om;
+    public final String alias;
+    public final String prefix;
+    public final boolean isInherited;
+
+    public String getParameterPrefix() {
+        FeatureDirectionKind fdk = f.getDirection();
+        if (fdk == null) return "";
+        switch (fdk) {
+        case IN:
+            return "<b>in</b> ";
+        case OUT:
+            return "<b>out</b> ";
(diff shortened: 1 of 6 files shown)


himi:
By this fix,

```
package AttributeTest {
	attribute a1 = 5 * 4;
	attribute a2 default 7;
	attribute a3 := 3 + 4;
}
```

is rendered as:

Screen Shot 2022-10-12 at 8 38 25 PM

seidewitz:
Despite the issue only being about attribute usages, it seems that values are actually rendered for all kinds of usages (at least in tree view).

image

This is good (there is no reason to limit it just to attributes), but the PR description should be clear that the update goes beyond the what was stated in the original issue. It should also be made clear that this applies specifically to "standalone" or "package-level" usages -- feature values were already rendered for nested usages.

seidewitz:
Sequence values are being rendered without the enclosing parentheses, which is incorrect. The textual notation requires the parentheses if there is more than one list element.
image
This is not a new problem, though, since sequence values were already rendered this way for nested usages. However, if it is not hard to fix this, it would be good to do it. Otherwise, it can be considered as a separate issue.

himi:
Actually, I did not notice that this issue is applicable beyond attributes ;)   Regarding sequence values, I changed to use the text of FeatureValue instead of Expression.  Needless to say, it's not perfect because it just renders the corresponding text.

seidewitz:
Why weren't the parentheses included in the text of a sequence expression? They are part of such an expression in the grammar.

himi:
I'm not so sure about the result.  But I guess it comes from the combination of the implementation of Xtext `AbstractNode` and KerMLExpression grammar definition.  `AbstractNode` implementation computes text offset and length by using ILeafNode, which (I think) corresponds to the terminal symbol (more specifically, the terminal symbol returning Ecore model element.  And `SequenceExpression` does not include parenthesis `(` and `)` as below:


```
BaseExpression returns SysML::Expression :
	  NullExpression
	| LiteralExpression 
	| FeatureReferenceExpression 
	| MetadataAccessExpression
	| InvocationExpression 
	| BodyExpression
	| '(' SequenceExpression ')'
;

...

SequenceExpression returns SysML::Expression :
	OwnedExpression
	( ','
	| {SysML::OperatorExpression.operand += current} operator = ',' 
	  operand += SequenceExpression
	)?	
;
```

Would you like to change the grammar for this issue?

seidewitz:
You are right, I forgot the grammar was like that! It actually would make more sense for the parentheses to be part of SequenceExpression. The grammar is the way it is to allow for the use of recursion to get right-associativity in SequenceExpression. But sequences don't nest anyway, so the associativity doesn't really matter.

I will try changing it, and, if that doesn't mess anything else up, I would be willing to make the update on this branch.

himi:
I agree and that would be better in the sense that expressions should correspond to the grammars especially when we do analysis.

seidewitz:
It turns out that there is one place that sequence expressions are allowed without parentheses, inside square brackets (mainly to allow a notation like `a[1,2,3]` for multi-dimensional array access). So I created a new `ParenthesizedExpression` production to replace `'(' SequenceExpression ')'` in `BaseExpression`.

See if that works better for you.

himi:
It did not change the result of getText().  By incorporating parentheses into SequenceExpression, it worked.

```
modified   org.omg.kerml.expressions.xtext/src/org/omg/kerml/expressions/xtext/KerMLExpressions.xtext
@@ -318,7 +318,7 @@ BaseExpression returns SysML::Expression :
 	| FeatureReferenceExpression 
 	| InvocationExpression 
 	| BodyExpression
-	| ParenthesizedExpression
+	| SequenceExpression
 ;
 
 // Expression Bodies
@@ -356,11 +356,11 @@ ParenthesizedExpression returns SysML::Expression :
 ;
 
 SequenceExpression returns SysML::Expression :
-	OwnedExpression
+	'(' OwnedExpression
 	( ','
 	| {SysML::OperatorExpression.operand += current} operator = ',' 
-	  operand += SequenceExpression
-	)?	
+	  operand += OwnedExpression
+	)* ')'
 ;
```

Since I screwed up with Xtext LL(*) grammar, could you please check it before commit?

seidewitz:
As I noted previously, this change will break the use of SequenceExpressions without parentheses within brackets. That is why I didn’t make this change. 

Also, your proposed change wouldn’t handle the optional final comma correctly.

himi:
I think I could address both with the following patch:
```
modified   org.omg.kerml.expressions.xtext/src/org/omg/kerml/expressions/xtext/KerMLExpressions.xtext
@@ -318,7 +318,7 @@ BaseExpression returns SysML::Expression :
 	| FeatureReferenceExpression 
 	| InvocationExpression 
 	| BodyExpression
-	| ParenthesizedExpression
+	| ParenthesizedSequenceExpression
 ;
 
 // Expression Bodies
@@ -351,8 +351,12 @@ BodyParameter returns SysML::Feature :
 
 // Sequence Expressions
 
-ParenthesizedExpression returns SysML::Expression :
-	'(' SequenceExpression ')'
+ParenthesizedSequenceExpression returns SysML::Expression :
+	'(' OwnedExpression
+	( ','
+	| {SysML::OperatorExpression.operand += current} operator = ',' 
+	  operand += OwnedExpression
+	)* ','? ')'
 ;
 
 SequenceExpression returns SysML::Expression :
@@ -360,7 +364,7 @@ SequenceExpression returns SysML::Expression :
 	( ','
 	| {SysML::OperatorExpression.operand += current} operator = ',' 
 	  operand += SequenceExpression
-	)?	
+	)?
 ;
 
 // Feature Reference Expressions
```
But it gets more complicated.

seidewitz:
Yeah, I wouldn’t want that redundancy in the grammar just for this purpose.

himi:
Agreed.  So let's keep the current version, using the result of getText() of FeatureValue instead of Expression.

seidewitz:
I reverted commit [974be14](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/411/commits/974be14b36d8d46956c45aadc81b200b53d305a1), since it was causing merge conflicts with master and not helping on this branch.

seidewitz (approved):

PR #412, opened 2022-10-13
ST6RI-587: Documentation is redundantly rendered as comments (PlantUML)

himi:
Documentations in RequirementUsages were redundantly rendered in the form of comments as well as in compartments because they are a kind of comments so that the visualizer render them as comments as well.  By this fix, the visualizer render documents in compartments and if it's the case, it does not render them as comments.

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/CompartmentEntry.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/CompartmentEntry.java
@@ -1,0 +1,200 @@
+/*****************************************************************************
+ * SysML 2 Pilot Implementation, PlantUML Visualization
+ * Copyright (c) 2022 Mgnite Inc.
+ *    
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of theGNU Lesser General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * @license LGPL-3.0-or-later <http://spdx.org/licenses/LGPL-3.0-or-later>
+ * 
+ * Contributors:
+ *  Hisashi Miyashita, Mgnite Inc.
+ *
+ *****************************************************************************/
+
+package org.omg.sysml.plantuml;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.omg.sysml.lang.sysml.AttributeUsage;
+import org.omg.sysml.lang.sysml.BindingConnector;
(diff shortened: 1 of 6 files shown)


himi:
```
package DocumentationTest {

part p1 {
	doc xxx /* tesafjadlfjalfjasldft */
	doc /* test2 */ 
	
	attribute aaa;
}

attribute a1 = 5 {
	doc /* attribute document */
}

requirement r1 {
	doc /* requirement documentation */
}

}
```
is rendered as:
Screen Shot 2022-10-13 at 9 49 17 AM

himi:
One remaining issue is the heading of `documentation` is not properly centred.  I'd like to improve PlantUML to support that.

himi:
Slightly changed the rendering.  Do not show a bullet if it has only one document and showed names if it has.  

Screen Shot 2022-10-13 at 2 47 08 PM

seidewitz (changes requested):
The rendering of multi-line comments in compartments is not very good. Documentation comments, in particular, will often be multi-line.
```
part p1 {
	doc multiline 
	/* 
	 * Documentation comments will generally have multiple lines.
	 * Like this.
	 */
	doc /* Here is another
	 * example.
	 */ 
	
	attribute aaa;
}
```
image
The line breaks needed to be included in the documentation text in the compartment, as they are in the textual notation. 

I also don't think the placement of the documentation name, or the use of bullets, is very natural. It would be better to just have a blank line between sequential documentation texts (a dashed or gray line separator would be even better, but not critical) and to have the name centered (to the extent possible) above the documentation text. Something like this:
```
                       multiline
Documentation comments will generally have multiple lines.
Like this.
 
Here is another
example.
```

himi:
Opps, it's my mistake on the line unfolding. How does it look to you? 

Screen Shot 2022-10-17 at 8 36 34 PM

Actually, I prefer bullet style for multiple documentations to this horizontal line because it's more compact and easier to read multiple documentation items.

seidewitz:
The rendering with the line breaks looks much better. I am still not thrilled with the placement of the comment name, but I can live with it.

I like the horizontal line better than the bullets. The documentation text could include bullets itself, and I think we should have a clearer visualization for actual multiple documentation elements.

The extra lines in the example seem to come from having a line break on the last line of the documentation text. Presumably that would be the same even if bullets were used. If the final line breaks are removed, the visualization is pretty compact.
```
part p1 {
	doc multiline
	/* 
	 * Documentation comments will generally have multiple lines.
	 * Like this.*/
	doc /* Here is another
	 * example.*/ 
	
	attribute aaa;
}
```
image
Would it be possible to automatically remove a final new-line in documentation text before displaying text in a compartment?

seidewitz:
Another point: Shouldn't documentation have a compartment header like "attributes" and the other compartments? That would solve the problem of centering the title, and it would be more consistent. A "documentation" compartment really should just be another kind of compartment.

himi:
OK.  Let's keep the current style.  And I trimmed the body to remove heading and trailing whitespaces.  Regarding the compartment title, I would like to use always centred italic titles  for all of the compartment.  But I have no idea why PlantUML does not support that style.  I'd like to enhance PlantUML to support that.  Before that, I disabled to show a title for documentation compartment because double header lines with long documentation looks quite ugly at least for me.

seidewitz:
Clearly centered italic titles for all compartments would be better, but it seems inconsistent to try to do it for only one compartment. Especially when the title isn't actually centered.

seidewitz (approved):

PR #413, opened 2022-10-13
ST6RI-603: 17a-Sequence-Modeling.sysml is not rendered (PlantUML)

himi:
The visualizer could not render sequence diagrams properly after messages had become FlowConnectionUsages.  I fixed VSequence to properly get ends of FlowConnectionUsages.


himi:
By this fix, 17a-Sequence-Modeling is rendered as:


Screen Shot 2022-10-13 at 3 07 57 PM

I checked 17b and AHFSequences as well.

seidewitz (changes requested):
ST6RI-580 (PR #410), now merged into master, changed the parsing of message declarations. They now parse to flow connection usages _without_ connection ends, instead having redefinitions of the new `Transfer::sourceEvent` and `targetEvent` features that reference subset the "from" and "to" events of the message.

When master is merged into this branch, the sequence diagram rendering no longer works.

himi:
During this work, I found Occurrences.kerml has inconsistent line terminators.  Is it ok to fix it?

seidewitz:
> During this work, I found Occurrences.kerml has inconsistent line terminators. Is it ok to fix it?

Sure.

seidewitz:
> During this work, I found Occurrences.kerml has inconsistent line terminators. Is it ok to fix it?

Sure. Hopefully the fix will survive further merges...

himi:
Hmmm, if it occurs, we should drop d864263 and fix line terminators again...

himi:
Now It seems OK.  I confirmed that this fix worked with 17a, 17b-SequenceModeling.sysml, and AHFSequences.sysml

oysteiha:
Sorry for the late response. I have had a look now and I am glad that working on this revealed other problems that needed mending.
/Øystein

oysteiha (approved):
I am sorry that I cannot really look at the program texts here, but I trust your work on the subject and checking the outcome.

seidewitz:
17a-Sequence-Modeling and 17b-Sequence-Modeling render well. In AHFSequences, `APIS_Transfer_Lifetime` and `MQTTForwarding` render, but `APIS_Transfer_Interface` does not.

In the diagram for `MQTTForwarding`, gate notation is not used for the in parameter `mq`. And presumably `APIS_Transfer_Interface` is not rendered because it is a connection (even though that is ultimately a kind of part), with the messages directly between its ends, not between events on its ends. I think handling these sorts of things can wait until for a future update.

However, there also seems to be an inconsistency in how containing definitions vs. usages are rendered around a sequence diagram. For a definition, the diagram is surrounded in a box with rounded corners with the name of the definition. For a usage, however, there is no bounding box at all.
```
occurrence def PubSubSequence {
   ...
}
```
image
```
occurrence PubSubSequence {
   ...
}
```
image

For consistency with the usual graphical notation conventions, the diagram for a definition should have a square-cornered box, while the diagram for a usage should have a rounded-cornered box, and both should show the name.

himi:
Regarding the box notation, it comes from the limitation of the current PlantUML.  I would like to replace the current sequence diagram rendering in the future.

 `MQTTforwarding` is partially rendered but as you said, gate is not properly rendered.  `APIS_transfer_interface` needs to be investigated (I guess I would need to support interfaces).

seidewitz:
If the box is a PlantUML limitation, then the current rendering is OK for now.

seidewitz (approved):

PR #414, opened 2022-10-14
ST6RI-598 Library models for standard views

hpdekoning:
Added initial version of `StandardViewDefinitions.sysml` library package, as designed in the SST Graphical Specification WG, and introduced in release 2022-08 of the SysML Specification, clause 9.2.18.1 Views Overview.

Note: The corresponding code snippet in SysML Specification, clause 9.2.18.1, has been updated as well, so it is identical.

### DIFF
--- a/sysml.library/Systems Library/StandardViewDefinitions.sysml
+++ b/sysml.library/Systems Library/StandardViewDefinitions.sysml
@@ -58,7 +58,7 @@ package StandardViewDefinitions {
     view def <afv> ActionFlowView specializes InterconnectionView {
         doc /*
              * View definition to present connections between actions.
-             * Valid nodes and edges in an AtionFlowView are:
+             * Valid nodes and edges in an ActionFlowView are:
              * - Actions with nested actions
              * - Parameters with direction
              * - Flow connection usages (e.g., kinds of transfers from output to input)
--- a/sysml.library/Systems Library/StandardViewDefinitions.sysml
+++ b/sysml.library/Systems Library/StandardViewDefinitions.sysml
@@ -21,7 +21,7 @@ package StandardViewDefinitions {
              * element.
              * The typical rendering in graphical notation is as a graph of nodes and edges. 
              */
-        filter @Package | @Membership;
+        filter @Package or @Membership;
     }
 
     view def <pv> PackageView {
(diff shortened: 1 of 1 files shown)


seidewitz (changes requested):
For the definitions that have filters, I think the filters are too restrictive.

1. `MemberView` is restricted to just `Packages` and `Memberships`, the same as `PackageView`. Was the intent for the filter to be `Namespace` and `Membership`? Note, however, that this would not allow, e.g., `Dependencies` or `Specialializations` to be shwon Should this view have any filter at all? The doc says this view "enables presentation of _any_ model element".
2. In all cases, the filters prevent inclusion of non-metadata annotations, such as comments and textual annotations. It seems that `AnnotatingElements` should be included in all cases.

It is also not clear in general that it is useful to include filters at all in just some cases, when they are not included in other cases.

hpdekoning:
> For the definitions that have filters, I think the filters are too restrictive.

Yes, you are right. The filter specs were just a rough first attempt at formalizing what elements would be potentially included / permitted in views of a particular view definition type. At the time of creating these definitions the filter syntax was not well understood.

> Should `MemberView` have any filter at all?

Indeed, the original intent was to specify a filter that permits any owned or non-owned member. We did not think of the consequence of excluding `Dependency`, `Specialization`, etc. Presentation of _any model element_, would imply `filter @Element;` which is effectively the same as no filter at all.

> It seems that AnnotatingElements should be included in all cases.

Quite right. If we regard `MemberView` as the most general view def should we build a specialization hierarchy descending from `MemberView`? I guess `filter` conditions are inherited if no new `filter` is specified, and implicitly redefined if a `filter` statement is added in a specialization, correct? Eventually the name `MemberView` probably needs to be revisited as there are model elements that are not members.

Finally, indeed all view definitions should have a `filter` (or explicitly have none), we just did not get there.

seidewitz:
I think it is best just to take all filters off for this release. It is more confusing to have incorrect ones, than to have none at all. But we should have the package in the library for this release, since we included it in the spec document in the last release. We will have more time to refine it before the final submission for the SysML spec.

hpdekoning:
> I think it is best just to take all filters off for this release.

Agreed. So I will remove the immature filter expressions, and add a task on the backlog to complete the library package. See https://openmbee.atlassian.net/browse/ST6RI-611 .

seidewitz (approved):

PR #420, opened 2022-11-01
ST6RI-602 Update implementation for 2022-10 metamodel baseline

seidewitz:
The 2022-10 metamodel baseline cleans up the abstract syntax model, removing implementation specific elements from the MOF abstract syntax model and the generated `.uml` file used in the implementation.

**Abstract Syntax**

_Metaclasses Added:_

1. `ItemFeature`
2. `ItemFlowEnd`

The above metaclasses had previously been included in the `.uml` file used in the implementation, but they were technically not part of the normative abstract syntax as documented in the KerML specification. They are included in the normative abstract syntax.

_Metaclasses Removed:_

1. `ItemFlowFeature`
2. `SourceEnd`
4. `TargetEnd`

_Metaclasses Modified:_

1. `ItemFlow`
   - `itemFlowEnd` – This property now subsets, rather than redefines, `connectorEnd`, and has a multiplicity lower bound of 0 rather than 1. Use of `ItemFlowEnds` is a convenience for parsing `ItemFlows` and adding implied relationships, but they are no longer required to be used as the `connectorEnds` of `ItemFlows`, as long as the given ends are properly formed.
   - `itemFlowFeature` – This property has been removed.
2. `Feature`
   - `type` – The multiplicity lower bound of this property has been changed from 1 to 0. The multiplicity lower bounds of all properties that (directly or indirectly) subset or redefine `Feature::type` (e.g., `Connector::association`, `Step::behavior`, etc.) have also been changed to 0. Semantic constraints require that all `Features` have at least one `type`, but it may be necessary to include implied relationships to satisfy these constraints. Changing the lower bounds to 0 allows for the interchange of models consistent with the multiplicity, even without implied relationships included.
3. `Relationship`
   - `relatedElement` – The multiplicity lower bound of this property has been changed from 2 to 0. This allows `Relationships` with less than two `relatedElements`. However, all specializations of `Relationship` other than `Association` and `Connector` (and their further specializations) currently redefined `Relationship::source` and `Relationship::target` to have multiplicity lower bounds of 1, which effectively requires them to have at least two `relatedElements`.
5. `Association` and `Connector` – Constraints have been added to require these kinds of elements to have at last two `relatedElements` (e.g., at least two ends) _if_ they are _not_ abstract. The semantic library currently includes abstract `Associations` with less than two ends – e.g., _`Base::Link`_, which has no ends – so this had been an inconsistency with the abstract syntax multiplicities.
6. `OperatorExpression`
   - `operand` – This property has been removed (but it is added back in `SysML.ecore`, see below).

**Ecore**

Previously, during preprocessing of the exported MOF model, the property `Feature::isNonunique` was added to `SysML.uml` for use as an workaround in parsing the `nonunique` keyword. This step has now been eliminated, so the `SysML.uml` file reflects only the normative abstract syntax. Instead, two features are added to `SysML.ecore` by the custom UML to Ecore converter, both of which are used for the Xtext parsing implementation:

1. `Feature::isNonunique`
2. `OperatorExpression::operand`

**Concrete Syntax**

1. _User-defined keywords._
   - User-defined keyword notation (`#`_`keyword`_) is now available in KerML.
   - In both KerML and SysML, multiple user-defined keywords can now be used in a single declaration.
2. _Binding connectors and successions._ In KerML only, it is now possible to declare the ends of a binding connector or succession only in the connector body, rather than using the special notation `... = ...` or `first ... then ...`.
3. _Cast operators._ The parsing of the cast operators **`as`** and **`meta`** has been changed to make the target type directly the type of the result parameter of the operator expression.


seidewitz:
@ivan-gomes @manasbajaj 
The `SysML.uml` file on this branch now reflects the normative abstract syntax with no pilot-implementation-specific additions. It also solves the multiplicity lower bound problem identified in the generation of the JSON schema for the 2022-09 release. The JSON schema needs to be updated using this `SysML.uml`, for inclusion as a normative artifact in the next revised submission.

ivan-gomes:
> @ivan-gomes @manasbajaj The `SysML.uml` file on this branch now reflects the normative abstract syntax with no pilot-implementation-specific additions. It also solves the multiplicity lower bound problem identified in the generation of the JSON schema for the 2022-09 release. The JSON schema needs to be updated using this `SysML.uml`, for inclusion as a normative artifact in the next revised submission.

Ack @seidewitz. Will leave a review once I have tested it.

ivan-gomes (changes requested):
https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/blob/28fccc5b980c2df0d28618e9843b62d413c5247c/org.omg.sysml/model/SysML.uml#L3702-L3703

`ConstraintUsage#constraintDefinition` has an invalid multiplicity of `0.1` for both upper and lower values. I am assuming that is just a typo and a multiplicity of `0..1` was intended.

ivan-gomes (commented):
https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/blob/28fccc5b980c2df0d28618e9843b62d413c5247c/org.omg.sysml/model/SysML.uml#L6114

The type used to be `UMLPrimitiveTypes.library.uml#Real`. Verifying that this change was intentional.

seidewitz:
> `ConstraintUsage#constraintDefinition` has an invalid multiplicity of `0.1` for both upper and lower values. I am assuming that is just a typo and a multiplicity of `0..1` was intended.

Gee, I would have thought this should not even have been possible without some sort of error or warning! But, yes, it is supposed to be `0..1`. This has now been corrected both in `SysML.uml` and in the source MOF model.

> The type used to be UMLPrimitiveTypes.library.uml#Real. Verifying that this change was intentional.

This is due to an annoying issue with the MagicDraw 19.0 export to Eclipse UML, that I currently have to remember to fix by hand each time. I have now changed the type back to `Real`, which is what it should be.

ivan-gomes (approved):
API and JSON Schema definitions have been regenerated from this branch after a manual review that the known issue that required a patch regarding multiplicities was addressed. The JSON Schema definition bundle - intended for spec inclusion - can be found @ https://github.com/Systems-Modeling/SysML-v2-API-Services/blob/master/conf/json/schema/sysml/OmgSysML.json with individual definitions - provided for convenience/implementation - in the parent directory @ https://github.com/Systems-Modeling/SysML-v2-API-Services/tree/master/conf/json/schema/sysml.

Once the 2022-10(-rc1) version of the REST API is deployed and seeded from the textual language pilot implementation I intend to validate the instances against the generated JSON Schema to ensure the issues were resolved and all the validation tests still pass.

himi (approved):
I could not fully check it (rather I found a bug of ST6RI-613) but I did not find any problems in ItemFlow.

PR #428, opened 2022-11-25
ST6RI-618: Show subjects as distinct nodes (PlantUML)

himi:
Previously, the visualizer could show subjects with typeOf relationship only such as
```
  subject s : Part;
```
However, it could not render bindings nor substrings such as:
```
  subject sub = part;
  subject sub :> part;
```
So we decided to render subjects with distinct nodes to show these relationships.

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/CompartmentEntry.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/CompartmentEntry.java
@@ -27,6 +27,7 @@ package org.omg.sysml.plantuml;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.omg.sysml.lang.sysml.ActorMembership;
 import org.omg.sysml.lang.sysml.AttributeUsage;
 import org.omg.sysml.lang.sysml.BindingConnector;
 import org.omg.sysml.lang.sysml.Element;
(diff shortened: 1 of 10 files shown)


himi:
'''
package SubjectTest {
	part def Vehicle;
	part vehicle : Vehicle;

	requirement a {
		subject subjectA : Vehicle;
	}
	
	requirement b {
	    subject subjectB = vehicle;	
	}
	
	requirement c {
		subject subjectC :> vehicle;
	}
}
'''
is rendered as:
Screenshot 2022-11-26 at 9 58 43 AM

himi:
Still `subjectB` has a `values` compartment but it's not so obvious which case we can omit it.  So in this PR, I do not hide it even if it shows a binding connector.

seidewitz:
If a requirement includes a constraint that references an inherited feature of the subject, e.g.,
```
package SubjectTest {
	part def Vehicle  {
		attribute fuelLevel : ScalarValues::Real;
	}

	requirement a {
		subject subjectA : Vehicle;

		assume constraint fuelLevelConstraint {
			subjectA.fuelLevel > 0
		}
	}
}
```
then that feature is visualized with the subject, even if the "Show inherited members" option is turned off:
image
I assume this intentional, because it seems unlikely to be happening by accident, but I found it unexpected.

himi:
By default, inherited members referred by others are rendered.   I did not make an option to hide all of the inherited members.

seidewitz:
OK, I guess I just hadn't noticed this before. It certainly seems to be an allowable default visualization choice.

seidewitz (approved):

PR #434, opened 2022-12-06
ST6RI-624: Short names are not rendered (PlantUML)

himi:
Due to my mistake in ST6RI-565 (PR #385), short names are not rendered.  This PR fix this issue by rendering short names in every element.


seidewitz (changes requested):
I don't believe that short names should be rendered in bold.

himi:
I removed bold tags.

himi:
The new result is:
image

PR #437, opened 2022-12-13
ST6RI-574 Organize model libraries into interchange projects

seidewitz:
Added hand-created `.project` and `.meta` files to each model library directory under `sysml.library`, consistent with the 2022-11 baseline specification for project interchange. With these metadata files included, compressing each directory into an archive gives a legal project interchange (`.kpar`) file using the textual notation as the model interchange format (`.sysml`).

Creating project interchange files using JSON as the interchange format will be handled in a future task.


ivan-gomes (changes requested):
Ran the `.meta` and `.project` files through a JSON Schema validator and noted the full set of errors. Once these are addressed as documented, I approve of the PR without the need for further review.

PR #440, opened 2023-01-03
ST6RI-631: Show short names in compartments, hide empty subjects

himi:
This PR shows short names in compartments.

PR #443, opened 2023-01-12
ST6RI-637: FeatureChains starting with subsetting or redefining features are not properly rendered (PlantUML)

himi:
Feature chains starting with subsetting or redefining features are not rendered.  This PR also fixes redundant circumflexes for inherited features, caused by PR #440, and changed the documentation compartment title to `doc` to align with the decision of Graphical Specification WG.

Other minor updates (per [comment](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/443#issuecomment-1381047169) below):
1. changed the compartment title to `subject` from `subjects`
2. Hide an empty subject in Case view
3. Show types in use cases


himi:
After this fix,
```
package TestFeatureChain1 {
	part p0 {
		part p01 {
			port pp;
		}
	}

    part p1 :> p0 {
    	part p11 :> p01;
   		part p13 {
        	port p2 = p11.pp;
    	}
    }
    
    part p2 :> p0 {
    	part :>> p01;
    	part p21 {
    		port p2 = p01.pp;
    	}
    }
}
```
is rendered as:

Screenshot 2023-01-12 at 5 03 45 PM

seidewitz:
> p2.p01 should redefine p2.p01 instead of p1.p01

I received an e-mail of your comment, but it read "p2.p01 should redefine **p0**.p01 instead of p1.p01", which makes more sense.

In any case, `p2::p01` is actually not redefining either `p0.p01` or `p1.p01` (which would be different feature chains). Rather, it is redefining `p0::p01`, which is the _same thing_ as `p1::p01` (because the latter is inherited). So it is arguable that, when showing inherited features on the diagram, it is not incorrect to show `p2::p01` redefining `p1.p01`.

Of course, the diagram also shows `p2::p01 :>> p0::p01`, so I guess that it would be consistent to have the graphical representation be the same.

seidewitz:
@himi 
I am now working on ST6RI-628 Update implementation for 2022-12 baseline. This will result in some metaclass changes that will effect the PlantUML code. I would like to get this PR merged in first, if possible. When will it be ready for review again?

himi:
@ed, your comment is correct (Sorry, I quickly wrote my comments).  I'd like to finish it today or tomorrow.  The issue on specialization arrows is already fixed.  I'll work on labels using feature chains.

himi:
I got confused about subsettings or redefines with feature chains.  The example above does not use it.  Now it is correctly rendered as:
Screenshot 2023-01-13 at 12 14 27 AM

himi:
Also I updated to support subsettings or redefinitions with feature chains:
```
package TestFeatureChain2 {
	part p0 {
		part p01 {
			part p011 {
			   part p0111;
			   part p0112;
		    }
		}
	}
  
    part p3 :> p0.p01 {
    	part :>> p011.p0111;
    	part p3_0112 :>> p011.p0112;
    }
}
```
gives
Screenshot 2023-01-13 at 12 19 24 AM
But I'm not so sure we can redefine features with feature chains but it's parsed.   In that case, effectiveName is null.  Is it really ok?

himi:
I also made minor updates:

1. changed the compartment title to `subject` from `subjects` (4d31206)
2. Hide an empty subject in Case view (96609d1)
3. Show types in use cases (8605974)

seidewitz:
> But I'm not so sure we can redefine features with feature chains but it's parsed. In that case, effectiveName is null. Is it really ok?

Yes, redefining feature chains is allowed, though it is not clear that it will be much different semantically than just using subsetting, at least in the SysML syntax (where each feature chain is always owned by a different chaining feature). What the effective name should be in that case is open for future adjustment.

seidewitz (approved):
I have confirmed that this works, both with the tests provided, as well as in a couple of other sample cases.

PR #445, opened 2023-01-13
ST6RI-628 Update implementation for 2022-12 baseline

seidewitz:
This pull request updates the implementation for the 2022-12 metamodel baseline. It includes the following changes.

**Abstract Syntax**

1. `Element`
   * This metaclass is now abstract.
   * The (non-derived) `shortName` and `name` properties have been renamed to `declaredShortName` and `declaredName`.
   * The (derived) `effectiveName` property has been renamed to `name`, and a new, derived `shortName` property has been added. The default derivation for these properties is to just have the same value as `declaredName` and `declaredShortName`.
2. `Relationship`
   - This metaclass is now abstract.
3. `Dependency`
   - In the normative abstract syntax, this metaclass has been moved from the SysML `Systems` package to the KerML `Root` package. However, this has no effect on the flat `SysML.uml` model.
4. `Feature`
   * The `name` and `shortName` properties are derived as being the same as the `declaredName` and `declaredShortName`, unless _both_ are empty, in which case they are derived together as being the same as the `name` and `shortName` of the result of the `namingFeature()` operation (if that is non-null).
   * The `namingFeature()` operation is unchanged. Subclasses of `Feature` that have different effective `name` rules overide only the `namingFeature()` operation, so they all inherit the new derivation computation for `name` and `shortName`.

A large number of additional constraints with OCL specifications have been added to the SysML abstract syntax. These are available in the `SysML.uml` and `SysML.ecore` files.

**Concrete Syntax**

1. _Elements and Relationships._ – Since `Element` and `Relationship` are now abstract, the bare **`element`** and **`relationship`** notations have been removed from KerML.
2. _Dependencies._ – Since `Dependency` is now part of the KerML abstract syntax, a **`dependency`** notation has been added to KerML. It has the same syntax as the already existing SysML notation (except for the additional ability to include `ownedRelatedElements` in its body, as for any KerML relationship body).

**Model Library**

1. _`KerML` and `SysML`_ – The reflective `KerML` and `SysML` models have been updated for the abstract syntax changes.

** Other Fixes**

1. Revised `DependencyImpl::getTarget` to add `ownedRelatedElements` at the end of the returned list.
2. Fixed the display of filtered Import relationships in the outline tree view.


himi:
I modified the visualizer to show "effective" names.  For example,

```
part <'P-0'> p0 {
  attribute <'A-1'> a1;
  part <'Part-2'> part2;
}

part p1 :> p0 {
	attribute :>> a1;
	part :>> part2;
}

enum def E0 {
  <'Enum 0'> e0;
  <'Enum 1'> e1;
}

enum def E1 :> E0;
```
is rendered as:
Screenshot 2023-01-15 at 5 14 06 PM

with `SHOWINHERITED` style.

himi (approved):

PR #450, opened 2023-01-18
ST6RI-639 Incorrect rendering if no separator before a feature value (PlantUML)

himi:
The current visualizer prints textual notation for feature values and if it does not have a separator such as leading spaces or equal sign, the results looks just concatenated with the feature name.  This PR inserts an equal or space before the feature values if the corresponding text has no leading spaces.


himi:
After this fix, 
```
action a1 {
		action sendAction send es via a1 {
			in es;
		}
} 
```
gives
Screenshot 2023-01-18 at 12 50 46 PM

seidewitz (changes requested):
This model:
```
package FeatureValueTest {
	attribute a = (1,2);
	
	part p1;	
	part p2 = p1;
	
	action a1;
	action a2 = a1;
}
```
Is now rendering like this:
image
Note the `= =` in the values compartment.

himi:
Sorry.  My silly mistake.

seidewitz (approved):

PR #455, opened 2023-02-07
ST6RI-641 Access to triggering event from target state

conradbock:
Adds StatePerformance::incomingTransitionTrigger, constrained in StateTransitionPerformance to be a MessageTransfer accepted to trigger into the transition target (if it's a StatePerformance).


conradbock:
Pushed null default on StatePerformance::incomingTransitionTrigger (for state performances started by something other than an incoming transition).

seidewitz (approved):

PR #466, opened 2023-02-10
ST6RI-645 Indexing operator

seidewitz:
**Background**

The current expression notation for indexing into a sequence of values is `seq[index]`. This notation is also used for indexing ordered collections and arrays, including multi-dimensional arrays. The textual notation parses to an `OperatorExpression` with `[` as the operator. The function for this operator has its base definition in `BaseFunctions` and it is specialized in `SequenceFunctions` for sequences of values and in `CollectionFunctions` for `OrderedCollection` and `Array`.

**Concerns**

The following uses of the bracket notation for multiplicity and units seem to be more familiar to SysML modelers than its use for indexing:

1. A bracket notation is also used for declaring multiplicities. This can sometimes be confusing, particularly when using multiplicity for connection ends vs. using indexing notation in feature value expressions.
2. Bracket notation is also used in expressions for constructing quantity values with units: `30.0[km/hour]`. While this is _not_ a kind of indexing, it is syntactically the same as an indexing expression, distinguished only by the types of its arguments: a single numerical value and a single unit expression (or any scalar measurement reference).

**Changes**

_Concrete Syntax_

This PR changes the operator used for indexing from `seq[index]` to `seq#(index)`, which parses to an `OperatorExpression` with `#` as the operator. The parentheses around the `index` are required. The second argument can be a list of indexes to accommodate multi-dimensional array indexing.
```
attribute list [*] ordered = ("a", "b", "c");
attribute a = list#(1); // value is "a"
attribute b = list#(2); // value is "b"

attribute arr : Collections::Array {
  :>> dimensions = (2, 3);
  :>> elements = ("a", "b", "c",
                  "x", "y", "z");
}
attribute c = arr#(1,3); // value is "c"
attribute y = arr#(2,2); // value is "y"
```
The bracket notation remains in the expression syntax, parsing to an `OperatorExpression` for the `[` operator, but this will now only used for quantity value construction.

_Abstract Syntax_

Since the new notation parses to an `OperatorExpression`, there is no change to the abstract syntax.

_Validation_

1. _KerML._ A warning whenever the bracket notation is used: "Use #(...) for indexing".
2. _SysML._ A warning when the bracket notation is used and the second argument is _not_ a measurement reference: "Should be a measurement reference (unit)".

_Model Libraries_

1. _BaseFunctions._ An abstract function for `#` has been added.
2. _SequenceFunctions and CollectionFunctions._ The specializations of `[` have been change to corresponding specializations of `#`.
3. A few other library models with indexing expressions have been updated to use the new operator.


hpdekoning (approved):

himi:
Does this PR implements the parser only?  If so,  I'll check the parsed abstract syntax only.

BTW, I think your example on `arr` should be `attribute arr : Collections::Array { :>> dimensions... }`.  We need to specify the definition.

seidewitz:
> Does this PR implements the parser only? If so, I'll check the parsed abstract syntax only.

Indexing was already implemented for model-level evaluable expressions using the old syntax. This has been updated for the new syntax. (See commit [3310368](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/466/commits/3310368213f52cf34a84a1723471a4686671201c).)

> BTW, I think your example on `arr` should be `attribute arr : Collections::Array { :>> dimensions... }`. We need to specify the definition.

Oops! I'll fix that.

himi:
> Indexing was already implemented for model-level evaluable expressions using the old syntax. This has been updated for the new syntax. (See commit [3310368](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/466/commits/3310368213f52cf34a84a1723471a4686671201c).)

My point was that this does not touch with the expression evaluator.  Anyway, I'll check the parsed results.

seidewitz:
> My point was that this does not touch with the expression evaluator.

My point was that it _does_ "touch with the expression evaluator" to the extent that the name under which the `IndexFunction` library function implementation is registered was changed from `[` to `#`. So, an index expression should evaluate correctly using the new notation.

I am not sure what else you might have meant...

himi:
I see. Thank you.  If so, I'll check the evaluation results as well. I'm now building this branch...

himi:
I quickly checked the example below could be parsed properly:
```
package TestIndexOperator {
	import Collections::*;
	part pList {
		attribute list [*] ordered = ("a", "b", "c");

        attribute l0 = list#(0);
		attribute l1 = list#(1);
        attribute l2 = list#(2);
    }

    part pA1 {
		attribute arr : Array {
			:>> dimensions = (4);
			:>> elements = (1, 2, 3, 4);
		}
		
		attribute a1 = arr#(1);
		attribute a0 = arr#(0);
		attribute a5 = arr#(5);
    }
     
    part pA2 {
        attribute arr : Collections::Array {
          :>> dimensions = (2, 3);
          :>> elements = ("a", "b", "c",
                          "x", "y", "z");
        }
        attribute a13 = arr#(1,3); // value is "c"
        attribute a22 = arr#(2,2); // value is "y"
	}
}
```
However, the array index operators could not be evaluated:

Screenshot 2023-02-10 at 8 13 16 PM

seidewitz:
Ah, you are right, indexing for `OrderedCollections` and `Arrays` was not implemented before. So I went ahead and implemented them.
image

himi (approved):
Thank you. I confirmed it worked.

PR #467, opened 2023-02-10
ST6RI-534 Expression arithmetic for quantities and measurement references

hpdekoning:
The update to the Quantities and Units Domain Library enables automated unit / scale conversion, support for coordinate frame transformations, and type checking of expressions that contain quantity values.

The main points are:

1. Reviewed and updated the concepts and use of `QuantityDimension` and `unitPowerFactors : UnitPowerFactor` on `MeasurementUnit`, also in relation with free and bound vector spaces (`isBound : Boolean`).
3. Updated `VectorCalculations` for calculation definition signatures and to ensure compatibility with the KerML packages `VectorValues` and `VectorFunctions` for coordinate frame computations.
4. Added `TensorCalculations` with calculation definition signatures.
5. Added `MeasurementRefCalculations` to support measurement unit arithmetic.

This update is builds on PR #406 (ST6RI-523 Extended coordinate transformation).


seidewitz (changes requested):
This is still missing expression arithmetic for _measurement references_, which is critical. There needs to be a `MeasurementReferenceCalculations` model (or maybe just `UnitCalculations`), and the semantic of this needs to be covered in the technical note and specification.

hpdekoning:
> There needs to be a MeasurementReferenceCalculations model (or maybe just UnitCalculations), and the semantic of this needs to be covered in the technical note and specification.

Just added library `MeasurementRefCalculations.sysml` (keeping the name a little shorter). Will update the Technote and Spec accordingly.

hpdekoning:
@seidewitz Realized that also multiplication '*' and division '/' operators for CoordinateFrame (VectorMeasurementReference) and MeasurementUnit are required. Can we do that in the same library MeasurementRefCalculations, or since it gives name clashes should it be another library package?

seidewitz:
> Can we do that in the same library MeasurementRefCalculations, or since it gives name clashes should it be another library package?

You can put them in the same package, just give them different names, but specializing the general functions with the standard operator names. The expression syntax keys off the generic function names, so the names of the specialized functions aren't supposed to actually matter (we just try to keep them the same for simplicity when we can).

There is an example of this in `CollectionFunctions` in which `[` is specialized for `OrderedCollection`, but then it is also further specialized for `Arrays`, with the name `array[`.

seidewitz:
@hpdekoning 
Also, make the imports private in `MeasurementRefCalculations`.

seidewitz (approved):

PR #469, opened 2023-02-11
ST6RI-482: Render <<perform>> and <<exhibit>> for just `perform` and `exhibit` instead of <<perform action>> and <<exhibit state>>

himi:
So far the visualizer did not distinguish `perform action` and `perform` or `exhibit state` and `exhibit`.  So it visualize actions and states with `<<perform action>>` and `<<exhibit state>>` for both cases.  This PR renders such `perform` and `exhibit` with `<<perform>>` or `<<exhibit>>`

This PR also fixes the following issues:

- `^` was not shown in compartments.
-  specializations are not shown in the action view.
-  `<<state>>` stereotype is not shown in the mixed view.

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/SysML2PlantUMLStyle.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/SysML2PlantUMLStyle.java
@@ -500,13 +500,6 @@ public class SysML2PlantUMLStyle {
     }
 
     public static class StyleStereotypeDefaultSwitch extends StyleStereotypeSwitch {
-        private static boolean hasRefSubsettingWithoutDeclaredName(Feature f) {
-            if (f.getOwnedReferenceSubsetting() == null) return false;
-            if (f.getDeclaredName() != null) return false;
-            if (f.getDeclaredShortName() != null) return false;
-            return f.getOwnedRedefinition().isEmpty();
-        }
-
 		@Override
 		public String caseClass(Class object) {
             if (SysMLPackage.Literals.CLASS.equals(object.eClass())) return " ";
@@ -515,7 +508,7 @@ public class SysML2PlantUMLStyle {
 
 		@Override
 		public String caseExhibitStateUsage(ExhibitStateUsage esu) {
-            if (hasRefSubsettingWithoutDeclaredName(esu)) {
+            if (VStructure.hasRefSubsettingWithoutDeclaredName(esu)) {
                 return " exhibit>> ";
             } else {
                 return " exhibit state>> ";
@@ -534,7 +527,7 @@ public class SysML2PlantUMLStyle {
 
 		@Override
 		public String casePerformActionUsage(PerformActionUsage pau) {
-            if (hasRefSubsettingWithoutDeclaredName(pau)) {
+            if (VStructure.hasRefSubsettingWithoutDeclaredName(pau)) {
                 return " perform>> ";
             } else {
                 return " perform action>> ";
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/SysML2PlantUMLStyle.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/SysML2PlantUMLStyle.java
@@ -508,7 +508,11 @@ public class SysML2PlantUMLStyle {
 
 		@Override
 		public String caseExhibitStateUsage(ExhibitStateUsage esu) {
-            return "<<exhibit state>> ";
+            if (esu.getOwnedReferenceSubsetting() != null) {
+                return " exhibit>> ";
+            } else {
+                return " exhibit state>> ";
+            }
 		}
 
 		@Override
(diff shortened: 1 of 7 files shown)


himi:
After this fix
```
package TestPerformExhibit {
	part p1 {
		action aa {
			perform action pa1;
			then perform a0;
		}
		
		state ss {
			exhibit state es1;
			then exhibit s0;
		}
	}
	action a0;
	state s0;
}
```
is rendered in the tree view as
Screenshot 2023-02-11 at 4 57 58 PM

And the mixed view renders it as:
Screenshot 2023-02-11 at 4 58 41 PM

seidewitz:
This isn't quite right. 

The idea is that, when «perform» is used in the graphical view, then the reference subsetting is _not_ shown, like in the textual notation. That's the point of the notation. Also, when «perform action» is used, then the reference subsetting, if it exists, should be shown using the `::>` symbol, _not_ the `:>` symbol.

For the purposes of the PlantUML rendering, I would suggest that «perform» be used when the `declaredName` and `declaredShortName` are empty and there is a `referenceSubsetting` that provides an effective `name` (and/or `shortName`). So, there would be the following three cases for what is shown in the name compartment for a `PerformActionUsage` (and similarly for an `ExhibitStateUsage`):

1. No reference subsetting (`perform action a1;`):
   | «perform action»<br/>a1 |
   | :---: |
2. Reference subsetting with a declared name (`perform action a2 ::> a0;`):
   | «perform action»<br/>a2 ::> a0 |
   | :---: |
3. Reference subsetting without a declared name (`perform a0;` or `perform action ::> a0;`):
   | «perform»<br/>a0 |
   | :---: |

The should also be a new kind of specialization arrow for reference subsetting, but that is not critical if the reference subsetting is properly identified in the name compartment text.

himi:
I had that idea too in the second case, but the only difference is just give a declared name.  We can come up with this pattern as well for Case 2.

`perform action <aa> ::> a0;`

In my preference, I would like to distinguish Cases 2 and 3 from Case 1.  Semantically it's the separation rather than just the notation, while I understand it aligns more with the textual notation.

himi:
BTW, to introduce another edge notation for `::>`, we need to change PlantUML itself.

himi:
I can come up with this pattern as well:

```
action a0;
	
part def PP {
	action aa;
}
	
part p2 : PP {
	perform action :>> aa ::> a0;
}
```

himi:
I got Sandy's agreement and updated the code to align with the textual notation:

```
package TestPerformExhibit {
	part p1 {
		action aa {
			perform action pa1;
			then perform a0;
			then perform action <'A-0'> ::> a0;
			then perform action a00 ::> a0;
		}
		
		state ss {
			exhibit state es1;
			then exhibit s0;
			then exhibit state <'S-0'> ::> s0;
			then exhibit state s00 ::> s0;
		}
	}
	
	
	part p2 :> p1 {
		perform action :>> aa.pa1 ::> a0; 
	}
	
	state s0;
	action a0;
}
```
gives
Screenshot 2023-02-12 at 10 31 01 AM

seidewitz:
The perform should not show the subsetting in the name compartment (e.g., it should not show `:> TestPeformExhibit::a0`. Otherwise, it is not really different from `perform action`. The point is to just show the name of what is performed in this case, since the name of the perform action usage is then redundant.

seidewitz (changes requested):
The perform should not show the subsetting in the name compartment (e.g., it should not show `:> TestPerformExhibit::a0`. Otherwise, it is not really different from `perform action`. The point is to just show the name of what is performed in this case, since the name of the perform action usage is then redundant.

himi:
I'm almost screwed up with so many requirements.  Finally the example below:
```
package TestPerformExhibit {
	part p1 {
		action aa {
			perform action pa1;
			then perform a0;
			then perform action <'A-0'> ::> a0;
			then perform action a00 ::> a0;
		}
		
		state ss {
			exhibit state es1;
			then exhibit s0;
			then exhibit state <'S-0'> ::> s0;
			then exhibit state s00 ::> s0;
		}
	}
	
	
	part p2 :> p1 {
		perform action :>> aa.pa1 ::> a0;
	}
	
	part p3 {
		perform a0 :> a00;
		perform action a1 ::> a0 :> a0;
	}
	
	action a00;
	state s0;
	action a0;
}
```
is rendered as:

Screenshot 2023-02-12 at 5 14 36 PM

seidewitz:
I think the rules are actually fairly simple, as I stated my earlier comment, just based on whether there is a declared name or short name. The intent is to _avoid_ redundant reference subsetting notation like

| «perform action»<br/>a0 ::> TestPerformExhibit::a0 |
| :---: |

In this regard, e.g., `perform action :>> aa.pa1 ::> a0;` is just the same as `perform a0 :>> aa.pa1;` – there is no difference in the abstract syntax, other than the order of the subsetting relationships, and neither has a declared name, so I would expect them to be rendered the same way graphically using «perform».

seidewitz:
@himi 
Please let me know if you will be able to address my latest comments today, or if we need to defer this PR to the next release. Thanks.

himi:
Sorry, I'm quite busy today and I do not fully understand your point and let me suggest to defer it.  Possibly we need to discuss it in Graphical WG.

seidewitz:
I made one change: in `VStructure::hasRefSubsettingWithoutDeclaredName` I changed `return f.getOwnedRedefinition().isEmpty();` to `return true;`. The result is that the example is rendered as below (the only difference is the one action I mentioned in my previous comment):

image

If this is acceptable, then I am ready to approve and merge this pull request.

seidewitz (approved):

PR #485, opened 2023-06-28
ST6RI-697 Feature chains in Action or Interconnection views are not properly rendered

himi:
The visualizer could not properly render feature chains in behavioural rendering, used by Action, State, Interconnection, Case views.  The example below

```
action a {
    action a1 {
        action a11;
    }
    action a12;
    first a1.a11 then a12;
}
```

is wrongly rendered as:
Screenshot 2023-06-28 at 3 06 12 PM


himi:
After the fix, the example is correctly rendered as:
Screenshot 2023-06-28 at 5 01 27 PM

seidewitz (approved):

PR #489, opened 2023-07-18
ST6RI 690/692/694 Various derivation implementations are incorrect

seidewitz:
This pull request fixes bugs in the implementation of several derived property computations.

1. _ST6RI-690 Connector::connectorEnd implementation is incorrect._ The KerML specification defines the derivation for `Type::endFeature` to be all `features` of a `Type` that have `isEnd = true`. The property `Connector::connectorEnd` redefines `endFeature` and inherits the same derivation. This was implemented correctly in `Type_endFeature_SettingDelegate`, but `Connector_connectorEnd_SettingDelegate` had a separate derivation that only returned owned end features.
2. _ST6RI-692 AssignmentActionUsage_referent_SettingDelegate is incorrect._ The implementation in this setting delegate presumed that the `referent` of an `AssignmentActionUsage` would be the first non-parameter member of the action. This is different than the definition in the SysML specifiation (which is that the `referent` is the first `Feature` that is the `memberElement` of an `ownedMembership` that is not an `OwningMembership`), and it did not work properly when the `referent` was a parameter of a containing action rather than the `AssignmentActionUsage` itself.
3. _ST6RI-694 Derivations of sourceType, targetType, sourceFeature and targetFeature are incorrect.` The utility methods `TypeUtil::getSourceTypeOf`, `TypeUtil::addTargetTypes`, `ConnectorUtil::getSourceFeatureOf` and `ConnectorUtil::addTargetFeatures` did not correctly implement the derivations in the specification of `Association::sourceType/targetType` and `Connector::sourceFeature/targetFeature`.

The PR also changes existing calls in the code of `Connector::getConnectorEnd` to `TypeUtil::getOwnedEndFeatureOf`, so that they produce functionally the same result as before the change to `getConnectorEnd` as described in item 1 above.

### DIFF
--- a/org.omg.kerml.xtext/src/org/omg/kerml/xtext/validation/KerMLValidator.xtend
+++ b/org.omg.kerml.xtext/src/org/omg/kerml/xtext/validation/KerMLValidator.xtend
@@ -461,7 +461,7 @@ class KerMLValidator extends AbstractKerMLValidator {
 		}
 
 		val relatedFeatures = c.relatedFeature				
-		val connectorEnds = c.connectorEnd
+		val connectorEnds = TypeUtil.getOwnedEndFeaturesOf(c)
 		for (var i = 0; i < relatedFeatures.size; i++) {
 			val relatedFeature = relatedFeatures.get(i)
 			if (!(relatedFeature.featuringType.empty || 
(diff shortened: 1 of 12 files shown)


seidewitz:
@himi 
Please note the change to `Connector::getConnectorEnd` from returning just owned ends to returning both owned and inherited ends. I changed all calls to `getConnectorEnd` in the PlantUML code to instead use `TypeUtil.getOwnedEndFeaturesOf` in order to preserve the existing functionality.

seidewitz:
The branch for this PR was renamed, so the PR was closed. See instead PR #490.

PR #508, opened 2023-11-05
[ST6RI-716] Inherited TransitionUsage and Succession are not properly rendered

himi:
Inherited TransitionUsage and Sucession are not correctly rendered because the visualizer failed to recognized the inherited sources and targets.  The main reason was that it did not use the ends of Successions.  Rather it used getSource() and getTarget() of the derived features to render such connections.  However, these are just referenced features and do not have information of the hierarchy of instances.  This PR lets the visualizer use the end features to render Succession and Succession owned by TransitionUsage.


himi:
After this fix, the example,
```
package TestInheritedTransitions2 {
    state s3 {
        state s11 {
            state s111;
            state s112;
            transition first s111 then s112;    
        }
    }
                
    state s2 :> s3 {
        state s02 :> s3; 
    }
 }
```,
is correctly rendered as:
![Screenshot 2023-11-05 at 3 05 11 PM](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/assets/10537/5b87e176-5662-4a17-b9f3-c8a77096e243)
in Interconnection view and with `SHOWINHERITED` style.

**Comment by himi** [comment]
Because `ConnectorUtil.getRelatedFeaturesOf()` cannot be used to resolve each end of connectors, I split this function into `getRelatedFeatureOfEnd()` and moved the same one from `VPath`.  @seidewitz, if you do not like this change, please revert the change of a13dca3.  But I believe this function is useful to manipulate connector ends.  I also simplified `getRelatedFeaturesOf()` as well.

**Comment by himi** [comment]
Oops, I think using ` .toArray()` is to avoid `ConcurrentModificationException` and I restored the code.

**Comment by himi** [comment]
I fixed to support specifications as well as connector ends.  Without 4e40cfd, for example, the generalizations in
```
package TestSpecializations {
    	part p0 {
    		part p01;
    	}
    
    	action a1 {
    		action a11;
    	}

    	part p1 :> p0.p01 {
    		perform a1.a11;
    	}    
}
```
are not rendered.

seidewitz (approved):

PR #511, opened 2023-11-21
ST6RI-723 Resolution SYSML2-467 from SysML v2 FTF Ballot #8

seidewitz:
This PR implements the resolution to the following issue, as approve in SysML v2 FTF ballot #8:

[SYSML2-467](https://issues.omg.org/issues/SYSML2-467) RequirementConstraintUsage should not have a RequirementBody

Specifically:

1. Previously, the textual notation production `RequirementConstraintUsage` allowed a **`require`** or **`assume`** declaration to have a `RequirementBody`, even though the production produced a `ConstraintUsage`, for which requirement-specific nested declarations are not actually allowed. This has now been changed to a `CalculationBody`, which is what is used in other productions for `ConstraintUsages`. 
2. The `ConstraintUsageAdapter` no longer physically inserts subject features into requirements constraints.
3. `SysMLValidator::checkSubjectMembership` no longer makes an exception to `validateSubjectMembershipOwningType` for requirement constraints.

This means that it is now syntax error to have **`subject`**, **`require`** and **`assume`** declarations in the bodies of requirement constraints in all cases. (In some cases, **`require`** and **`assume`** declarations were syntactically allowed previously, but still violated the `validateRequirementConstraintMembershipOwningType` constraint.)


himi (approved):
I checked it was successfully built and could parse requirements with ReferenceSubsettings, and also confirmed CalculationBody could be specified in require and assume blocks.

PR #516, opened 2023-11-23
[ST6RI-729] Do not render unnamed alias memberships

himi:
The visualizer rendered the unnamed memberships, which are typically created by InitialNodeMember such as `first xxx;`. So it looked actions are rendered in a duplicated manner.  This PR lets the visualizer ignore such memberships.


himi:
After this fix, 
```
action a1 {
	action a11;
	first a11;
	then done;
}
```
is rendered as:
Screenshot 2023-11-23 at 1 13 13 PM

seidewitz (changes requested):
> But this rendering is not correct. The model is parsed as having a succession from `start` to `done`. There is no succession from `a11` to `done`. The proper rendering would include a symbol for `a11` and, in addition, a succession arrow from an "initial node" symbol to a "final node" symbol.

Oops, never mind, I misread the textual notation!

himi:
Sorry, I do not get what you said.  The parsed results does not have any successions involving start.  I could find only `a11` to `done`, no successions involving `a1`.
Screenshot 2023-11-24 at 4 06 55 PM

seidewitz (approved):

PR #525, opened 2024-01-06
ST6RI-727 Implicit redefinitions should be added even if there are owned redefinitions

seidewitz:
1. Corrects `FeatureAdapter::isComputeRedefinitions` so that implicit redefinitions are added even if a `Feature` already has `ownedRedefinitions`, consistent with the rules for implied redefinitions per the specifications.
    * Except that, for now, implicit redefinitions are still _not_ added for parameters that already have `ownedRedefinitions`. See also 
       * [KERML-308 checkFeatureParameterRedefinition fails for named-argument invocations](https://issues.omg.org/issues/KERML-308)
2. Removes the special handling of the redefinition of subject parameters, in which a subject will redefine the subject of a supertype, regardless of its position in the list of parameters. Instead, subject parameter redefinition is handled by the normal rule for implied redefinitions of parameters in order, per the specification. (Note that, in a valid model, a subject parameter must always be the first parameter, so, in this case, an owned subject parameter will still always redefine any subject parameters of superypes.)
3. Corrects the `TradeStudies::TradeStudy` model per the resolution to the issue below (because otherwise the above changes result in validation errors being flagged for the model).
   * [SYSML2-552 Errors in the TradeStudy domain model](https://issues.omg.org/issues/SYSML2-552)

### DIFF
--- a/org.omg.sysml.execution/src/org/omg/sysml/execution/expressions/ExpressionEvaluator.java
+++ b/org.omg.sysml.execution/src/org/omg/sysml/execution/expressions/ExpressionEvaluator.java
@@ -1,6 +1,6 @@
 /*******************************************************************************
  * SysML 2 Pilot Implementation
- * Copyright (c) 2022, 2023 Model Driven Solutions, Inc.
+ * Copyright (c) 2022-2024 Model Driven Solutions, Inc.
  *    
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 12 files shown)


seidewitz:
@himi 
I don't think this one will affect visualization. However, let me know if you would like to review it.

himi:
I think so.  The visualizer currently uses owned redefinitions only.  I can check it for sure with some tricky examples.  But if you'd like to complete it quickly, please do so without my review.

seidewitz:
@himi
There will also be another PR soon resolving ST6RI-720 (Remove automatic insertion of explicit subject and objective features), which might actually have a bigger impact. Perhaps you could just review that change, which will be in addition to the changes from this PR.

himi:
@seidewitz, thank you for letting me know about it.  I'll do so.  This PR itself looks quite reasonable because SysML/KerML parameters are positional as you previously said.

seidewitz:
@himi
FYI, it turns out I will not be creating a PR for ST6RI-720 after all. Implementation for this issue (which has been pushed to the ST6RI-720 branch) has identified additional considerations on the semantics of redefinition that will likely need to be addressed by the FTF(s).

PR #537, opened 2024-02-17
ST6RI-744 Properly show ends in particular start or done actions (PlantUML)

himi:
As the current visualizer uses textual notation to print ends in compartment, it may show cluttered text in ends such as `start` and `done` actions in successions.  This PR fixes this issue by making proper names from the referenced features.

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VCompartment.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VCompartment.java
@@ -361,15 +361,11 @@ public class VCompartment extends VStructure {
     }
 
     private void addEnd(Feature f) {
-        Element e = ConnectorUtil.getRelatedFeatureOfEnd(f);
-        if (e == null) return;
-        // We should construct proper text from end but for the time being, we use just the text as it is.
-        /*
-        if (e instanceof Subsetting) {
-            Subsetting ss = (Subsetting) e;
-            e = ss.getSubsettedFeature();
-        }*/
-        appendText(getText(e), true);
+        Feature relatedFeature = ConnectorUtil.getRelatedFeatureOfEnd(f);
+        if (relatedFeature == null) return;
+        append(getRefName(relatedFeature));
+        // Previously, textual notation was used to render ends.
+        // appendText(getText(e), true);
     }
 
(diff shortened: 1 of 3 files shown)


himi:
I tested this PR with the example below:
```
package TestRefsInComp {
	action a1 {
		first start then cmdOnOff;
        action cmdOnOff;
        succession s1 first cmdOnOff then done;
	}
	
	part def B {
		part b;
		connection b1 connect B::b to C::h; 
	}
	
	part def C {
  	    part h;
  	    connection c1 connect h to a.f;
  	    part a {
  	    	part f;
  	    }
  	}
  	
  	part a0;
  	
  	package P1 {
  		part a1 :> a0;
  		part a11 :> a1;
  	}
  	
  	part a2 :> P1::a1 {
  		action a2;
  		perform a1;
  		perform action a22 ::> a2;
  	}
}
```
The rendered result in Tree view with COMPMOST style is:

seidewitz (dismissed):
There seem to be issues with the "Show inherited members" option. When a connection is shown in a compartment between inherited features, then those features are also shown, even if "Show inherited members" is off. However, if "Show inherited members" is turned on, then those inherited features are _not_ rendered.

seidewitz:
Rendering the action `a1` from your example with "Show inherited members" _off_ results in a rendering as you should previously:
image
Both `start` and `done` are shown as inherited members. However, if "Show inherited members" is turned _on_, then the action is rendered _without_ the inherited features:
image
I could reproduce this behavior when other features inherited from `Action` where used as connection ends, but not when using features inherited from other elements in the same file.

himi:
Thank you for the report.  The problem you reported is related to rendering standard library model element.  The visualizer should not render the standard library element without `SHOWLIB` style.  However, currently by default, it renders elements referred by connectors even they are the standard library elements.   Anyway, this issue has nothing to do with this PR and I will make another JIRA issue to deal with it.

I'm afraid that I do not have enough time to solve this issue by 6th because I'm having a hard deadline on 5th.

seidewitz:
OK, I have confirmed that the problem I reported existed before the changes in this PR. Please submit a bug issue for it. I will approve this PR.

seidewitz (approved):

himi:
Thank you.  I quickly made ST6RI-748 JIRA issue.  I'll work on it possibly after the deadline...

PR #548, opened 2024-03-06
ST6RI-748: Standard library elements are rendered without SHOWLIB and SHOWINHERITED styles (PlantUML)

himi:
he standard library elements are rendered even without SHOWLIB and SHOWINHERITED style if they are referred by some connectors.  Also this PR limits the maximum number of elements to be processed, which is useful for rendering with `SHOWLIB` style.

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/SysML2PlantUMLText.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/SysML2PlantUMLText.java
@@ -73,6 +73,8 @@ public class SysML2PlantUMLText {
         MIXED;
     }
 
+    private static final int MAX_VISITS = 1000;
+
     private final List<SysML2PlantUMLStyle> styles = new ArrayList<SysML2PlantUMLStyle>();
     private final Map<String, String> styleValueMap = new HashMap<String, String>();
     private StyleSwitch styleSwitch;
@@ -468,6 +470,7 @@ public class SysML2PlantUMLText {
         
         init();
 
+        numVisits = 0;
         for (EObject eObj : eObjs) {
             if (eObj instanceof Element) {
                 Element e = (Element) eObj;
@@ -475,6 +478,8 @@ public class SysML2PlantUMLText {
             }
         }
         vpath.init();
+
+        numVisits = 0;
         for (EObject eObj : eObjs) {
             if (eObj instanceof Element) {
                 Element e = (Element) eObj;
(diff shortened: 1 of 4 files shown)


himi:
With `COMPMOST` style and in Tree view, after this fix, 
```
package TestStdLibElemInherit {
	action a1 {
		first start then a11;
		action a11;
	}
	
	action a2 {
		action a21;
	}
	
	action a3 :> a2 {
		first a21 then a31;
		action a31;
	}	
}
```
is rendered as:
Screenshot 2024-03-06 at 10 05 04 AM

And with `SHOWLIB` style, it is rendered as:
Screenshot 2024-03-06 at 10 06 15 AM

himi:
Even with `SHOWINHERITED` style (and without `SHOWLIB`), we get the same result:
Screenshot 2024-03-06 at 10 08 30 AM

But with `SHOWLIB` style, it renders enormous amount of inherited elements:
Screenshot 2024-03-06 at 10 10 02 AM

It takes some time to render it but due to the new limitation mechanism, it won't take so long time, it's finished in a few seconds in my Mac mini M1 environment.

himi:
Note that it is related to https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/537#issuecomment-1974914500

seidewitz (changes requested):
I was able to reproduce all the behavior as given in previous comments. However, if action `a1` is changed to be explicitly defined by `Actions::Action` (or to explicitly subset `Action::actions`), then the library element `start` is no longer shown if SHOWLIB is true but SHOWINHERITED is false:
image
It doesn't seem like the behavior of this option should depend on whether the specialization is explicit or implicit.

himi:
Hmmm... In that case, how can we tell `a1::start` does not belong to the standard library?  `ElementUtil.isStandardLibraryElement(e)` does not seem to work for that purpose.

seidewitz:
In addition, if the declaration for action `a1` is changed to
```
	action a1 {
		first a3.a31 then a11;
		action a11;
	}
```
then this is rendered as
image
While `a3` acts semantically as if it is featured by `Anything`, so it can be used as shown in `a1`, syntactically it is not actually an inherited member of `a1`, so showing it as such is not really correct.

himi:
It looks the different issue.  Again, it's related to a tricky inherit key issue.  It can be reproduced with a simpler example:
```
action a1 {
	first a3.a31 then a11;
	action a11;
}
action a3 {
	action a31;
}
```

himi:
After the investigation so far, both issues you reported are quite tricky, related to `VPath`.  And I'm now too busy to work on these.  I guess I can work on them in April.

seidewitz:
If you add these issues to the Jira backlog, then I am OK with approving this PR for the current sprint.

himi:
Thanks.  I'd like to make two Jira issues to capture these hopefully tonight.

seidewitz (approved):

PR #552, opened 2024-03-16
ST6RI-751 References to result parameters, subject parameters and objectives do not always resolve correctly

seidewitz:
Currently, there are three cases in which elements are still being physically added to the parse tree (rather than cached in adapters like is done for implicit relationships), when they are not declared explicitly in a model: result parameters for expressions, subject parameters for requirements and cases and objectives for cases. This is not actually called for in the specifications, but is still necessary in order for implied redefinitions to be added correctly in the case of multiple specializations. 

For example, in the following model:
```
analysis def A {
    subject s;
    objective o;
    return r;
}
part context {
    part p {
        v = a.v;
        o = a.o;
        r = a.r;
    }
    analysis a : A {
      // subject added that redefines s.
      // objective added that redefines o.
      // return parameter added that redefines r.
    }
}
```
subject, objective and result features are added to the parse tree for the analysis case usage `a`. These features implicitly redefine the corresponding features `s`, `o` and `r` from the analysis case definition `A`. Because of these redefinitions, the references to `s`, `o` and `r` within part usage `p` should resolve to the features of the usage `a`. 

This PR fixes a bug that could result in these references incorrectly resolving to the features of the definition `A` when the declaration of `a` is lexically after the declaration of `p`. It also moves utility methods related to subjects and objectives out of class `TypeUtil` into the SysML-specific class `UsageUtil`.

### DIFF
--- a/org.omg.kerml.xtext/src/org/omg/kerml/xtext/scoping/KerMLScope.xtend
+++ b/org.omg.kerml.xtext/src/org/omg/kerml/xtext/scoping/KerMLScope.xtend
@@ -54,6 +54,7 @@ import org.omg.sysml.lang.sysml.util.ISysMLScope
 import com.google.inject.Inject
 import org.eclipse.xtext.naming.IQualifiedNameConverter
 import org.eclipse.emf.ecore.util.EcoreUtil
+import org.omg.sysml.util.NamespaceUtil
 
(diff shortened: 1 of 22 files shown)


himi (approved):
At least I confirmed all of the feature chains properly redefine the features in `a` in your example.

PR #556, opened 2024-04-05
ST6RI-684 (alternate update) Implement invocation delegates for operations

arminzavada:
This PR proposes a solution to the performance hit introduced in PR #554 (commit 32e3509fc5ca11b8ac364c617a1f67322b651526), as an alternative to the change proposed in PR #555. The update in this PR is to simply add a cache for invocation delegates. There is no change to the implementation of the import and inheritance related operations from commit 32e3509fc5ca11b8ac364c617a1f67322b651526.

### DIFF
--- a/org.omg.sysml/src/org/omg/sysml/delegate/invocation/OperationInvocationDelegateFactory.java
+++ b/org.omg.sysml/src/org/omg/sysml/delegate/invocation/OperationInvocationDelegateFactory.java
@@ -21,6 +21,9 @@
 
 package org.omg.sysml.delegate.invocation;
 
+import java.util.HashMap;
+import java.util.Map;
+
 import org.eclipse.emf.ecore.EClass;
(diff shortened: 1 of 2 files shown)


seidewitz:
The change proposed in this PR is much smaller than that proposed in PR #555, relative to what was already adopted for PR #554. However, further analysis indicates that the performance the change in this PR has slightly worse performance than that of PR #555 (which has performance essentially the same as before PR #554). Further, the adapter-based approach in #555 simply uses the normal Java dynamic dispatching mechanism (essentially the same as when the operations in question were in the `Impl` classes), rather than using the special dynamic delegate dispatch mechanism introduced in PR #554, whose performance needs to be improved with additional caching in this PR.

So, this PR will be closed in favor of the solution in PR #555. Nevertheless, this branch (ST6RI-684-2) will be retained for possible future reference.

PR #557, opened 2024-04-10
ST6RI-762: %viz command causes ConcurrentModificationException

himi:
The changes made in PR #552 resulted in uses of the `%viz` command in the Jupyter environment to sometimes throw a `ConcurrentModificationException`. This happens when traversing the `ownedRelationships` of an element. If the traversal enters into a library model, it is possible (after PR #552) that implicit specialization processing will cause the "on demand" insertion of new elements into the abstract syntax tree in a small number of cases (result parameter for an expression, subject for a requirement or case, objective for a case). If this happens, a new `ownedRelationship` is added (for the membership of the new element), which can cause the `ConcurrentModificationException`.

This PR resolves the problem by adding utility methods that create arrays for those lists for which direct iteration may be unsafe, and revises the traversal code to use those utilities instead of directly iterating over the unsafe lists.

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/InheritKey.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/InheritKey.java
@@ -1,6 +1,6 @@
 /*****************************************************************************
  * SysML 2 Pilot Implementation, PlantUML Visualization
- * Copyright (c) 2022-2023 Mgnite Inc.
+ * Copyright (c) 2022-2024 Mgnite Inc.
  *    
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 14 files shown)


seidewitz (changes requested):
The change made to resolve this bug should be as small as possible. Using an array for iteration is necessary in only the following cases:

- `toOwnedRelationshipArray` (which resolves the actual bug)
- `toOwnedMembershipArray`
- `toOwnedFeatureMembershipArray`

The following are _not_ needed, because direct iteration over the corresponding `Elists` is safe:

- `toOwnedElementArray`
- `toOwnedFeatureArray`
- `toInheritedMembershipArray`
- `toOwnedSpecializationArray`
- `toOwnedTypingArray`
- `toOwnedSubsettingArray`
- `toOwnedRedefinitionArray`

himi:
I quickly dealt with your change request.  (But I have not checked it well)

himi:
I think I could make relevant changes.  I'll check it further later.

seidewitz (approved):

seidewitz:
I verified that the latest update fixes the bugs in the models in which I had previously found it:

- `1a-Parts Tree`
- `3a-Function-based Behavior-1`
- `SysML v2 Spec Annex A SimpleVehicleModel`

This seems sufficient for the current release.

himi:
Ed,thank you for the review.

PR #559, opened 2024-04-13
ST6RI-756: `%viz Actions::AcceptAction::aState::aTransition` in Jupyter causes a PlantUML crash (PlantUML)

himi:
This PR mainly fixes the issue that `%viz Actions::AcceptAction::aState::aTransition` crashes PlantUML.  This is caused by  inappropriate newline in name.  I also refactored SendActionUsage and AcceptActionUsage rendering because they excessively used getText() and may contain cluttered text.

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VBehavior.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VBehavior.java
@@ -131,11 +131,23 @@ public abstract class VBehavior extends VDefault {
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


himi:
By this fix, `%viz Actions::AcceptAction::aState::aTransition` outputs:
Screenshot 2024-04-13 at 4 20 50 AM

himi:
We should update Send/AcceptActionUsage notations to align with the update of the Graphical Specification (standard compartment notation) but it requires some modifications of PlantUML and additions to icons.  So I will defer it.

seidewitz (approved):

PR #560, opened 2024-04-13
ST6RI-761: Variations of definitions are not rendered (PlantUML)

himi:
This PR fixes the issue that `«variation»` is missing in a definition with variation.


himi:
After this fix, 
`variation part def P;`
is rendered as:
Screenshot 2024-04-13 at 12 44 04 PM

seidewitz (approved):

PR #561, opened 2024-04-19
ST6RI-767: Some redefinitions are not correctly rendered (PlantUML)

himi:
So far, redefinitions are handled with inherit key of redefined features but that did not work well in some cases such as that redefined target is missing. 

This PR enhances PC (PathContext) to properly handle redefinitions by differentiating the equivalence classes of redefinitions and non-redefinitions.  By this enhancement, we can use the same inherit key with non-redefinition cases.

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/InheritKey.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/InheritKey.java
@@ -25,7 +25,6 @@
 package org.omg.sysml.plantuml;
 
 import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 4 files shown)


himi:
I checked this PR with many test cases.  For example,

```
package Case1 {
        part p1 {
            attribute a11;
        }
        part p2 :> p1 {
            attribute a21 :>> a11 {
                /* comment */
            }
        }
    }
    package Case2 {
        part p1 {
            attribute a11;
        }
        part p2 {
            attribute a21 :>> p1.a11 {
                /* comment */
            }
        }
    }
    package Case3 {
        part p1 {
            attribute a11;
        }
        part p2 :> p1 {
            attribute a21 :>> p1.a11 {
            	/* comment */
            }
            attribute a22 = p1.a11;
        }
    }
```
The result in interconnection view looks like:
Screenshot 2024-04-19 at 12 59 43 AM

himi:
Even if we visualize,  for example,  `Case2::p2::a21` only (the comment is for this purpose in Eclipse), we get:
Screenshot 2024-04-19 at 1 01 46 AM

This does not render a spurious redefinition.

himi:
This PR also uses `FeatureUtil.getAllRedefinedFeaturesOf()` rather than iterating redefined features.  It is important to deal with implicit elements.   Therefore, we can properly handle 
```
package TestRedefinedFeatures {
    action a0 {
        action a01 {
            out output;
        }
        flow f1 from a01.output to a02.input;
        action a02 {
            in input;
        }
    }

    action a1 :> a0 {
        action :>> a02 {
        	in input;
        }
    }
}
```
The rendered result is:

Screenshot 2024-04-19 at 1 05 53 AM

Previously, we had to explicitly redefine `a1.a02.input` to correctly render it.

PR #566, opened 2024-05-26
ST6RI-771 With `start` action, some flow connections are not rendered properly (PlantUML)

himi:
Since `start` action is a standard library element having lots of features, the visualizer needed to process them, exceeding the limit of elements that can be processed, even though the `SHOWLIB` style is not selected.  This PR fixes the issue by preventing `VPath` from traversing standard library elements if `SHOWLIB` style is not specified.

This PR also adds a style configuration to `SuccessionFlowConnectionUsage` to properly render it and an error message shown in cases when the visualizer still hits the element limit.


himi:
After this fix, the example below:
```
package TestFlowConnection2 {
    action a0 {
        first start;
        then 
        action a01 {
            action a001 {
                out o1;
            }
            succession flow from a001.o1 to a002.i1;
            action a002 {
                in i1;
            }       
        }
    }
}
```

is properly rendered as:

Screenshot 2024-05-26 at 12 08 57 PM

himi:
I noticed this limitation hard to find if it happens.  I changed to show an error message when the visualizer hits the limitation.
So, if `SHOWLIB' style is enabled, the example above is rendered as:

Screenshot 2024-05-26 at 1 16 08 PM

seidewitz (approved):

PR #567, opened 2024-05-29
ST6RI-772 Shorthand notation successions are not rendered (PlantUML)

seidewitz:
This PR fixes a bug that was causing successions written using the "shorthand" SysML textual notation to not be rendered in PlantUML in Eclipse when a model file is initially opened. 

**Background**

Previously, when a file with the model
```
action def A {
  action a;
  then a;
}
```
was first opened, it was rendered as
image
However, if a change was made to the model text, and the model was re-rendered, it was shown correctly with the succession:
image

This bug was inadvertently caused by the changes made in PR #562, as a result of which transformations are no longer run when a file is first opened. Adding the implied related features of a "shorthand" succession was implemented using transformations in the `SuccessionAsUsageAdapter`, so these were no longer being added when a model file was open, so the successions could no be rendered correctly.

**Update**

The ends of a SysML `SuccessionAsUsage` are always `ReferenceUsages`, which are related to the related features of the succession using `ReferenceSubsettings`. Rather than adding implied `ReferenceSubsettings` to the ends of a `SuccessionAsUsage` using a transformation in the `SuccessionAsUsageAdapter`, the changes in this PR moves this to the implied specialization computation in the `ReferenceUsageAdapters` for the ends. This can then be triggered "on demand" in the `FeatureAdapter::getReferencedFeature` method, as for other kinds of `Feature` specializations.

The PR also ensures that the `source` is computed on-demand for a `TransitionUsage` (including a conditional succession) notated without an explicit source, because this is necessary for properly setting the source related feature of the `SuccessionAsUsage` nested in the `TransitionUsage`.


himi (approved):
I confirmed it worked.  Thank you.

himi:
Thank you.  I confirmed both of 
`training/16. Conditional Succession/Conditional Succession Example-2.sysml`
and 
`training/17. Control/Decision Example.sysml`
worked properly.

himi (approved):

PR #568, opened 2024-05-31
ST6RI-764 Remove the use of Gradle for building the Jupyter kernel

TheKorpos:
**Background**

Previously, the [Jupyter kernel installer](https://github.com/SpencerPark/Jupyter-kernel-installer-gradle) Gradle plugin has been used to generate the configuration files and scripts necessary to boot up the Jupyter kernel for SysML. However, this plugin is no longer maintained to the degree needed to properly support using newer Java versions. Additionally the mixing of Gradle and Maven in the current setup is undesirable as it complicates the build process.

**Changes**
 - Removed the Gradle build from the `jupyter-sysml-kernel` project.
 - Preserved the generated resources produced by the [Jupyter kernel installer](https://github.com/SpencerPark/Jupyter-kernel-installer-gradle) Gradle plugin.
 - Updated the Maven configuration to produce a shaded jar and a zip containing the installable Jupyter kernel.
 - Added `.sh` and `.bat` scripts that automatically build and install the Jupyter SysML kernel.
 - Updated documentation.


seidewitz:
@TheKorpos 

The dependency on `org.omg.sysml` in the `org.omg.sysml.jupyer.kernel` project POM was causing an error in my Eclipse workspace. I removed the dependency from the POM and, instead, added the `org.omg.sysml` project to the classpath for `org.omg.sysml.jupyer.kernel`. This eliminates the error, allows the kernel project to build within Eclipse and does not seem to effect the Maven build.

Please confirm that this change is acceptable.

seidewitz (approved):

TheKorpos:
@seidewitz
Yes, everything works as expected on my end too.

PR #569, opened 2024-06-08
ST6RI-760 Actor and stakeholder property derivations only return owned usages

seidewitz:
According to the Beta 2 version of the SysML v2 specification, the derivations of the `actorParameter` and `stakeholderParameter` properties for `RequirementDefinitions` and `RequirementUsages` and of `actorParameter` for `CaseDefinitions` and `CaseUsages` are supposed to included both owned and inherited actors and stakeholders. Previously, however, the implementation only returned owned actors and stakeholders. This PR corrects the implementation to be conformant with the specification.


ujhelyiz (approved):
Looks reasonable to me.

PR #573, opened 2024-06-20
ST6RI-778: Inherited elements of non-standard libraries are not shown (PlantUML)

himi:
Because of my mistake in ST6RI-771 (PR #566), inherited features of non-standard library are wrongly hidden.  So the example below:
```
package TestStdLib {
	part def P1 {
		part p1;
	}
	
	part p : P1 {
		part p2;
		connect p1 to p2;
	}	
}
```
is wrongly rendered as:
image


himi:
By this fix, it is correctly rendered as:

image

seidewitz (approved):

PR #574, opened 2024-06-20
ST6RI-780: Nested inherited features are not correctly rendered (PlantUML)

himi:
When the visualizer renders a connector, it also renders inherited features referred by the connector.  However, it may fail to render nested features.  In the example below:
```
package TestInheritedPart2 {
    part def Vehicle {
        part powerTrain {
            part shaft {
                port rotation;
            }
        }
    }

    part vehicle : Vehicle {
        part :>> powerTrain {
            part motor {
                part shaft {
                    port rotation;
                }
            }
            bind motor.shaft.rotation = shaft.rotation;
        }
    }
}
```
rendering just part `vehicle` gives:
image
which lacks `vehicle.powerTrain.shaft.rotation` even though it is referred to by the binding connector.

This PR also fixes `VSSRMember.isRedefining()` by using `FeatureUtil`.


himi:
After this fix, the example above is correctly rendered as:
image

PR #577, opened 2024-06-23
ST6RI-782: Memberships of inherited features are not correctly rendered; and ST6RI-784:  Subsettings of inherited feature from nested type are not rendered(PlantUML)

himi:
Since the fix in ST6RI-767 has a problem to find a target of PRelation, the example below:
```
package TestInheritedPart1 {
    part def Vehicle {
        part powerTrain;
    }
    part vehicle : Vehicle;
}
```
is wrongly rendered with `SHOWINHERIT` style as below:
image

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/InheritKey.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/InheritKey.java
@@ -35,6 +35,7 @@ import org.omg.sysml.lang.sysml.FeatureValue;
 import org.omg.sysml.lang.sysml.Membership;
 import org.omg.sysml.lang.sysml.Namespace;
 import org.omg.sysml.lang.sysml.Relationship;
+import org.omg.sysml.lang.sysml.Specialization;
 import org.omg.sysml.lang.sysml.Type;
 import org.omg.sysml.util.FeatureUtil;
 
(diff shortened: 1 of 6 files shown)


himi:
By this fix, the example above is correctly rendered:
Screenshot 2024-06-23 at 12 23 07 AM

himi:
Since this fix corrected the algorithm to handle redefinitions, I also checked about it.  For example, 

```
package TestRedef2 {
	action a1 {
      	action a11;       	
   	}
   	action a1x {
   		action a110;
   	}
   	action a2 :> a1;
    action a3 :> a2 {
    	action :>> a11;
    }
    action a4 :> a2 {
    	action :>> a11;
    }
    action a5 :> a2;
    action a6 :> a5, a1x {
    	action :>> a11;
    	action :>> a110;
    }
}
```
is correctly rendered as:

Screenshot 2024-06-23 at 12 25 35 AM

seidewitz:
For the `TestRedef2` example, the diagram is displayed the same even if the `SHOWINHERIT` style is _off_. I think it would actually be clearer in this case if all the `a11` redefinition arrows pointed to the `a11` symbol in action `a1`. It really is the case in the abstract syntax that, for all the `a11` redefinitions, the actual redefined action is `a1::a11`, and that's what I would expect to see _without_ `SHOWINHERIT`. Indeed, even if I change the textual notation to explicitly redefine `a1::a11`, it still shows the redefinition arrows pointing to the inherited `a11` symbols, which is a little confusing, particularly when `SHOWINHERIT` is off.

How difficult would this be to implement?

himi:
Thank you for your valuable suggestion.  I think I could do what you requested.  Without `SHOWINHERITED`, the result would be:
image

himi:
I also want to fix ST6RI-784 here if you are ok because Sandy reported and I would like to fix it quickly.  This is actually a silly mistake I made.

seidewitz:
> I also want to fix ST6RI-784 here if you are ok because Sandy reported and I would like to fix it quickly. This is actually a silly mistake I made.

Sure, add that to the PR.

himi:
In order to test redefining redefined target, I slightly improved `TestRedef2`:
```
package TestRedef2 {
	action a1 {
      	action a11;       	
   	}
   	action a1x {
   		action a110;
   	}
   	action a2 :> a1;
    action a3 :> a2 {
    	action :>> a11;
    }
    action a4 :> a2 {
    	action :>> a11;
    }
    action a5 :> a4 {
    	action :>> a11;
    }
    action a6 :> a2, a1x {
    	action :>> a11;
    	action :>> a110;
    }
}
```
The visualization result without `SHOWINHERITED` is:
image
and that with `SHOWINHERITED' is:
image

himi:
By f8bd28c,
```
package TestSubsetting1 {
    part p0 {
        perform a1;
        part p01 {
            perform a1.a01;
        }
    }
    action a1 {
        action a01;
    }
}
```
gives
image
This correctly renders a subsetting from `p0.p01.a01` to `p0.a1.a01`.  (Note that it should be a reference subsetting arrowhead but I need to update PlantUML to add new arrowhead symbols)

himi:
Also
```
package TestSubsetting2 {
    part p0 {
    	part p01 :> p1;
        part p02 {
            part p021 :> p01.p10;
        }
    }
    part p1 {
        part p10;
    }
}
```
is correctly rendered as:
image

seidewitz (approved):

PR #584, opened 2024-07-20
ST6RI-781: Show multiplicities in nodes (PlantUML)

himi:
This PR supports rendering multiplicities in nodes.  So far the visualizer renders multiplicities in edges only.  I added three new styles:

- NODEMULTIPLICITY
    Show multiplicities in nodes.
- EDGEMULTIPLICITY
   Show multiplicities in edges.
- IMPLICITMULTIPLICITY
   Show implicit multiplicities as well.  That means even if you do not explicitly specify multiplicities, the visualizer renders default multiplicities.

If you do not specify any styles, view choose proper multiplicity rendering.  For example, `Tree` view shows multiplicities in nodes by default.


himi:
This is just an example with `Tree` view:
```
package Vehicle {
	part def Vehicle {
		part tire[*];
	}
	
	part car : Vehicle {
		part :>> tire [4];
	}
	part bycicle: Vehicle {
		part :>> tire[2];
	}
	part unicycle: Vehicle {
		part :>> tire[1];
	}
}
```
This is rendered as below:
image

seidewitz (approved):
I also confirmed that
```
	part unicycle: Vehicle {
		part :>> tire;
		part tire1;
	}
```
when visualized using IMPLICITMULTIPLICITY, shows the correct default multiplicity of `[1]` for `tire1`:
image

PR #586, opened 2024-08-02
ST6RI-786 Update JSON Schema for the 2024-07 baseline

ivan-gomes:


seidewitz:
It seems I neglected to regenerate the `kerml.ecore` file, so the KerML JSON does not include the KerML abstract syntax updates (the SysML JSON does). I have now regenerated the metamodel `.uml` and `.ecore` files, including `kerml.ecore` (on this branch). Please regenerate the JSON files using the new `.ecore` files. (There should be no change to the SysML JSON, but you can regenerate that too as a check.)

ivan-gomes:
> It seems I neglected to regenerate the `kerml.ecore` file, so the KerML JSON does not include the KerML abstract syntax updates (the SysML JSON does). I have now regenerated the metamodel `.uml` and `.ecore` files, including `kerml.ecore` (on this branch). Please regenerate the JSON files using the new `.ecore` files. (There should be no change to the SysML JSON, but you can regenerate that too as a check.)

Done. In the process, we have identified another spot that could use a fixed ordering: subclasses. But I think we should address that in the next release so we don't have to revalidate from the top.

PR #587, opened 2024-08-09
ST6RI-785 Implemented performance improvement using a generated index of the library models

TheKorpos:
**The following features have been implemented:**
- Right click menu action in Eclipse to generate library indexes for projects
- Headless application to generate library indexes
- Added check to the `KerMLScope` that prevents long searches in library namespaces in case the searched name is not resolvable from the namespace based on the pre-calculated data stored in the library index
- Added default index loading/caching logic for headless applications
- Added eclipse specific index loading/caching logic
- Updated the maven build to auto-generate the library index for `sysml.library`
- Updated CI build to ensure that the last up-to-date generated library index is also stored in the git repository


seidewitz:
I corrected the copyright headings in several files. Please check that all new files correctly have the correct copyright holder and that the copyright date has been updated on all revised files.

seidewitz:
Do the Xpect tests run using the library index? Each of `org.omg.kerml.xpect.tests` and `org.omg.sysml.xpect.tests` have there own library directories separate from `sysml.library`. But perhaps the index is still loaded from `sysml.library` when they are run.

seidewitz (changes requested):
Two minor points:
- When the index is generated in Eclipse, it first does a build and then generates the index. After the build is complete, the progress indication says at 100% for a noticeable amount of time (presumably while the index is actually being generated). Would it be possible to provide additional progress monitoring during this time, or at least some indication of what is happening ("Generating..." or "Finalizing..." or something like that)?
- The installation instructions in the README for the SysML v2 Pilot Implementation repository call for a clean build on `sysml.library`. Should this now be a library index generation here? Or is just a build still sufficient, since the index will be saved in the repository?

ujhelyiz:
> I corrected the copyright headings in several files. Please check that all new files correctly have the correct copyright holder and that the copyright date has been updated on all revised files.

I have also checked them, nothing seems to be missing, but I have added Laszlo's name to have the same format as others.

> Do the Xpect tests run using the library index? Each of org.omg.kerml.xpect.tests and org.omg.sysml.xpect.tests have there own library directories separate from sysml.library. But perhaps the index is still loaded from sysml.library when they are run.

Right now, those tests are not using the index file as it is not found where it is available. I see two ways to make sure it is used: either copy manually the index.json files from the normal sysml.library project to the copies, however, this will increase the repository size greatly (as the index file is quite large); or we could use Maven to copy that file before the tests are run. @seidewitz Which option seems better to you?

> When the index is generated in Eclipse, it first does a build and then generates the index. After the build is complete, the progress indication says at 100% for a noticeable amount of time (presumably while the index is actually being generated). Would it be possible to provide additional progress monitoring during this time, or at least some indication of what is happening ("Generating..." or "Finalizing..." or something like that)?

I have replaced the progress mechanism, in my testing the progress bar finishes together with the task, but for some reason the status message I have tried to add does not seem to show up, so in the latter half only the progress bar is growing, but no textual information is present. @seidewitz do you think this is appropriate or should I look more into the textual display?

> The installation instructions in the README for the SysML v2 Pilot Implementation repository call for a clean build on sysml.library. Should this now be a library index generation here? Or is just a build still sufficient, since the index will be saved in the repository?

Considering the index file is always up-to-date in the repository (otherwise the build fails), the clean build should be enough.

TheKorpos:
@seidewitz,

@ujhelyiz has updated the progress bar. It shows what step is executing and the bar is divided into smaller segments so it tracks the progress much better.

> I think the best option for this would be to have the Maven build copy the file before the tests are run. However, would it be necessary to update the XPECT_SETUP for the ResourceSet/Workspace in every test file, in order for the index to be found during test runs? If so, then it is probably not worth it (unless you can easily make this update in an automated way).

Unfortunately, copying the index file is not enough. The default logic (`PrecalculatedLibraryIndexProvider`) expects library resources to be in a folder named 'sysml.library' and searches for the index there. I've tried modifying this logic in the xpect tests, which can be done fairly easily, so we don't have to update all the test files but I ran into another problem. It looks like that xpect creates a logical copy of the referenced resources with a URI that doesn't match with their file path.
E.g
`/org.omg.kerml.xpect.tests/library/Base.kerml`
becomes
`/org.omg.kerml.xpect.tests/src/org/omg/kerml/xpect/tests/visibility/Base.kerml`

This makes it a bit more difficult to link library resources to their index. I think the easiest way is listing all the files from the library folders (library, library.system, library.kernel...) before test execution and using that list to decide which resource is a library resource and then get the index from a set location for them.

seidewitz:
@TheKorpos 
Yeah, Xpect is tricky with resources. Let's not worry about that for now. We can figure out how to integrate the index into the Xpect testing later.

seidewitz (approved):

PR #589, opened 2024-09-04
ST6RI-791 Remove ability to have import declarations without explicit visibility

seidewitz:
This PR removes the allowance of deprecated default-public import and expose declarations, included in PR #582 for release 2024-07. It includes the following changes:

1. Removed the manual code changes in the abstract syntax model to give imports a default multiplicity of **`public`**, restoring the normative default of **`private`**.
2. Removed KerML validation warning of use of a deprecated default-public import declaration.
3. Updated the KerML and SysML textual notation grammars so that a visibility indicator is required on import declarations and expose declarations are parsed as always having **`protected`** visibility.
4. Added KerML and SysML Xpect tests related to import and expose visibility (including testing the validation that the visibility of a top-level import declaration is **`private`**).


ujhelyiz (approved):
Looks reasonable to me.

PR #592, opened 2024-09-16
ST6RI-792: Show metadata in compartments as well

himi:
So far, metadata is rendered in distinct callout nodes only even when we use compartments.  By this fix, we render metadata in compartments if metadata is owned by an element that is rendered with compartments.


himi:
For example:
```
ackage TestMetadata1 {
	metadata def MD1 {
		attribute a11;
		attribute a12;
	}
	metadata def MD2 {
		@MD1 { a11 = "a00"; }
		attribute a21;
		attribute a22;
	}
	part p1 {
		@MD1 { a11 = "aaa"; a12 = "bbb"; } 
		@MD2 { a21 = "ccc"; }
	}
}
```
is rendered as:
Screenshot 2024-09-15 at 8 28 08 PM

himi:
This PR also fixes the confusing notation of `#-` name.  We use `#name` notation only if the metadata is SemanticMetadata. So in the example below, only part `p2` is rendered with `<<#MD2>>`.
```
package TestMetadata2 {
	metadata def MD1 {
		attribute a11;
	}
	metadata def MD2 :> Metaobjects::SemanticMetadata {
		attribute a21;
	}
	part p1 {
		@MD1 { a11="a"; }
	}
	#MD2 part p2 {
	}
}
```
Screenshot 2024-09-15 at 11 03 09 PM

seidewitz:
> But this treatment is not perfectly aligning with the textual notations. Even if we specify `#MD1 part p3;`, it is not rendered with `<<#MD1>>`. Also if we specify `part p4 { @MD2; }`, it is rendered with `<<#MD2>>`. TTBOMK, we have no way to distinguish them from the AST.

I am not sure that using the `<<#MD1>>` notation for semantic metadata is the best decision. The `#` notation is not inherently tied to semantic data. In both the textual and graphical notation, it is simply a notation option for a metadata annotation with no nested feature bindings.

While it is correct that there is no way to determine from the abstract syntax what option was used in the textual notation, I would suggest the following rule for automatic rendering of the graphical notation: If a metadata annotation has no nested features, then use the `<<#...>>` notation. Otherwise, display it using a compartment or a callout.

himi:
I thought `<<#-name>>` notation should represent a kind of metatype because we use guillemet to represent metaclass or metatype. Therefore, I thought that notation should be used for `SemanticMetadata` because it is used for language extension.   However, your proposal seems reasonable.  I implemented it and the example below:

```
package TestMetadata2 {
	metadata def MD1 {
		attribute a11;
	}
	metadata def MD2 {
		attribute a21;
	}
	part p1 {
		@MD1 { a11="a"; }
	}
	#MD2 part p2 {
		@MD1 {a11 ="c";}
	}
	part p3 {
		@MD1;
		@MD2 { a21 = "d"; }
	}
}
```
is rendered as:
Screenshot 2024-09-16 at 8 29 58 PM

I think it would be confusing part `p3` is rendered with `<<#MD1>>` but we can use `@` notation even if it's SemanticMetadata.  I think it's acceptable.  Thank you for the suggestion.

seidewitz (approved):

PR #593, opened 2024-10-03
ST6RI-794: Specify UTF-8 encoding to access index files.

himi:
Previously, `LibraryIndex` read and wrote index files without specifying encoding. Therefore, if the default encoding of the JRE was not UTF-8, those files could be corrupted. In a Japanese Windows environment, the default encoding is CP932, and the library index was corrupted because it could not encode all of the Unicode characters used in the SysMLv2 libraries.

This PR updates `LibraryIndex` to use UTF-8 encoding. It also fixes an issue with constructing a proper index filename at PrecalculatedLibraryIndexProvider.getIndexFile(). To convert an EMF URI to a filename, `toFileString()` must be used.

### DIFF
--- a/org.omg.kerml.xtext/src/org/omg/kerml/xtext/library/PrecalculatedLibraryIndexProvider.java
+++ b/org.omg.kerml.xtext/src/org/omg/kerml/xtext/library/PrecalculatedLibraryIndexProvider.java
@@ -27,7 +27,6 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
-import java.util.Arrays;
 
 import org.apache.log4j.Logger;
 import org.eclipse.emf.common.util.URI;
--- a/org.omg.kerml.xtext/src/org/omg/kerml/xtext/library/PrecalculatedLibraryIndexProvider.java
+++ b/org.omg.kerml.xtext/src/org/omg/kerml/xtext/library/PrecalculatedLibraryIndexProvider.java
@@ -61,7 +61,7 @@ public class PrecalculatedLibraryIndexProvider implements ILibraryIndexProvider 
 	@Override
 	public LibraryIndex getIndexFor(Resource resource) {
 		
-		if (disabled) {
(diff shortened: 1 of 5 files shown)


org.omg.kerml.xtext/src/org/omg/kerml/xtext/library/PrecalculatedLibraryIndexProvider.java
```diff
@@ -78,13 +78,13 @@ public LibraryIndex getIndexFor(Resource resource) {
...
 			if (indexFile == null || !indexFile.exists()) return LibraryIndex.EMPTY_INDEX;
 			
-			try (FileReader fileReader = new FileReader(indexFile)){
-				
-				index = LibraryIndex.fromJson(fileReader);
-				
+			try {
+				FileInputStream fis = new FileInputStream(indexFile);
+				InputStreamReader isr = new InputStreamReader(fis, StandardCharsets.UTF_8);				
+				index = LibraryIndex.fromJson(isr);
```
seidewitz, line 84: Previously, the `fileReader` declaration was in a try-with-resources. The `FileInputStream fis` and `InputStreamReader isr` declarations are now within the try body, but they don't ever seem to be closed. Shouldn't the be declared in a try-with-resources, like `fileReader` was? I.e.,
```
try (FileInputStream fis = new FileInputStream(indexFile);
       InputStreamReader isr = new InputStreamReader(fis, StandardCharsets.UTF_8)) {				
    index = LibraryIndex.fromJson(isr);
} catch ...
}
```

seidewitz (changes requested):

org.omg.kerml.xtext/src/org/omg/kerml/xtext/library/PrecalculatedLibraryIndexProvider.java
himi: Opps.  You are right.  I learned try-with-resource in the document:
https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

seidewitz (approved):

PR #595, opened 2024-10-18
ST6RI-796 Leftover caches in InvocationDelegates

TheKorpos:
[Feature_effectiveShortName_InvocationDelegate](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/blob/d20a563968c7b78616ee80481829e0e4f6aee2e1/org.omg.sysml/src/org/omg/sysml/delegate/invocation/Feature_effectiveShortName_InvocationDelegate.java#L43) and [Feature_effectiveName_InvocationDelegate](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/blob/d20a563968c7b78616ee80481829e0e4f6aee2e1/org.omg.sysml/src/org/omg/sysml/delegate/invocation/Feature_effectiveName_InvocationDelegate.java#L43) cache their result. The caches are stored by the Invocation Delegates and they are never invalideted. This can lead to values getting stuck or possibly even memory leaks.

- Moved `effectiveName` cache from `Feature_effectiveName_InvocationDelegate`to `FeatureAdapter`
- Moved `effectiveShortName` cache from `Feature_effectiveShortName_InvocationDelegate `to `FeatureAdapter`
- Added cache invalidation
- Moved calculation logic from the Invocation Delegates to `FeatureUtil`

### DIFF
--- a/org.omg.sysml/src/org/omg/sysml/adapter/FeatureAdapter.java
+++ b/org.omg.sysml/src/org/omg/sysml/adapter/FeatureAdapter.java
@@ -107,6 +107,24 @@ public class FeatureAdapter extends TypeAdapter {
 	// Caching
 	
 	EList<Type> types = null;
+	String storedEffectiveName = null;
+	String storedEffectiveShortName = null;
+	
+	public void storeEffectiveName(String effectiveName) {
+		storedEffectiveName = effectiveName;
+	}
+	
+	public String getEffectiveName() {
+		return storedEffectiveName;
+	}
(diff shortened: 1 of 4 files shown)


seidewitz:
I removed the use of `FeatureImpl` in `FeatureUtil`. This was left over from moving old code into the invocation delegate in the first place, and is not needed for the functionality as currently implemented.

seidewitz (approved):

PR #597, opened 2024-11-13
ST6RI-800 Terminate Actions

seidewitz:
This PR implements a new capability for terminate actions in SysML v2, as specified in the resolution to the following issue approved on Ballot 4 of the SysML v2 Finalization Task Force.

- [SYSML2_-44](https://issues.omg.org/issues/SYSML2_-44) Transformation of UML4SysML::ActivityFinalNode is not specified yet

**Language**

A terminate action usage is used to terminate the performance of some other action. It has the form

**`action`**_`name`_**`terminate`**_`expression`_`;`

The action _`name`_ (or other declaration information) is optional. The _`expression`_ is evaluates to the action to be terminated. The terminated action ends its performance by the completion of the terminate action. 

**`action`**`outerAction {`
    **`in action`**`a;`
    **`action`**`term_a`**`terminate`**`a; // terminates a`
    **`then terminate`**`; // terminates outerAction`
`}`

A terminate action usage may also be used to terminate a non-action occurrence. In this case, the lifetime of the terminated occurrence must end by the completion of the terminate action. That is, the terminate action usage effectively "destroys" the terminated occurrence.

**`part`**`container {`
    **`part`**`component : Component[0..1];`
    **`action`**`act {`
        **`assign`**`component := Component(); // Create component`
        **`then terminate`**`component; // Destroy component`
    `}`
`}`

**Model Library**

The following elements are added to the Systems Library model `Actions`:

- **`action def`**`TerminateAction` – The base type for all terminate action usages.
- **`action`**`terminateActions` – The base feature for all terminate action usages. It also provides the default for the terminated occurrence as the featuring occurrence of the terminate action (which will generally be a containing action).

The following feature is added to the base action definition `Action`:

- **`action`**`terminateSubactions` – The subactions of an action that are terminate actions.


seidewitz:
@ivan-gomes @manasbajaj 
This PR adds `TerminateActionUsage` to the SysML abstract syntax. The API implementation needs to be updated for this. The `SysML.uml` and `SysML.ecore` files are updated on the ST6RI-800 branch.

Weilkiti (approved):
Looks good; I cannot identy any difference to the specification.

PR #600, opened 2024-12-03
ST6RI-801 SuccessionAsUsage of TransitionUsage seems incorrect

seidewitz:
This pull request fixes a bug in which a `TransitionUsage` parsed with a feature chain did not correctly reflect this when `getSource` was called on it. This was because the `TransitionUsageAdapter::computeSource` method assumed that a source parsed for a `TransitionUsage` would be referenced by a non-owning `Membership`, while a feature chain is owned using an `OwnedMembership`. Therefore, `computeSource` used the shorthand rule to fill in the source as the lexically immmediately preceding state.

For example, the model 
```
state s0 {
	state s1 {
		state s2;
	}
	transition t1 first s1.s2 then s3.s4;
	state s3 {
		state s4;
	}
}
```
was previously rendered by PlantUML as
image
because the transition source has been set to `s1`, the state lexically previous to `t1`.

This PR changes the condition in `computeSource`to assume that no source has been parsed for the `TransitionUsage` if the first `ownedMembership` is a `ParameterMembership`, because in the shorthand notation without a source, a `TransitionUsage` is always parsed with an initial empty parameter. With this change, the above model renders properly as
image

### DIFF
--- a/org.omg.sysml.xpect.tests/src/org/omg/sysml/xpect/tests/simpletests/StateTest.sysml.xt
+++ b/org.omg.sysml.xpect.tests/src/org/omg/sysml/xpect/tests/simpletests/StateTest.sysml.xt
@@ -88,6 +88,16 @@ package StateTest {
 		exit act;
 	}
 	
+	state s0 {
+  		state s1 {
+    		state s2;
+  		}
+  		state s3 {
+  			state s4;
+  		}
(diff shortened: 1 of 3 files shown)


himi:
@seidewitz , thank you for fixing it so quickly.  I tried transitions from more deeply nested state and they worked correctly.

himi (approved):

PR #601, opened 2024-12-04
ST6RI-802: TransitionUsages may be wrongly rendered (PlantUML)

himi:
When rendering `TransitionUsage`, the visualizer uses its `SucessionAsUsage` as a connector.  The code for TransitionUsage to resolve end features was not updated previously, even though the algorithm to resolve end features had been fixed and updated. This PR removed the wrong code that was only used for `TransitionUsage`.

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VPath.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VPath.java
@@ -550,13 +550,6 @@ public class VPath extends VTraverser {
         return "";
     }
 
-    private String addContextForTransitionUsageEnd(Feature tu, Feature end) {
-        PC pc = makeEndFeaturePC(end);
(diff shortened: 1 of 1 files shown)


himi:
Previously, in the example below, 
```
package TestTransition1 {
	state s0 {
  		state s1;
  		transition t1 first s1 then s0;
  		succession sc1 first s1 then s0;
	}
}
```
`t1` was wrongly rendered while `sc1` was correctly rendered:
image

After this fix, 
both of them are correctly rendered:
Screenshot 2024-12-03 at 9 41 32 PM

himi:
By merging with ST6RI-801, I could correctly render the example below as well
```
state s0 {
  		state s1 {
    		state s2;
  		}
  		transition t1 first s1.s2 then s0;
  		succession sc1 first s1.s2 then s0;
	}
```
Screenshot 2024-12-03 at 9 43 08 PM

seidewitz (approved):

PR #610, opened 2025-01-09
ST6RI-813 Transfers across interfaces (SYSML2_-39)

seidewitz:
This PR implements the resolution to the following issue, as approved in SysML v2 FTF2 Ballot 6.

- [SYSML2_-39](https://issues.omg.org/issues/SYSML2_-39) Semantics of transfers across interfaces

This resolution updates the Systems Library models `Ports.sysml` and `Interfaces.sysml` to specify the semantics of automatically targeting a transfer whose source is a `Port` at one end of an `Interface` to a `Port` at a different end of the `Interface`.

### DIFF
--- a/org.omg.sysml.xpect.tests/library.systems/Interfaces.sysml
+++ b/org.omg.sysml.xpect.tests/library.systems/Interfaces.sysml
@@ -1,48 +1,89 @@
 standard library package Interfaces {
-	doc
-	/*
-	 * This package defines the base types for interfaces and related structural elements in the SysML language.
-	 */
-	 
-	private import Connections::Connection;
-	private import Connections::connections;
-	private import Connections::BinaryConnection;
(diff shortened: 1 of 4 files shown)


conradbock:
The changed files under sysml.library/Systems Library are identical to SYSML_-39's revised text, but the ones under xpect.tests are slightly different (xpect/Ports.sysml still has `outgoingTransfer`, without `FromSelf`, and xpect/Interfaces.sysml differs in whitespace in one of the docs). Can approve the PR if this doesn't matter.

seidewitz:
The Xpect tests basically just test syntax, not semantics (other than implicit relationships). So the differences in these model files isn't going to matter. Nevertheless, we might as well be consistent, so I have updated the files in the library copy used for Xpect.

Good catch!

conradbock (approved):

PR #611, opened 2025-01-10
ST6RI-811 Additional issue resolutions from KerML FTF2 Ballot #4 (KERML_99, KERML_-117, KERML_-119)

seidewitz:
This PR implements resolutions to the following issues (to be) approved in KerML FTF2 Ballot 4:

- [KERML_-99 ](https://issues.omg.org/issues/KERML_-99)CheckFeatureChainExpressionResultSpecialization constraint needs to be defined
   - Adds a new semantic constraint to the specification for the result of a feature chain expression. The implicit specialization for this had already been implemented, but the implementation has now been updated so the specialization targeted a feature chain, per the approved constraint.
- [KERML_-117](https://issues.omg.org/issues/KERML_-117) InvocationExpresion::result specialization
   - Adds new semantic constraints to the specification for the results of a select expression or index expression. Implicit specializations for these had already been implemented. However, the resolution also added a new metaclass `IndexExpression`, which has now also been implemented, and the implementation of the implied specialization of its result has been moved to the new `IndexExpressionAdapter`.
- [KERML_-119](https://issues.omg.org/issues/KERML_-119) Navigability of AnnotatingElement<->Annotation meta-association
   - Revises a bidirectional meta-association between `AnnotatingElement` and `Annotation` so that both end properties are derived. Also requires that an `Annotation` relationship be owned by either its `annotedElement` or its `annotatingElement` (which is consistent with how such relationships are parsed from the textual notation).

## Concrete Syntax

There are no changes to the concrete syntax, and all abstract syntax changes are backward compatible with existing notation.

## Abstract Syntax

Metaclass added:

- `IndexExpression`

Metaclasses changed:

- `FeatureChainExpression`
- `AnnotatingElement`
- `Annotation`

## Model Libraries

No changes.

## Technical Update

Previously, the the list of implicit general types of a type was ordered for a specific kind of specialization (e.g., "Subclassification", "Subsetting", "FeatureTyping", etc.), but was not ordered a cross the different kinds. This resulted in a possibile non-determinism in the building of the library index, which could cause a build to fail. To avoid this, this PR includes an update to sort the implicit general types by the `eClassifierIds` (which are integers) of their kind keys (which are `EClasses`).


seidewitz:
@TheKorpos 
Note the technical update.

TheKorpos (approved):
I have tested [KERML_-117](https://issues.omg.org/issues/KERML_-117) and it works as intended. The semantic constraint is satisfied: the subsetting is created in the non-array cases.

I have tested [KERML_-99](https://issues.omg.org/issues/KERML_-99) and it works as well.

For [KERML_-119](https://issues.omg.org/issues/KERML_-119) I haven't performed runtime testing. In the code changes I couldn't find anything that looks out of place or incorrect.

As for the techical update I belive that the changes address the issue and the undeterminism is gone now I have generated numerous indexes and the output is always the same.

PR #616, opened 2025-02-04
ST6RI-682 Implement loading of models from repository

TheKorpos:
This PR adds basic model download capability using the standard API.

### Overview
The model download functionality can be used
1. In Jupyter using `magic` commands
2. In Eclipse IDE as a "right click" action on projects
3. As command line java application

In Jupyter, the downloaded models are stored in memory and can be used throughout the work session. The Eclipse and command-line versions serialize the model into separate XMI files (`.sysmlx`). In the Eclipse IDE `.kermlx` and `.sysmlx` files are now added to the Xtext index so their content can be referenced using the textual syntax.

### Common

- `.kermlx` parsing and indexing: 
   - Added `org.omg.kerml.xtext.xmi`, `org.omg.kerml.xtext.ui.xmi`
- `.sysmlx` parsing and indexing: 
   - Added `org.omg.sysml.xtext.xmi`, `org.omg.sysml.xtext.ui.xmi`
- APIs for model download, JSON to EMF transformation
   - `org.omg.sysml.util.repository.ProjectRepository`: repository access, download
   - `org.omg.sysml.xtext.ui.util.RepositoryContentFetcher`: JSON to EMF transformation, XMI serialization (`.sysmlx`)

### Jupyter

New magic commands:

```
%projects

Print the name and identifier of all projects in the repository.
```

```
%load [--id=<PROJECT ID>] [--name=<NAME>] [--branchid=<BRANCH ID>] [--branch=<BRANCH_NAME>] [<NAME>]

Download previously published models from a project in the repository. (Use %projects to view repository contents.)
Named elements of the downloaded models may then be referenced by models in the notebook.
<NAME> is the full name of the project.
If <PROJECT ID> is given, then the project with that UUID is loaded. In this case, the <NAME> must not be given.
If <BRANCH NAME> or <BRANCH ID> is given, then the model is loaded from this branch of the project.
If no <BRANCH NAME> or <BRANCH ID> is given, the default branch is used.
If <BRANCH ID> is given, then a <BRANCH NAME> must not be given.
```

### Eclipse

- Added `SysMLv2` menu to the popup menu for projects, with the following options:
  - `Generate Library Index`: moved under `SysMLv2` menu.
  - `Pull model from API`: download models from a repository and saves in `.sysmlx`
     - Requires that  `.settings/org.omg.sysml.remote.properties` exists with the following contents:
       ```
       base.url=repository url
       remote.projectId=project id
       remote.branchId=branch id
       ```
[...]

PR #633, opened 2025-03-10
ST6RI-178 Versioned publishing

TheKorpos:
This PR is built on [https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/616](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/616).

The PR adds basic versioned publishing capability to Jupyter using the standard API.

### Summary

Currently ` %publish` command in Jupyter creates a new project on the API server and then it posts an all additions commit as the one and only commit to the 'main' branch every time it is called. The changes inlcuded in this PR allow users to target a specific project on the API server and to post incemental changes based on the delta between the remotely stored model and the local their local models.

### Changes

- Introduced new Java APIs to provide common functionality for API server interactions (`org.omg.sysml.util.repository`)
  - `ProjectRepository`: represents a standard API server
  - `RemoteProject`: represents a project that exists on the standard API server
  - `RemoteBranch`: represents a branch that exists in a project
  - `Revision`: represents a commit on a specific branch. Can be local only (e.g future initial commit)
  - `APIModel`: REST API compatible representation of the model
  - `APIModelDelta`: differences between two APIModels
 - Added logic to create APIModels from EMF models
   - Updated `JsonElementProcessingFacade`: now it can be configured to produce either an `APIModel` or an `APIModelDelta` (all additions).
   - Updated `ApiElementProcessingFacade` to use new Java APIs
 - Added logic to synch EMF models with APIModels
   - `EMFModelRefresher`: creates an `EMFModelDelta` from an EMF model and from an APIModel. `EMFModelDelta` can be used to refresh the EMF model.

### Jupyter changes
- Modified arguments of the `%publish` magic command.
  - Added `--project` to specify the target project, if none passed logic falls back to the old functionality, if project doesn't exist a new one is created with the specified name
  - Added `--branch` (experimental) to specify a branch. if none passed the default branch is used
 - Modified arguments of the `%load` magic command
   - Added  `--branch` (experimental) to specify a branch. if none passed the default branch is used

### Eclipse changes
 - Added optional property `remote.branchId` to `.settings/org.omg.sysml.remote.properties`

### CLI Apps
 - Updated cli apps to use new Java APIs and added new java args to support new functionality


TheKorpos:
Our testing has shown some issues worth addressing as a part of this PR.

 ### 1. The Jupyter publish/load doesn't support non-standard library packages as root namespaces.
During the upload process we have to create a number of proxy elements for the standard library. These elements will count as root elements (no owner). To account for these elements the download logic skips them when it processes project roots. When a namepsace is selected in Jupyter for `publish `the selected namespace is forced as a project root. So that namespace is a library element and a root element so it's skipped during `load`.

To address this LibraryPackages that are not standard library packages should also be allowed as root elements.

### 3. Derived fields significantly increase the size of the .json data and lead to longer processing time on the API server
Some derived fields like `member` are long lists as they contain all inherited elements from the library. This increases the number of proxy elements we have to add to the model to meet data integrity constraints. The derived fields and the proxies significantly inflate the data being sent to the API Server and lead to longer processing time as well.

While the CLI utilities the inclusion of derived fields can be controlled by the `'-d'` flag, Jupyter `publish` always includes them. The default behavior should be not to include derived fields but to support keep support for them the` '-d'` flag should be added to the `publish` command.

org.omg.sysml/src/org/omg/sysml/util/repository/APIModel.java
```diff
@@ -0,0 +1,253 @@
...
+	
+	/**
+	 * <b>NOT IMPLEMENTED</b>
+	 * <br>
+	 * Applies changes to the model.
+	 * 
+	 * @param delta changes to apply.
+	 */
+	public void merge(APIModelDelta delta) {
+		throw new UnsupportedOperationException("Not impelemnted");
```
ujhelyiz, line 121:
```suggestion
		throw new UnsupportedOperationException("Not implemented");

```

org.omg.sysml/src/org/omg/sysml/util/repository/APIModel.java
```diff
@@ -0,0 +1,253 @@
...
+		public boolean isLibraryElement() {
+			return isLibraryElement;
+		}
+
+		public UUID getUUID() {
+			return get(ProjectRepository.ID_FIELD);
+		}
+		
+		@Override
+		public boolean equals(Object o) {
```
ujhelyiz, line 238: When overriding Object#equals, also update hashcode as well to make sure that the generic contract of that method. Failing to do so could cause issues e.g. when using the objects in HashSets and HashMaps

org.omg.sysml/src/org/omg/sysml/util/repository/APIModelDelta.java
```diff
@@ -0,0 +1,226 @@
...
+		return deletions.isEmpty() && additions.isEmpty() && changes.isEmpty();
+	}
+	
+	/**
+	 * Converts the {@link APIModelDelta} to a data structure
+	 * that the standard API server accepts as a model delta.
+	 * 
+	 * @return list model changes acceptable by the API server
+	 */
+	public List<DataVersion> toTrasferableDelta() {
```
ujhelyiz, line 97: Misspelled method name.

org.omg.sysml/src/org/omg/sysml/util/repository/APIModelDelta.java
```diff
@@ -0,0 +1,226 @@
...
+	}
+	
+	/**
+	 * Creates a {@link Data} object which contains changed fields and their new values
+	 * 
+	 * @param thisElement element to compare with baseline
+	 * @param baselineElement element to use as a baseline
+	 * @return changed fields and their new values wrapped in a {@link Data} object.
+	 */
+	public static ChangedElement createElementDelta(Element thisElement, Element baselineElement) {
```
ujhelyiz, line 192: The name `thisElement` is strange here, as `this` is an entire delta. One option would be to rename it (consistently with other uses) to simply element (remove the prefix). Other alternatives might be to prefix it as `current` or `head`.

org.omg.sysml/src/org/omg/sysml/util/repository/EMFModelRefresher.java
```diff
@@ -0,0 +1,257 @@
...
+		{
+			if (!feature.isDerived() && remoteElement.get(feature.getName()) != null) {
+				if (feature instanceof EReference reference && reference.isContainment()) {
+					transformStructuralFeature(langElement, remoteElement, reference, true, true);
+				}
+			}
+		}
+	}
+	
+	private void tranformCrossreferences(EObject langElement, Element remoteElement) {
```
ujhelyiz, line 163: Mispelled method name, I'd use tran**s**formCross**R**eferences instead.

org.omg.sysml/src/org/omg/sysml/util/repository/EMFModelRefresher.java
```diff
@@ -0,0 +1,257 @@
...
+				}
+			}
+		}
+		
+		if (langElement instanceof FeatureTyping ft) {
+			ft.setTypedFeature(null);
+		}
+	}
+	
+	private void tranformAttributes(EObject langElement, Element remoteElement) {
```
ujhelyiz, line 178: Should be tran**s**formAttributes.

org.omg.sysml/src/org/omg/sysml/util/repository/EMFModelRefresher.java
```diff
@@ -0,0 +1,257 @@
...
+		{
+			if (!feature.isDerived() && remoteElement.containsKey(feature.getName())) {
+				if (feature instanceof EAttribute reference) {
+					transformStructuralFeature(langElement, remoteElement, reference, false, false);
+				}
+			}
+		}
+	}
+	
+	private static Collection<EStructuralFeature> getStructuralFeatuers(EObject langElement){
```
ujhelyiz, line 189: Mispelled method name: should be getStructuralFeatu**re**s

org.omg.sysml/src/org/omg/sysml/util/repository/EMFModelRefresher.java
```diff
@@ -0,0 +1,257 @@
...
+		Set<EObject> allRedefined = eClass.getEAllStructuralFeatures().stream()
+				.map(f -> f.getEAnnotation("redefines")).filter(Objects::nonNull)
+				.flatMap(a -> a.getReferences().stream()).collect(Collectors.toSet());
+		
+		return eClass.getEAllStructuralFeatures().stream().filter(f -> !allRedefined.contains(f)).collect(Collectors.toList());
+	}
+	
+	@SuppressWarnings("unchecked")
+	private void transformStructuralFeature(EObject langElement,  Element remoteElement, EStructuralFeature feature, boolean isReference, boolean isContainment) {
+		if (!disabledStructuralFeatures.contains(feature.getName())) {
```
ujhelyiz, line 201: Please review whether this can be simplified in some way, the very deep nesting makes this code very hard to read.

org.omg.sysml/src/org/omg/sysml/util/repository/EMFModelRefresher.java
```diff
@@ -0,0 +1,257 @@
...
+					} 
+				} else {
+					Object value = remoteElement.get(feature.getName());
+					if (value != null) {
+						langElement.eSet(feature, isReference? resolveReference(value, isContainment) : transformAttributeValue(value, feature));
+					}
+				}
+			}
+			catch (Exception e) {
+				System.out.printf("Unable to set structural feature %s::%s %n", feature.getEContainingClass().getName(), feature.getName());
```
ujhelyiz, line 222: Is there no reasonable alternative to error handling? This basically ignores the fact that an exception was thrown (any kind!), but can write the results to the standard output of the application that is not really meaningful in an Eclipse or Jupyter environment.

org.omg.sysml/src/org/omg/sysml/util/repository/EMFModelRefresher.java
```diff
@@ -0,0 +1,257 @@
...
+	}
+	
+	private EObject resolveReference(Object referenceValue, boolean isContainment) {
+		if (referenceValue instanceof Map referenceObjectMap) {
+			Object refUUID = referenceObjectMap.get("@id");
+			Element referencedElement = model.getElement(UUID.fromString(refUUID.toString()));
+			EObject referencedLangElement = isContainment? transform(referencedElement) : tracker.get(UUID.fromString(refUUID.toString()));
+			return referencedLangElement;
+		}
+		return null;
```
ujhelyiz, line 234: Can this be replaced by an Optional or some other structure? This design puts the burden on the caller to always check for null, but does not helps to ensure it is done.

org.omg.sysml/src/org/omg/sysml/util/traversal/facade/impl/JsonElementProcessingFacade.java
```diff
@@ -1,292 +1,294 @@
...
+		apiElement.put("@id", element.getElementId());
+		boolean isLibraryElement = ElementUtil.isStandardLibraryElement(element);
+		for (EStructuralFeature feature: eClass.getEAllStructuralFeatures()) {
+			String className = eClass.getName();
+			String featureName = feature.getName();
+			//always add the importedElement derived field. This field is used for a workaround that addresses
+			//unstable library membership UUIDs in the EMFModelRefresher.
+			if ((this.isIncludeDerived() || !feature.isDerived() || "importedElement".equals(feature.getName())) && 
+					// Skip implementation-specific features.
+					!(/*"Feature".equals(className) && */"isNonunique".equals(featureName) ||
```
ujhelyiz, line 185: Either remove the commented out section, or add a comment stating why is it necessary to keep it in the codebase.

org.omg.sysml/src/org/omg/sysml/util/traversal/facade/impl/ApiElementProcessingFacade.java
```diff
@@ -114,27 +113,71 @@ public Object process(Element element) {
...
+				UUID rootUUID = UUID.fromString(useAsRoot.getElementId());
+				var root = localState.getElement(rootUUID);
+				//reinsert element as root, this unsets its owners turning it into a root element
+				localState.addModelRoot(rootUUID, root);
+			}
+			
+			localState.addOutOfScopeReferencesAsProxies();
+			headRevision.setLocalState(localState);
+			APIModelDelta localChanges = headRevision.getLocalChanges();
+//			System.out.println(new org.omg.sysml.JSON().serialize(localChanges));
```
ujhelyiz, line 162: Remove this comment

org.omg.sysml/src/org/omg/sysml/util/repository/ProjectRepository.java
```diff
@@ -0,0 +1,397 @@
...
+			commit.setChange(changes);
+			commit.setPreviousCommit(new BranchHead().atId(previousCommitId));
+			return commitApi.postCommitByProject(projectId, commit, branchId);
+		} catch (ApiException e) {
+			throw new RemoteException("Error occured while trying to post commit", e);
+		}
+	}
+	
+	protected ProjectApi getProjectApi() {
+		return projectApi;	}
```
ujhelyiz, line 353: Add missing linebreak.

org.omg.sysml.xtext/src/org/omg/sysml/xtext/util/SysMLRepositoryLoadUtil.java
```diff
@@ -0,0 +1,170 @@
...
+				.addOption(targetOption)
+				.addOption(localLibrary);
+		
+		
+		CommandLineParser parser = new DefaultParser(false);
+		
+		CommandLine cli = parser.parse(options, args);
+		
+		return new SysMLRepositoryLoadUtil(
+					cli.hasOption(repositoryOption)? cli.getOptionValue(repositoryOption) : "http://localhost:9000",
```
ujhelyiz, line 96: This URL should be kept consistent with other base paths, e.g. the one in ApiElementProcessingFacade. The best would be if a single constant is used throughout the codebase.

ujhelyiz (changes requested):
I have quite a few a few minor comments that should be addressed before merging.

seidewitz:
@TheKorpos Please resolve any review comments from @ujhelyiz that have already been addressed.

seidewitz (changes requested):
- The help text for `publish` and `load` need to be updated to follow a style similar to that used for other commands.
- `publish` seems to still publish to a project name that includes a timestamp. This no longer makes sense. The command should have an option to provide a project name and use the given model element name as the default.
- I question whether we should include the `--branch` option as "experimental". Either we should include it and tests, with whatever limitations we currently have, or we should just not include it in this release.

org.omg.sysml/src/org/omg/sysml/util/repository/EMFModelRefresher.java
ujhelyiz: Handled in #616

org.omg.sysml/src/org/omg/sysml/util/repository/EMFModelRefresher.java
ujhelyiz: Resolved in #616

org.omg.sysml/src/org/omg/sysml/util/repository/EMFModelRefresher.java
ujhelyiz: Handled in #616

org.omg.sysml/src/org/omg/sysml/util/repository/EMFModelRefresher.java
ujhelyiz: Handled in #616

org.omg.sysml/src/org/omg/sysml/util/repository/EMFModelRefresher.java
ujhelyiz: Handled in #616

ujhelyiz (approved):

seidewitz:
I have merged `master` into this branch and updated the help messages for `%load`, `%projects` and `%publish`. I included documentation for the `--branch` options, but not labeled them "experimental".

This resolves the following issues from my review comment:

> * The help text for `publish` and `load` needs to be updated to follow a style similar to that used for other commands.

> * I question whether we should include the `--branch` option as "experimental". Either we should include it and tests, with whatever limitations we currently have, or we should just not include it in this release.

seidewitz:
I noticed some problems with the `%load` command.

1. Minor problem: When the branch given for the `--branch` option does not exist, the result is a null pointer exception, rather than an error message.
2. Major problem: The loading of models no longer seems to work correctly. I have verified that this problem existed (on my machine, at least) before the merge with `main`.

For example, consider the following model:
```
package 'Connections Example' {

    part def WheelHubAssembly;
    part def WheelAssembly;
    part def Tire;
    part def TireBead;
    part def Wheel;
    part def TireMountingRim;
    part def LugBoltMountingHole;
    part def Hub;
    part def LugBoltThreadableHole;
    part def LugBoltJoint;

    connection def PressureSeat {
        end : TireBead[1];
        end : TireMountingRim[1];
    }

    part wheelHubAssembly : WheelHubAssembly {

        part wheel : WheelAssembly[1] {
            part t : Tire[1] {
                part bead : TireBead[2];
            }
            part w: Wheel[1] {
                part rim : TireMountingRim[2];
                part mountingHoles : LugBoltMountingHole[5];
            }
            connection : PressureSeat connect t.bead to w.rim;
        }

        part lugBoltJoints : LugBoltJoint[0..5];
        part hub : Hub[1] {
            part h : LugBoltThreadableHole[5];
        }
        connect [0..1] lugBoltJoints
             to [1] mountingHoles references wheel.w.mountingHoles;
        connect [0..1] lugBoltJoints
             to [1] threadedHole ::> hub.h;
    }

}
```
As far as I can tell, publishing this to a new project in the repository still works correctly. However, when I load the project into a different notebook and attempt to visualize it using `viz --view=tree "Connection Example"`, the result is
image
This is clearly incomplete.

TheKorpos:
Hi @seidewitz

I've figured out

> 2\. Major problem: The loading of models no longer seems to work correctly. I have verified that this problem existed (on my machine, at least) before the merge with `main`.

This is a side effect of this [commit](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/633/commits/10fa86fd6b0db16cdf8b76a58d63b59a96162ad6) where we disabled derived properties by default in jupyter. We allowed `importedElement` to always go through, so the workaround that had been introduced to address the Membership UUID instability issue would still work. As it turns out `owner` is also required as the  The SysML-v2-API-Servies uses this derived field to calculate root elements. Because of the missing owner many elements got scattered around in random EMF resources because they were essentially marked as root.

The reason why you can still see PressureSeat in the diagram is because it is a ConnectionDefinition which is also a Relationship. Those are not returned as root elements by the [repository](https://github.com/Systems-Modeling/SysML-v2-API-Services/blob/9484b6f3753a27f3cbe58ced3d015ab7460e9bf6/app/dao/impl/jpa/JpaElementDao.java#L189). This means that you can't actually upload/download a specific Relationship form Jupyter, whithout nesting it in a package or something, as they can't be forced to act as root elements (in the repo) by setting `owner` to null.

seidewitz:
@TheKorpos 
I have confirmed both the load worked correctly before the latest commit when the `-d` option was used to publish the model, and that it works after the commit even without the `-d` option.

That leaves the following open comments:

> * Minor problem: When the branch given for the `--branch` option does not exist, the result is a null pointer exception, rather than an error message.

and

> * `publish` seems to still publish to a project name that includes a timestamp. This no longer makes sense. The command should have an option to provide a project name and use the given model element name as the default.

On the latter comment, I have updated the help description for `%publish` with the intended functionality:
```
Usage: %publish [-d] [--project=<PROJECT NAME>] [--branch=<BRANCH NAME>] <NAME>

Publish the model elements rooted in <NAME> to the repository. <NAME> must be fully qualified.
Use the -d flag to include derived properties.
If <PROJECT NAME> is given, it is used as the name of the project to create or update.
If <PROJECT NAME> is not given, the (simple) name of the model element is used.
    If no project exits with the given name, then a new project with that name is created.
    Otherwise, a new project is created with the given name.
If <BRANCH NAME> is given, then the model is written to this branch of the project.
If <BRANCH NAME> is not given, the default branch is used.
```

And, oops, I just realized that the help text is wrong! It should say:
```
    If no project exists with the given name, then a new project with that name is created.
    Otherwise, the existing project is updated with a new commit. 
```

seidewitz:
> Is lexicographic ordering good enough (Project11 coming before Project2) or should I use alphanumeric ordering?

Lexicographic ordering is fine.

seidewitz:
The command
```
%load --branch=test --id="e8d6c606-728a-43ae-84a3-6fe8853e460f"
```
results in an exception:
```
java.lang.IllegalArgumentException: Invalid UUID string: test
	at java.base/java.util.UUID.fromString1(UUID.java:280)
	at java.base/java.util.UUID.fromString(UUID.java:258)
	at org.omg.sysml.interactive.SysMLInteractive.loadById(SysMLInteractive.java:471)
	at org.omg.sysml.jupyter.kernel.magic.Load.load(Load.java:56)
	at jdk.internal.reflect.GeneratedMethodAccessor8.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at io.github.spencerpark.jupyter.kernel.magic.registry.Magics.invoke(Magics.java:89)
	at io.github.spencerpark.jupyter.kernel.magic.registry.Magics.access$000(Magics.java:8)
	at io.github.spencerpark.jupyter.kernel.magic.registry.Magics$LineReflectionMagicFunction.execute(Magics.java:149)
	at io.github.spencerpark.jupyter.kernel.magic.registry.Magics.applyLineMagic(Magics.java:26)
	at org.omg.sysml.jupyter.kernel.SysMLKernel.eval(SysMLKernel.java:94)
	at io.github.spencerpark.jupyter.kernel.BaseKernel.handleExecuteRequest(BaseKernel.java:334)
	at io.github.spencerpark.jupyter.channels.ShellChannel.lambda$bind$0(ShellChannel.java:64)
	at io.github.spencerpark.jupyter.channels.Loop.lambda$new$0(Loop.java:21)
	at io.github.spencerpark.jupyter.channels.Loop.run(Loop.java:78)
```
This is true whether or not `test` is a valid branch (in this case it was). The branch option works fine if a project name is given.

seidewitz:
If a name and ID are both given in a `%load` command, this results in the error message:
```
Name and id cannot be provided at the same time.
```
For consistency, this message should begin with "`Error:`".

TheKorpos:
> %load --branch=test --id="e8d6c606-728a-43ae-84a3-6fe8853e460f"

Currenlty the load command is set up in a way that referencing a project by UUID also requires the branch to be referenced by UUID.

The question is do we want to refeference branches by UUID? If the answer is yes, we probably should add `--branchId` too. Otherwise `--branch` should be the name of the branch.

seidewitz:
> > %load --branch=test --id="e8d6c606-728a-43ae-84a3-6fe8853e460f"
> 
> Currently the load command is set up in a way that referencing a project by UUID also requires the branch to be referenced by UUID.

I don't think this is intuitive. It certainly wasn't what I expected as a user. And it is not documented in the help

> The question is do we want to refeerence branches by UUID? If the answer is yes, we probably should add `--branchId` too. Otherwise `--branch` should be the name of the branch.

I hadn't really considered using UUIDs to identify branches. But I supposed it is consistent to have that option, too. It does need to be a separate option, though.

TheKorpos:
@seidewitz 
I've actually prepared an update, I haven't pushed yet, which makes `load `a little bit more flexible in the `SysMLInteractive` and allows all 4 cases.
I haven't made the `--branchId` option in Jupyter though (maybe `--branch-id` is a better name).

> I hadn't really considered using UUIDs to identify branches. But I supposed it is consistent to have that option, too. It does need to be a separate option, though.

By not having it as a separate option you mean testing for UUID and going with name if it's not a valid UUID?

> Note that this will also require an update to the help text. But perhaps we should wait and to that on ST6RI-823 (PR #649), so we don't impact the work of @ujhelyiz on that.

I agree. We should revisit the help text after @ujhelyiz's update is in.

TheKorpos:
@seidewitz
I pushed the update. `--branch` uses name now regardless of how the project was referenced. This is consistent with the current help text.

Allowing UUID for branches by either method can be easly done on top of this.

TheKorpos:
> Name and id cannot be provided at the same time.

I've addressed this too.

seidewitz:
> Allowing UUID for branches by either method can be easily done on top of this.

What I meant previously was that `--branch` should always take the branch name, as you have now implemented. Go ahead and add a new option (`--branchid` is fine) that takes the branch ID. As for project, the name and ID options should be mutually exclusive.

seidewitz:
When a project is loaded, the `%list` command fails on the top-level element of the project, though it seems to work on lower-level elements. E.g.,

image

If this is run in the Eclipse environment, there is a more useful exception trace:
```
1> %load AnotherTest
Selected branch main
Caching library UUIDs...
Downloading model...
Loaded Project AnotherTest (0e5b9fb8-4bb8-4296-a37a-648c8444ba2f)
3> %list AnotherTest
1    [main] ERROR xt.linking.lazy.LazyLinkingResource  - resolution of uriFragment '|0' failed.
java.lang.NullPointerException: Cannot invoke "org.eclipse.emf.ecore.EObject.eClass()" because "result" is null
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.getEObject(LazyLinkingResource.java:280)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.getEObject(LazyLinkingResource.java:236)
	at org.eclipse.emf.ecore.resource.impl.ResourceSetImpl.getEObject(ResourceSetImpl.java:223)
	at org.eclipse.emf.ecore.util.EcoreUtil.resolve(EcoreUtil.java:209)
	at org.eclipse.emf.ecore.util.EcoreUtil.resolve(EcoreUtil.java:269)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eResolveProxy(BasicEObjectImpl.java:1516)
	at org.omg.sysml.lang.sysml.impl.MembershipImportImpl.getImportedMembership(MembershipImportImpl.java:74)
	at org.omg.sysml.lang.sysml.impl.MembershipImportImpl.getTarget(MembershipImportImpl.java:121)
	at org.omg.sysml.lang.sysml.impl.RelationshipImpl.eGet(RelationshipImpl.java:325)
	at org.omg.sysml.lang.sysml.impl.ImportImpl.eGet(ImportImpl.java:446)
	at org.omg.sysml.lang.sysml.impl.MembershipImportImpl.eGet(MembershipImportImpl.java:149)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eGet(BasicEObjectImpl.java:1050)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eGet(BasicEObjectImpl.java:1042)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eGet(BasicEObjectImpl.java:1037)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.doResolveLazyCrossReference(LazyLinkingResource.java:175)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.resolveLazyCrossReference(LazyLinkingResource.java:162)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.resolveLazyCrossReferences(LazyLinkingResource.java:148)
	at org.eclipse.xtext.EcoreUtil2.resolveLazyCrossReferences(EcoreUtil2.java:505)
	at org.eclipse.xtext.validation.ResourceValidatorImpl.resolveProxies(ResourceValidatorImpl.java:162)
	at org.omg.kerml.xtext.validation.KerMLResourceValidator.resolveProxies(KerMLResourceValidator.java:42)
	at org.eclipse.xtext.validation.ResourceValidatorImpl.validate(ResourceValidatorImpl.java:75)
	at org.omg.sysml.interactive.SysMLInteractive.validate(SysMLInteractive.java:212)
	at org.omg.sysml.interactive.SysMLInteractive.process(SysMLInteractive.java:246)
	at org.omg.sysml.interactive.SysMLInteractive.listQuery(SysMLInteractive.java:337)
	at org.omg.sysml.interactive.SysMLInteractive.list(SysMLInteractive.java:351)
	at org.omg.sysml.interactive.SysMLInteractive.list(SysMLInteractive.java:357)
	at org.omg.sysml.interactive.SysMLInteractive.run(SysMLInteractive.java:765)
	at org.omg.sysml.interactive.SysMLInteractive.main(SysMLInteractive.java:842)
5    [main] ERROR xt.linking.lazy.LazyLinkingResource  - resolution of uriFragment '|0' failed.
java.lang.NullPointerException: Cannot invoke "org.eclipse.emf.ecore.EObject.eClass()" because "result" is null
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.getEObject(LazyLinkingResource.java:280)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.getEObject(LazyLinkingResource.java:236)
	at org.eclipse.emf.ecore.resource.impl.ResourceSetImpl.getEObject(ResourceSetImpl.java:223)
	at org.eclipse.emf.ecore.util.EcoreUtil.resolve(EcoreUtil.java:213)
	at org.eclipse.emf.ecore.util.EcoreUtil.resolve(EcoreUtil.java:269)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eResolveProxy(BasicEObjectImpl.java:1516)
	at org.omg.sysml.lang.sysml.impl.MembershipImportImpl.getImportedMembership(MembershipImportImpl.java:74)
	at org.omg.sysml.lang.sysml.impl.MembershipImportImpl.getTarget(MembershipImportImpl.java:121)
	at org.omg.sysml.lang.sysml.impl.RelationshipImpl.eGet(RelationshipImpl.java:325)
	at org.omg.sysml.lang.sysml.impl.ImportImpl.eGet(ImportImpl.java:446)
	at org.omg.sysml.lang.sysml.impl.MembershipImportImpl.eGet(MembershipImportImpl.java:149)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eGet(BasicEObjectImpl.java:1050)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eGet(BasicEObjectImpl.java:1042)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eGet(BasicEObjectImpl.java:1037)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.doResolveLazyCrossReference(LazyLinkingResource.java:175)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.resolveLazyCrossReference(LazyLinkingResource.java:162)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.resolveLazyCrossReferences(LazyLinkingResource.java:148)
	at org.eclipse.xtext.EcoreUtil2.resolveLazyCrossReferences(EcoreUtil2.java:505)
	at org.eclipse.xtext.validation.ResourceValidatorImpl.resolveProxies(ResourceValidatorImpl.java:162)
	at org.omg.kerml.xtext.validation.KerMLResourceValidator.resolveProxies(KerMLResourceValidator.java:42)
	at org.eclipse.xtext.validation.ResourceValidatorImpl.validate(ResourceValidatorImpl.java:75)
	at org.omg.sysml.interactive.SysMLInteractive.validate(SysMLInteractive.java:212)
	at org.omg.sysml.interactive.SysMLInteractive.process(SysMLInteractive.java:246)
	at org.omg.sysml.interactive.SysMLInteractive.listQuery(SysMLInteractive.java:337)
	at org.omg.sysml.interactive.SysMLInteractive.list(SysMLInteractive.java:351)
	at org.omg.sysml.interactive.SysMLInteractive.list(SysMLInteractive.java:357)
	at org.omg.sysml.interactive.SysMLInteractive.run(SysMLInteractive.java:765)
	at org.omg.sysml.interactive.SysMLInteractive.main(SysMLInteractive.java:842)
6    [main] ERROR xt.linking.lazy.LazyLinkingResource  - resolution of uriFragment '|0' failed.
java.lang.NullPointerException: Cannot invoke "org.eclipse.emf.ecore.EObject.eClass()" because "result" is null
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.getEObject(LazyLinkingResource.java:280)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.getEObject(LazyLinkingResource.java:236)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.doResolveLazyCrossReference(LazyLinkingResource.java:183)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.resolveLazyCrossReference(LazyLinkingResource.java:162)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.resolveLazyCrossReferences(LazyLinkingResource.java:148)
	at org.eclipse.xtext.EcoreUtil2.resolveLazyCrossReferences(EcoreUtil2.java:505)
	at org.eclipse.xtext.validation.ResourceValidatorImpl.resolveProxies(ResourceValidatorImpl.java:162)
	at org.omg.kerml.xtext.validation.KerMLResourceValidator.resolveProxies(KerMLResourceValidator.java:42)
	at org.eclipse.xtext.validation.ResourceValidatorImpl.validate(ResourceValidatorImpl.java:75)
	at org.omg.sysml.interactive.SysMLInteractive.validate(SysMLInteractive.java:212)
	at org.omg.sysml.interactive.SysMLInteractive.process(SysMLInteractive.java:246)
	at org.omg.sysml.interactive.SysMLInteractive.listQuery(SysMLInteractive.java:337)
	at org.omg.sysml.interactive.SysMLInteractive.list(SysMLInteractive.java:351)
	at org.omg.sysml.interactive.SysMLInteractive.list(SysMLInteractive.java:357)
	at org.omg.sysml.interactive.SysMLInteractive.run(SysMLInteractive.java:765)
	at org.omg.sysml.interactive.SysMLInteractive.main(SysMLInteractive.java:842)
java.lang.NullPointerException: Cannot invoke "org.omg.sysml.lang.sysml.Namespace.getImportedMembership()" because the return value of "org.omg.sysml.interactive.SysMLInteractiveResult.getRootElement()" is null
	at org.omg.sysml.interactive.SysMLInteractive.listQuery(SysMLInteractive.java:341)
	at org.omg.sysml.interactive.SysMLInteractive.list(SysMLInteractive.java:351)
	at org.omg.sysml.interactive.SysMLInteractive.list(SysMLInteractive.java:357)
	at org.omg.sysml.interactive.SysMLInteractive.run(SysMLInteractive.java:765)
	at org.omg.sysml.interactive.SysMLInteractive.main(SysMLInteractive.java:842)
4>
```

seidewitz:
The use of `next` here is a problem:
```
	private String load(RemoteBranch branch) {
		...
		delta.getProjectRoots().forEach((eObject, dto) -> {
			next(SYSMLX_EXTENSION);
			Resource xmiResource = getResource();
			if (eObject instanceof Namespace) {
				xmiResource.getContents().add(eObject);
			} else {
				Namespace root = SysMLFactory.eINSTANCE.createNamespace();
				NamespaceUtil.addOwnedMemberTo(root, (Element) eObject);
				xmiResource.getContents().add(root);
			}
			addResourceToIndex(xmiResource);
		});
		
		return "Loaded Project " + remoteProject.getProjectName() + " (" + remoteProject.getRemoteId().toString() + ")\n";
	}
```
The `next` method increments the `counter`, so it will get incremented for each root namespace loaded. But the intent is that the `counter` always reflect the corresponding Jupyter cell number, and the use of `next` hear will throw that out of sync, particularly if there are multiple root namespaces.

To fix this, I have replaced the lines
```
			next(SYSMLX_EXTENSION);
			Resource xmiResource = getResource();
```
with
```
			Resource resource = this.createResource(eObject.toString() + SYSML_EXTENSION);
			this.addInputResource(resource);
```

seidewitz:
The command `%list AnotherTest` is implemented by processing `private import AnotherTest;`. It is the import that is failing. Any import of the loaded top-level element fails:
```
11> %
package Test {
    public import AnotherTest;
}
%
2185596 [main] ERROR xt.linking.lazy.LazyLinkingResource  - resolution of uriFragment '|0' failed.
java.lang.NullPointerException: Cannot invoke "org.eclipse.emf.ecore.EObject.eClass()" because "result" is null
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.getEObject(LazyLinkingResource.java:280)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.getEObject(LazyLinkingResource.java:236)
	at org.eclipse.emf.ecore.resource.impl.ResourceSetImpl.getEObject(ResourceSetImpl.java:223)
	at org.eclipse.emf.ecore.util.EcoreUtil.resolve(EcoreUtil.java:209)
	at org.eclipse.emf.ecore.util.EcoreUtil.resolve(EcoreUtil.java:269)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eResolveProxy(BasicEObjectImpl.java:1516)
	at org.omg.sysml.lang.sysml.impl.MembershipImportImpl.getImportedMembership(MembershipImportImpl.java:74)
	at org.omg.sysml.lang.sysml.impl.MembershipImportImpl.getTarget(MembershipImportImpl.java:121)
	at org.omg.sysml.lang.sysml.impl.RelationshipImpl.eGet(RelationshipImpl.java:325)
	at org.omg.sysml.lang.sysml.impl.ImportImpl.eGet(ImportImpl.java:446)
	at org.omg.sysml.lang.sysml.impl.MembershipImportImpl.eGet(MembershipImportImpl.java:149)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eGet(BasicEObjectImpl.java:1050)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eGet(BasicEObjectImpl.java:1042)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eGet(BasicEObjectImpl.java:1037)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.doResolveLazyCrossReference(LazyLinkingResource.java:175)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.resolveLazyCrossReference(LazyLinkingResource.java:162)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.resolveLazyCrossReferences(LazyLinkingResource.java:148)
	at org.eclipse.xtext.EcoreUtil2.resolveLazyCrossReferences(EcoreUtil2.java:505)
	at org.eclipse.xtext.validation.ResourceValidatorImpl.resolveProxies(ResourceValidatorImpl.java:162)
	at org.omg.kerml.xtext.validation.KerMLResourceValidator.resolveProxies(KerMLResourceValidator.java:42)
	at org.eclipse.xtext.validation.ResourceValidatorImpl.validate(ResourceValidatorImpl.java:75)
	at org.omg.sysml.interactive.SysMLInteractive.validate(SysMLInteractive.java:212)
	at org.omg.sysml.interactive.SysMLInteractive.process(SysMLInteractive.java:246)
	at org.omg.sysml.interactive.SysMLInteractive.process(SysMLInteractive.java:239)
	at org.omg.sysml.interactive.SysMLInteractive.run(SysMLInteractive.java:732)
	at org.omg.sysml.interactive.SysMLInteractive.run(Unknown Source)
	at org.omg.sysml.interactive.SysMLInteractive.main(Unknown Source)
2185597 [main] ERROR xt.linking.lazy.LazyLinkingResource  - resolution of uriFragment '|0' failed.
java.lang.NullPointerException: Cannot invoke "org.eclipse.emf.ecore.EObject.eClass()" because "result" is null
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.getEObject(LazyLinkingResource.java:280)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.getEObject(LazyLinkingResource.java:236)
	at org.eclipse.emf.ecore.resource.impl.ResourceSetImpl.getEObject(ResourceSetImpl.java:223)
	at org.eclipse.emf.ecore.util.EcoreUtil.resolve(EcoreUtil.java:213)
	at org.eclipse.emf.ecore.util.EcoreUtil.resolve(EcoreUtil.java:269)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eResolveProxy(BasicEObjectImpl.java:1516)
	at org.omg.sysml.lang.sysml.impl.MembershipImportImpl.getImportedMembership(MembershipImportImpl.java:74)
	at org.omg.sysml.lang.sysml.impl.MembershipImportImpl.getTarget(MembershipImportImpl.java:121)
	at org.omg.sysml.lang.sysml.impl.RelationshipImpl.eGet(RelationshipImpl.java:325)
	at org.omg.sysml.lang.sysml.impl.ImportImpl.eGet(ImportImpl.java:446)
	at org.omg.sysml.lang.sysml.impl.MembershipImportImpl.eGet(MembershipImportImpl.java:149)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eGet(BasicEObjectImpl.java:1050)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eGet(BasicEObjectImpl.java:1042)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eGet(BasicEObjectImpl.java:1037)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.doResolveLazyCrossReference(LazyLinkingResource.java:175)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.resolveLazyCrossReference(LazyLinkingResource.java:162)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.resolveLazyCrossReferences(LazyLinkingResource.java:148)
	at org.eclipse.xtext.EcoreUtil2.resolveLazyCrossReferences(EcoreUtil2.java:505)
	at org.eclipse.xtext.validation.ResourceValidatorImpl.resolveProxies(ResourceValidatorImpl.java:162)
	at org.omg.kerml.xtext.validation.KerMLResourceValidator.resolveProxies(KerMLResourceValidator.java:42)
	at org.eclipse.xtext.validation.ResourceValidatorImpl.validate(ResourceValidatorImpl.java:75)
	at org.omg.sysml.interactive.SysMLInteractive.validate(SysMLInteractive.java:212)
	at org.omg.sysml.interactive.SysMLInteractive.process(SysMLInteractive.java:246)
	at org.omg.sysml.interactive.SysMLInteractive.process(SysMLInteractive.java:239)
	at org.omg.sysml.interactive.SysMLInteractive.run(SysMLInteractive.java:732)
	at org.omg.sysml.interactive.SysMLInteractive.run(Unknown Source)
	at org.omg.sysml.interactive.SysMLInteractive.main(Unknown Source)
2185598 [main] ERROR xt.linking.lazy.LazyLinkingResource  - resolution of uriFragment '|0' failed.
java.lang.NullPointerException: Cannot invoke "org.eclipse.emf.ecore.EObject.eClass()" because "result" is null
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.getEObject(LazyLinkingResource.java:280)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.getEObject(LazyLinkingResource.java:236)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.doResolveLazyCrossReference(LazyLinkingResource.java:183)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.resolveLazyCrossReference(LazyLinkingResource.java:162)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.resolveLazyCrossReferences(LazyLinkingResource.java:148)
	at org.eclipse.xtext.EcoreUtil2.resolveLazyCrossReferences(EcoreUtil2.java:505)
	at org.eclipse.xtext.validation.ResourceValidatorImpl.resolveProxies(ResourceValidatorImpl.java:162)
	at org.omg.kerml.xtext.validation.KerMLResourceValidator.resolveProxies(KerMLResourceValidator.java:42)
	at org.eclipse.xtext.validation.ResourceValidatorImpl.validate(ResourceValidatorImpl.java:75)
	at org.omg.sysml.interactive.SysMLInteractive.validate(SysMLInteractive.java:212)
	at org.omg.sysml.interactive.SysMLInteractive.process(SysMLInteractive.java:246)
	at org.omg.sysml.interactive.SysMLInteractive.process(SysMLInteractive.java:239)
	at org.omg.sysml.interactive.SysMLInteractive.run(SysMLInteractive.java:732)
	at org.omg.sysml.interactive.SysMLInteractive.run(Unknown Source)
	at org.omg.sysml.interactive.SysMLInteractive.main(Unknown Source)
org.eclipse.emf.common.util.WrappedException: java.lang.NullPointerException: Cannot invoke "org.eclipse.emf.ecore.EObject.eClass()" because "result" is null
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.getEObject(LazyLinkingResource.java:244)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.doResolveLazyCrossReference(LazyLinkingResource.java:183)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.resolveLazyCrossReference(LazyLinkingResource.java:162)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.resolveLazyCrossReferences(LazyLinkingResource.java:148)
	at org.eclipse.xtext.EcoreUtil2.resolveLazyCrossReferences(EcoreUtil2.java:505)
	at org.eclipse.xtext.validation.ResourceValidatorImpl.resolveProxies(ResourceValidatorImpl.java:162)
	at org.omg.kerml.xtext.validation.KerMLResourceValidator.resolveProxies(KerMLResourceValidator.java:42)
	at org.eclipse.xtext.validation.ResourceValidatorImpl.validate(ResourceValidatorImpl.java:75)
	at org.omg.sysml.interactive.SysMLInteractive.validate(SysMLInteractive.java:212)
	at org.omg.sysml.interactive.SysMLInteractive.process(SysMLInteractive.java:246)
	at org.omg.sysml.interactive.SysMLInteractive.process(SysMLInteractive.java:239)
	at org.omg.sysml.interactive.SysMLInteractive.run(SysMLInteractive.java:732)
	at org.omg.sysml.interactive.SysMLInteractive.run(Unknown Source)
	at org.omg.sysml.interactive.SysMLInteractive.main(Unknown Source)
Caused by: java.lang.NullPointerException: Cannot invoke "org.eclipse.emf.ecore.EObject.eClass()" because "result" is null
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.getEObject(LazyLinkingResource.java:280)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.getEObject(LazyLinkingResource.java:236)
	... 13 more
12> 
```

TheKorpos:
@seidewitz 
Thank you for the updates.

> When a project is loaded, the `%list` command fails on the top-level element of the project, though it seems to work on lower-level elements. E.g.,

The problem here is that when you publish a namespace from Jupyter the root `Namespace` and the `OwningMembership` is not uploaded as they're out of scope. So the membership import can't reference the `OwningMembership`.

These lines of code, that you've since adjusted, meant to account for this:

> 		delta.getProjectRoots().forEach((eObject, dto) -> {
> 			next(SYSMLX_EXTENSION);
> 			Resource xmiResource = getResource();
> 			if (eObject instanceof Namespace) {
> 				xmiResource.getContents().add(eObject);
> 			} else {
> 				Namespace root = SysMLFactory.eINSTANCE.createNamespace();
> 				NamespaceUtil.addOwnedMemberTo(root, (Element) eObject);
> 				xmiResource.getContents().add(root);
> 			}
> 			addResourceToIndex(xmiResource);
> 		});
> 		

This should wrap non-namepsace roots in a `Namespace` with an `OwningMembership`, but I made a mistake. Instead of `eObject instanceof Namespace` it should be `eObject.eClass() == SysMLPackage.eINSTANCE.getNamespace()`.

This is only done in the Jupyter %load command. However Eclipse and the CLI utility should also wrap project roots under a rootnamespace.

Alternatively, we could modify the %publish command, so that it uploads the whole ownership hierarchy leading to the selected `Namespace`. That would eliminate the need of ceating this wrapper `Namespace`. However the API doesn't require the model to be rooted in Namepsaces (at least I couldn't find such constraint) so I think it's the loading that should account for these kind of models.

seidewitz (approved):

PR #621, opened 2025-02-15
ST6RI-606 Code generator tooling for ISQ, SI, US Customary Units libraries

hpdekoning:
This PR provides a new `quantities_lib_generator` tool under `tool-support`.

This tool is able to reproduce the SysML release 2024-12 ISQ library packages with a few differences that are all due to minor naming mistakes in the published release 2024-12 libraries and also three editorial issues.

In particular the naming of coordinate frames in release 2024-12 was inconsistent:
* Most vector quantities and related coordinate frames were using e.g. `Cartesian3dXxxVector` instead of `CartesianXxx3dVector`, The latter is the correct naming convention that naturally fits with the quantity taxonomy. More specialized concepts prepend qualifiers to the name, e.g., a specialization chain from `VectorQuantityValue` to `'3dVectorQuantityValue'` to `Position3dVector` to `CartesianPosition3dVector` or `CylindricalPosition3dVector`.
* Also `VectorQuantityValue` was used in many places instead of the correct `'3dVectorQuantityValue'`. Same for `VectorMeasurementReference` instead of `'3dCoordinateFrame'`.

Also, the following editorial issues were found and fixed:
1. In `ISQCondensedMatter.sysml` Item 12-19 _Hall coefficient_: replaced erroneous unit m^3/C*m^3*s^-1*A^-1 with correct m^3/C, m^3*s^-1*A^-1 (in the Excel workbook `data/iso-iec-80000/iso_iec_80000_quantities_and_units.xlsx`). As a consequence, the declaration of `attribute def HallCoefficientUnit` was changed.
2. ISQSpaceTime.sysml contained redundant `private import ScalarValues::String;` . It was not generated, which is in line with all other ISQ libraries.
3. The ISO/IEC 80000 references were checked and where needed updated to the latest correct titles, editions and URIs. This results in minor editorial corrections in the `doc` content of the libraries.

A detailed review of all differences between the ISQ library packages in folder `libs_baseline_release-2024-12` and the auto-generated library packages in folder `libs_generated` has been performed. No other differences than those documented above have been found.

There are currently no changes to the generation of `SI` and `USCustomaryUnits`.

PR #634, opened 2025-03-13
ST6RI-606-1 Code generator tooling for ISQ, SI, US Customary Units libraries

hpdekoning:
This PR is an amendment on [PR#621](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/621) that provides a new `quantities_lib_generator` tool under `tool-support`.

This tool is able to reproduce the SysML release 2024-12 ISQ library packages with a few differences that are all due to minor naming mistakes in the published release 2024-12 libraries and also three editorial issues.

In particular the naming of coordinate frames in release 2024-12 was inconsistent:
* Most vector quantities and related coordinate frames were using e.g. `Cartesian3dXxxVector` instead of `CartesianXxx3dVector`, The latter is the correct naming convention that naturally fits with the quantity taxonomy. More specialized concepts prepend qualifiers to the name, e.g., a specialization chain from `VectorQuantityValue` to `'3dVectorQuantityValue'` to `Position3dVector` to `CartesianPosition3dVector` or `CylindricalPosition3dVector`.
* Also `VectorQuantityValue` was used in many places instead of the correct `'3dVectorQuantityValue'`. Same for `VectorMeasurementReference` instead of `'3dCoordinateFrame'`.

Also, the following editorial issues were found and fixed:
1. In `ISQCondensedMatter.sysml` Item 12-19 _Hall coefficient_: replaced erroneous unit m^3/C*m^3*s^-1*A^-1 with correct m^3/C, m^3*s^-1*A^-1 (in the Excel workbook `data/iso-iec-80000/iso_iec_80000_quantities_and_units.xlsx`). As a consequence, the declaration of `attribute def HallCoefficientUnit` was changed.
2. ISQSpaceTime.sysml contained redundant `private import ScalarValues::String;` . It was not generated, which is in line with all other ISQ libraries.
3. The ISO/IEC 80000 references were checked and where needed updated to the latest correct titles, editions and URIs. This results in minor editorial corrections in the `doc` content of the libraries.

A detailed review of all differences between the ISQ library packages in folder `libs_baseline_release-2024-12` and the auto-generated library packages in folder `libs_generated` has been performed. No other differences than those documented above have been found.

There are currently no changes to the generation of `SI` and `USCustomaryUnits`.

### DIFF
--- a/tool-support/quantities_lib_generator/data_templates.py
+++ b/tool-support/quantities_lib_generator/data_templates.py
@@ -149,7 +149,7 @@ standard library package ISQBase {{
         attribute :>> isOrthogonal = true;
     }
 
-    readonly attribute universalCartesianSpatial3dCoordinateFrame : CartesianSpatial3dCoordinateFrame[1] {
+    attribute universalCartesianSpatial3dCoordinateFrame : CartesianSpatial3dCoordinateFrame[1] {
         doc
         /*
          * A singleton CartesianSpatial3dCoordinateFrame that can be used as a default universal Cartesian 3D coordinate frame.
--- a/tool-support/quantities_lib_generator/libs_generated/ISQSpaceTime.sysml
+++ b/tool-support/quantities_lib_generator/libs_generated/ISQSpaceTime.sysml
@@ -2,7 +2,7 @@ standard library package ISQSpaceTime {
     doc
     /*
      * International System of Quantities and Units
-     * Generated on 2025-03-13T15:00:05Z from standard ISO-80000-3:2019 "Space and Time"
+     * Generated on 2025-03-13T17:19:25Z from standard ISO-80000-3:2019 "Space and Time"
      * see also https://www.iso.org/standard/64974.html
      * 
      * Note 1: In documentation comments, AsciiMath notation (see http://asciimath.org/) is used for mathematical concepts,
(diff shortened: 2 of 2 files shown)


hpdekoning:
@seidewitz 
This PR fixes the remaining problems with [PR#621](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/621):
* Two minor corrections to the iso_iec_80000_quantities_and_units.xlsx Excel workbook
* Fix to the `quantities_lib_generator` so that the generated ISQ standard libraries correctly validate and parse in the latest Eclipse reference implementation. The problems were with correct naming of vector quantities, coordinate frames and vector quantity aliases.

seidewitz:
For SysML, `readonly` has been replaced by `constant`, but this can no longer be used on package-level usages. So the one case of `readonly` simply needs to be removed.

hpdekoning:
> For SysML, `readonly` has been replaced by `constant`, but this can no longer be used on package-level usages. So the one case of `readonly` simply needs to be removed.

Just committed the removal in [this commit](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/634/commits/afe2d8ec4c446112c2884901db674f6eebffa3be).

PR #636, opened 2025-03-15
ST6RI-829 Standard model library element IDs (KERML_-36, SYSML2_-107)

seidewitz:
This PR implements resolutions to the following issues, as approved in KerML FTF2 Ballot 5 and SysML v2 FTF 2 Ballot 8:

- [KERML_-36](http://issues.omg.org/issues/KERML_-36) KerML Libraries' elements shall have an elementId defined
- [SYSML2_-107](http://issues.omg.org/issues/SYSML2_-107) SysML Libraries' elements shall have an elementId defined

**Paths**

The resolution to KERML_-36 introduces the concept of a "path" for an element that is unique within the composition structure under a specific root namespace. The element path is constructed as follows:

1. If the element has a non-null qualified name, that is used as the path. This means that the UUIDs generated for elements with non-null qualified names will remain unchanged from previous versions of the normative library models.
2. If the element has a null qualified name and a non-null owning relationship, then its path is constructed from the path of its owning relationship by appending the character "`/`", followed by its position in the list of owned related elements of it owning relationship.
3. If the element is a relationship with a non-null owning related element (except for an owning membership as below), then its path is instead constructed from the path of its owning related element by appending the character "`/`", followed by its position in the list of owned relationships of it owning related element.
4. If the element is an owning membership whose owned member element has a non-null qualified name, then its path is constructed from the qualified name of its owned member element by appending the string "`/owningMembership`". Since owning memberships are one-to-one with their owned member elements, this allows owning memberships of named elements in library models to remain stable even if there is a reordering of member elements within the owning namespace.

The following shows some examples of element paths:
```
// Path of package: TopLevel
// Path of owning membership: TopLevel/owningMembership
package TopLevel {

    // Path of classifier: TopLevel::A
    // Path of owning membership: TopLevel::A/owningMembership
    classifier A;

    // Path of classifier: TopLevel::B
    // Path of owning membership: TopLevel::B/owningMembership
    // Path of owned subclassification: TopLevel::B/1
    classifier B specializes A;
        // Path of owning membership: TopLevel::B/2
        // Path of feature:  TopLevel::B/2/1
        feature;
    }

    // Path of owning membership: TopLevel/3
    // Path of classifier: TopLevel/3/1
    classifier {
        // Path of owning membership: TopLevel/3/1/1
        // Path of feature: TopLevel/3/1/1/1
        feature f;
    }
}
```

**Unique UUIDs**

The resolutions to both KERML_-36 and SYSML2_-107 change the algorithm for generating UUIDs, for KerML and SysML, respectively, to use element paths instead of qualified names. This allows a distinct UUID to be generated for every element in a standard library model, not just the ones with non-null qualified name. For elements with non-null qualified names, though, the path is just the qualified name, so the generated UUID is the same as previously. 

The resolution to KERML_-36 also revises the derivation for `qualifiedName` so that, if there are multiple owned members in a namespace with the same name, all but the first of these have a null `qualifiedName`. While the `validateNamespaceDistinguishability` constraint makes it invalid for a namespace to have duplicate names, the revised derivation for `qualifiedName` ensures that no two elements have the same qualified name, even for models that violate the constraint. This eliminates the possibility of different elements with the same qualified name getting the same generated UUID. 

Technically, elements are only "standard library elements" if they are a top-level standard library package or an element in the composition structure of such a package. However, for each library model file, this leaves out two elements that are outside the top-level library package: the root namespace and the owning membership from the root namespace to the top-level library package. The implementation in this PR gives stable UUIDs to these elements, too, by generating name-based UUIDs for them using the libary package UUID as the namespace UUID and a name constructed from the library package name appended with one of the following:

- "`/owner`" for the root namespace
- "`/owningMembership`" for the owning membership

In this way, every element in a library model file has a stable UUID every time the textual notation in the file is parsed without change.

### DIFF
--- a/org.omg.sysml.interactive.tests/.launch/Derived Property and Operation Test.launch
+++ b/org.omg.sysml.interactive.tests/.launch/Derived Property and Operation Test.launch
@@ -1,8 +1,9 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <launchConfiguration type="org.eclipse.jdt.junit.launchconfig">
     <stringAttribute key="bad_container_name" value="/org.omg.sysml.interactive.tests/.lau"/>
+    <booleanAttribute key="org.eclipse.debug.core.ATTR_FORCE_SYSTEM_CONSOLE_ENCODING" value="false"/>
     <listAttribute key="org.eclipse.debug.core.MAPPED_RESOURCE_PATHS">
-        <listEntry value="/org.omg.sysml.interactive.tests/src/org/omg/sysml/interactive/tests/DerivedPropertyTest.java"/>
+        <listEntry value="/org.omg.sysml.interactive.tests/src/org/omg/sysml/interactive/tests/DerivedPropertyAndOperationTest.java"/>
     </listAttribute>
     <listAttribute key="org.eclipse.debug.core.MAPPED_RESOURCE_TYPES">
         <listEntry value="1"/>
@@ -12,10 +13,11 @@
     <stringAttribute key="org.eclipse.jdt.junit.TESTNAME" value=""/>
     <stringAttribute key="org.eclipse.jdt.junit.TEST_KIND" value="org.eclipse.jdt.junit.loader.junit4"/>
     <booleanAttribute key="org.eclipse.jdt.launching.ATTR_ATTR_USE_ARGFILE" value="false"/>
+    <booleanAttribute key="org.eclipse.jdt.launching.ATTR_SHOW_CODEDETAILS_IN_EXCEPTION_MESSAGES" value="true"/>
     <booleanAttribute key="org.eclipse.jdt.launching.ATTR_USE_CLASSPATH_ONLY_JAR" value="false"/>
     <booleanAttribute key="org.eclipse.jdt.launching.ATTR_USE_START_ON_FIRST_THREAD" value="true"/>
     <stringAttribute key="org.eclipse.jdt.launching.CLASSPATH_PROVIDER" value="org.eclipse.m2e.launchconfig.classpathProvider"/>
-    <stringAttribute key="org.eclipse.jdt.launching.MAIN_TYPE" value="org.omg.sysml.interactive.tests.DerivedPropertyTest"/>
+    <stringAttribute key="org.eclipse.jdt.launching.MAIN_TYPE" value="org.omg.sysml.interactive.tests.DerivedPropertyAndOperationTest"/>
     <stringAttribute key="org.eclipse.jdt.launching.PROJECT_ATTR" value="org.omg.sysml.interactive.tests"/>
     <stringAttribute key="org.eclipse.jdt.launching.SOURCE_PATH_PROVIDER" value="org.eclipse.m2e.launchconfig.sourcepathProvider"/>
     <stringAttribute key="org.eclipse.jdt.launching.VM_ARGUMENTS" value="-ea -DlibraryPath=${workspace_loc:/SysML-v2-Pilot-Implementation/sysml.library}"/>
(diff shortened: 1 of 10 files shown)


himi:
I have not yet checked the results of `path()` but I'm guessing that even if a relationship has a qualified name, `path()` does not return it.  Also I'm afraid if a qualified name contains `/`, we cannot properly split  the result of `path()`.   I'm trying to check these.

seidewitz:
> I have not yet checked the results of `path()` but I'm guessing that even if a relationship has a qualified name, `path()` does not return it.

A relationship can only have a qualified name if it has an `owningNamespace`. It can only have an `owningNamespace` if it is owned by a namespace via an `OwningMembership` (e.g., in the case of a dependency that is a member of a package), which would then be its `owningRelationship`. If a relationship has an `owningRelationship`, then its path is computed as if it was a regular element, not a relationship, so, if it has a `qualifiedName`, that will be its path.

> Also I'm afraid if a qualified name contains /, we cannot properly split the result of path(). I'm trying to check these.

"`/`" is not a legal identifier character, so it would have to appear within single quotes in a qualified name segment, because each of the segments of a qualified name have to be properly escaped names. So, if an element has, e.g., the name "`a/b`", then this will appear in its `qualifiedName` (if it has one) as "`'a/b'`". Any algorithm to to split up paths simply has to take into account that "`/`" characters inside single quotes are not path segment separators.

himi:
Ed, thank you.  But in my understanding, in a parsed model, `qualifiedName` should have a name without single quotes.  For example, we get `a/b` if we write `'a/b'` in textual notation.

seidewitz:
> Ed, thank you. But in my understanding, in a parsed model, `qualifiedName` should have a name without single quotes. For example, we get `a/b` if we write `'a/b'` in textual notation.

That is not correct. By design, the derivation for `qualifiedName` is such that the value is directly parsable according to KerML/SysML syntax. Otherwise, the result would be ambiguous if, e.g., "`::`" was included in a segment name (which is legal).

The OCL for `deriveElementQualifiedName`, as modified by the resolution to KERML_-36 is
```
qualifiedName =
    if owningNamespace = null then null
    else if name <> null and 
        owningNamespace.ownedMember->
        select(m | m.name = name).indexOf(self) <> 1 then null
    else if owningNamespace.owner = null then escapedName()
    else if owningNamespace.qualifiedName = null or 
            escapedName() = null then null
    else owningNamespace.qualifiedName + '::' + escapedName()
    endif endif endif endif
```
Note the use of `escapedName()` (even in the case of a single segment).

himi:
Thank you.  I did not know that about `qualifiedName` , unlike `name` feature.

himi:
Sorry, I should check it by myself but still I'm struggling to run this version.  So let me ask you about aliases, where `Membership` can have `qualifiedName` in my understanding.  I'm not sure that `qualifiedName` is used to make `path()` in this case as well.

seidewitz:
> So let me ask you about aliases, where `Membership` can have `qualifiedName` in my understanding. I'm not sure that `qualifiedName` is used to make `path()` in this case as well.

An alias is represented in the abstract syntax as a `Membership` that is not an `OwningMembership`, whose `memberName` acts as an alias name for its `memberElement`. Note that the `memberName` is _not_ the `name` of the `Membership`, which is not relevant to name resolution. A `Membership` is always an `ownedRelationship` of its `Namespace`, _not_ an `ownedMember`, so it's `qualifiedName` property is always null. The `path` for a `Membership` that is not an `OwningMembership` will always be based on its position in the list of `ownedRelationships` of its `Namespace`.

Just to further clarify, only `Elements` that are `ownedMembers` of some `Namespace` can have qualified names. While aliases and imports can result in multiple qualified names resolving to the same `Element`, the `qualifiedName` property of the `Element` is always its "ownership" qualified name, that is, the qualified name constructed from the names of the hierarchy of containing `Namespaces` in which the `Element` is nested, as long as all those `Namespaces` (and the `Element`) have `names`.

himi:
Thank you for your answer.  I misunderstood that point as well.  Anyway, I could check path() and ID in this build and I think it works as you explained:
image

himi:
Also I briefly checked it by a simple example:
image

himi (approved):

TheKorpos (approved):
@seidewitz
I've generated some data sets from the library that lists the Qn, path and UUID for all elements (grouped by UUID). The generated data remains the same across runs as expected.

PR #638, opened 2025-03-25
ST6RI-818 Constructor expressions (KERML_-132/222, SYSML2_-766)

seidewitz:
This PR implements resolutions to the following issues, as approved in KerML FTF2 Ballots #5 and #6 and SysML v2 FTF 2 Ballot #8:

- [KERML_-132](http://issues.omg.org/issues/KERML_-132) Constructor invocation expressions are semantically inconsistent
- [KERML_-222](http://issues.omg.org/issues/KERML_-222) validateParameterMembershipOwningType will fail for ConstructorExpressions
- [SYSML2_-766](http://issues.omg.org/issues/SYSML2_-766) Constructor expressions in SysML

Previously, an invocation expression where the invoked type was _not_ a function acted as a _constructor_ for an instance of the invoked type. The resulting usage was both typed by the invoked type and a subset of `Performances::evaluations`. But, if the invoked type was a data type or structure, then this was a semantic inconsistency, because `Performance` (a supertype of `evaluations`) is disjoint with `DataValue` and `Object`. The resolution to KERML_-132 fixes this by introducing a new kind of _constructor expression_ and requiring that the invoked type of an invocation expression be a behavior or function. 

**Abstract Syntax**

1. `InstantiationExpression`
   * Added as a abstract superclass of `InvocationExpression` and `ConstructorExpression`.
   * Defines a derived property `instantiatedType`, which is the `Type` invoked/instantiated by the expression.
   * In order to allow more consistent parsing, the `instantiatedType` of any kind of `InstantiationExpression`, other than an `OperatorExpression` or `TriggerInvocationExpression` (in SysML), is parsed as a member of the expression. (There are special derivation rules for the `instantiatedType` of an `OperatorExpression` or `TriggerInvocationExpression`.)
   * An `InstantiationExpression` must own its `result` parameter.
2. `InvocationExpression`
   * An `InvocationExpression` is required to specialize the `instantiatedType`, as before.
   * The `instantiatedType` of an `InvocationExpression` must be a `Behavior`, normally a `Function`.
   * If the `instantiatedType` is a `Behavior` that is not a `Function`, then the `InvocationExpression` is still typed by the `Behavior`, with its `result` bound to the expression itself. This allows non-`Function` `Behaviors` to still be essentially "constructed" using an `InvocationExpression`, as before. (This does not cause semantic problems, because it is acceptable to specialize a _`Performance`_ of the `Behavior` to an _`Evaluation`_ that returns itself.)
   * The `arguments` of an `InvocationExpression` are parsed as `FeatureValues` of `ownedFeatures` of the expression itself, as before.
3. `ConstructorExpression`
   * Added to replace the use of `InvocationExpressions` with constructor semantics. 
   * The result of a `ConstructorExpression` is required to specialize the `instantiatedType`.
   * The `arguments` of a `ConstructorExpression` are parsed as `FeatureValues` of `ownedFeatures` of the `result` of the `ConstructorExpression`. (In this way, a value of the `instantiatedType` is "constructed" as the `result` of the `ConstructorExpression`, while the expression itself can still be typed as an `Evaluation`.)
4. `FeatureReferenceExpression`
   * `FeatureReferenceExpression` is also updated with a new validation constraint requiring that it own its `result` parameter.

**Note.** The Pilot Implementation previously inserted an owned `result` parameter for _all_ kinds of expressions. Now this is only done for `InstantiationExpressions` and `FeatureReferenceExpressions`.

**Concrete Syntax**

The concrete syntax for a constructor expression is similar to that of an invocation expression, but preceded by the keyword **`new`**.
```
class Member {
    feature firstName : String;
    feature lastName : String;
    feature memberNumber : Integer;
    feature sponsor : Member[0..1];
}
feature thisMember = new Member("Jane", "Doe", 1234, null);
feature nextMember = new Member(
    firstName = "John", lastName = "Doe", sponsor = thisMember,
    memberNumber = thisMember.memberNumber + 1);
```

**Model Libraries**

_Kernel Semantic Library_
1. `Performances` – Add _`constructorEvaluations`_ as the base feature of all `ConstructorExpressions`. _`constructorEvaluations`_ is a specialization of _`evaluations`_ that restricts the multiplicity of its `result `parameter to `1..1`, requiring a _`constructorEvaluation`_ to result in a single value.
2. `Triggers` – Use the new syntax to construct the result values for `TriggerWhen` and `TriggerAt`.

_SysML Domain Libraries_
1. Analysis Domain Library, `SampleFunctions – Use the new syntax to construct the result value for the calculation definition `Sample`.
2. Metadata Domain Library, `RiskMetadata` – Use the new syntax to construct the enumerated values of the enumeration definition `RiskLevel`.
3. Quantities and Units Library, `MeasurementReferences` – Use the new syntax to construct the value of the attribute usage `one`.


himi:
I am reviewing this.  I could not find any issues so far.  But I found ConstructorExpression can be used for behavior as well as InvocationExpression.

For example, 
```
behavior B1 {
   feature f0;
}
feature f1 = B1();
feature f2 = new B1();
```

Are these expressions having the same semantics?

P.S.
I confirmed InvocationExpression could not be used for instantiating non-behavior Classes.

seidewitz:
> For example,
> 
> ```
> behavior B1 {
>    feature f0;
> }
> feature f1 = B1();
> feature f2 = new B1();
> ```
> 
> Are these expressions having the same semantics?

Yes, this is mentioned in the PR description:

> 2. `InvocationExpression`
>    ...
>    * If the `instantiatedType` is a `Behavior` that is not a `Function`, then the `InvocationExpression` is still typed by the `Behavior`, with its `result` bound to the expression itself. This allows non-`Function` `Behaviors` to still be essentially "constructed" using an `InvocationExpression`, as before. (This does not cause semantic problems, because it is acceptable to specialize a _`Performance`_ of the `Behavior` to an _`Evaluation`_ that returns itself.)
>
>    ...
>

sysml.library/Kernel Libraries/Kernel Function Library/BaseFunctions.kerml
```diff
@@ -45,25 +45,25 @@ standard library package BaseFunctions {
...
-		abstract feature 'type': Anything;
+		in 'type': Anything;
 		return : Boolean[1];
 	}
 	
 	abstract function 'hastype'{ 
 		in seq: Anything[0..*];
-		abstract feature 'type': Anything;
+		in 'type': Anything;
 		return : Boolean;
```
hpdekoning, line 55: For consistency with other functions in package BaseFunctions type should be `Boolean[1]`, i.e. add explicit multiplicity.
Aternative remove `[1]` (which is the default multiplicity) from the other three functions.

sysml/src/training/25. Transitions/Local Clock Example.sysml
```diff
@@ -21,7 +21,7 @@ package 'Local Clock Example' {
 			state waiting;
 			accept request : Request via requestPort
 				then responding;
-			accept at Time::Iso8601DateTime(today + "11:59:00")
+			accept at new Time::Iso8601DateTime(today + "11:59:00")
```
hpdekoning, line 24: Not related to this PR, but should `today` be given a value like "2024-10-25" and then change the string concatenation expression to `today + "T" + "11:59:00"` to create an example that is complete?

hpdekoning:
👍 Good for merge. I did not find any overlooked constructor expresssions that need a `new` keyword.

sysml/src/validation/05-State-based Behavior/5-State-based Behavior-1.sysml
```diff
@@ -82,7 +82,7 @@ package '5-State-based Behavior-1' {
 					first off
 					accept 'Vehicle Start Signal' 
 					if vehicle1_c1.'brake pedal depressed'
-					do send 'Start Signal'() to vehicle1_c1.vehicleController
+					do send new 'Start Signal'() to vehicle1_c1.vehicleController
```
hpdekoning, line 85: In line with the change, in the comment block below
`The notation "'Start Signal'()" constructs` should be changed to
`The notation "new 'Start Signal'()" constructs`

sysml/src/validation/05-State-based Behavior/5-State-based Behavior-1a.sysml
```diff
@@ -83,7 +83,7 @@ package '5-State-based Behavior-1a' {
 					first off
 					accept 'Vehicle Start Signal' 
 					if vehicle1_c1.'brake pedal depressed'
-					do send 'Start Signal'() to vehicle1_c1.vehicleController
+					do send new 'Start Signal'() to vehicle1_c1.vehicleController
```
hpdekoning, line 86: In line with the change, in the comment block below
`The notation "'Start Signal'()" constructs` should be changed to
`The notation "new 'Start Signal'()" constructs`

In the comment there is also a typo: `parhentheses` -> `parentheses`.

sysml/src/training/25. Transitions/Local Clock Example.sysml
seidewitz: As noted, this is not really relevant to this PR. It could be considered when the training material is updated. But it is not clear that it is important to the example. Plus, `today` probably shouldn't have a fixed value.

sysml/src/validation/05-State-based Behavior/5-State-based Behavior-1.sysml
seidewitz: Fixed.

sysml/src/validation/05-State-based Behavior/5-State-based Behavior-1a.sysml
seidewitz: Fixed.

sysml/src/training/25. Transitions/Local Clock Example.sysml
hpdekoning: OK. The `"T"` is needed in the string concatenation to be a valid `val: Iso8601DateTimeEncoding` actual parameter to `Iso8601DateTime`.

himi:
Ed, thank you for your answer.  I found a bit strange behavior of the validation.  In KerML, the number of parameter  could be checked:

```
	class C1 {
		feature c11;
		feature c12;
	}
	feature f2 = new C1(1, 3, 4);
```

This causes an error of `Bound features should have conforming types".  However, in SysML, it does not seem to be checked:

```
	attribute def AD {
		attribute a1 : ScalarValues::String;
		attribute a2;
	}
	
	attribute attr = new AD("str", 5, 6);
```

Am I missing something?  I found the parameter types are correctly checked. So `new AD(4, 5, 6)` raises an error.

seidewitz:
> Ed, thank you for your answer. I found a bit strange behavior of the validation. In KerML, the number of parameter could be checked:
> 
> ```
> 	class C1 {
> 		feature c11;
> 		feature c12;
> 	}
> 	feature f2 = new C1(1, 3, 4);
> ```
> 
> This causes an error of `Bound features should have conforming types". However, in SysML, it does not seem to be checked:
> 
> ```
> 	attribute def AD {
> 		attribute a1 : ScalarValues::String;
> 		attribute a2;
> 	}
> 	
> 	attribute attr = new AD("str", 5, 6);
> ```
> 
> Am I missing something? I found the parameter types are correctly checked. So `new AD(4, 5, 6)` raises an error.

I believe this is correct per the approved specification, but it is not particularly intuitive, so it is something we might want to address in the RTF.

What is happening here is actually _not_ a check on the number of parameters. Rather, the class `C1` and attribute def `AD` each features inherited from their base library classes, in addition to the owned features given in their declaration. These are ordered after the owned features, so the third argument in each constructor expression is being bound to the first inherited features in each case. 

In the case of `C1`, first inherited parameter happens to be `Occurrence::portionOfLife`. Since this has type `Occurrence`, there is a type conflict with the argument `4`, which is a data value. On the other hand, in the case of `AD`, the first inherited feature happens to be `DataValue::self`. This has type `DataValue`, so there is no conflict with the argument `6`.
(Of course, re-binding `self` is semantically inconsistent, but that is not being checked statically right now.) If you changed `AD` to be an item def instead of an attribute def, then you would get a type conflict on the third argument, just like in the case of the class `C1` in KerML.

himi:
Ed, thank you for the detailed explanation.  I confirmed if I change `attribute def` to `item def`, I got `Bound features should have conforming types` error.  Also if I changed to `new AD("str", 5, 6, 7, 8, 9);`, I got `Must correspond to one feature of instantiated type` error.

himi (approved):

PR #642, opened 2025-04-29
ST6RI-838 Control node bodies (SYSML2_-262)

seidewitz:
This PR implements the resolution to the following issue, as approved in SysML v2 FTF 2 Ballot 7:

- [SYSML2_-262](http://issues.omg.org/issues/SYSML2_-262) Flows cannot connect ControlNodes

(This should have been implemented in release [2025-02](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/releases/tag/2025-02), but it was missed.)

In the SysML abstract syntax, `ControlNodes` (i.e., **`fork`**, **`join`**, **`decide`** and **`merge`** nodes) are kinds of `ActionUsages` that are typed by _`ControlActions`_ from the Systems Model Library. The abstract syntax allows control nodes to have any sort of owned members that other kinds of action usages can. However, the textual notation syntax previously allowed only annotations to be declared in the body of a control node. The resolution to SYSML2_-262 corrects this by revising the SysML textual notation grammar to allow control nodes to have regular action bodies.

In particular, with this change, it is possible to declare control nodes with parameters, enabling the modeling of flows through them. For example:
```
action def A {
    action a1 {
        out x1;
    }
    then j1;

    action a2 {
        out x2;
    }
    then j1;

    flow a1.x1 to j1.x1;
    flow a2.x2 to j1.x2;

    join j1 {
        in x1;
        in x2;
        out y = (x1, x2);
    }
    then a3;

    flow j1.y to a3.y;

    action a3 {
        in y;
    }
}
```
Note that it is still necessary to have successions (or succession flows) into and out of control nodes in order to get the proper "control" semantics.


himi:
I confirmed it worked.  However, the graphical notation will be an open issue.   The visualization result of the example in action view is as below:

image

Since the standard notation of join node is just a thick line and we have not discussed any notations to specify parameters with it, the flow targets are not rendered.

himi (approved):

seidewitz:
> Since the standard notation of join node is just a thick line and we have not discussed any notations to specify parameters with it, the flow targets are not rendered.

Yes, updating the graphical notation was specifically left open in the resolution to SYSML2_-262, because we didn't have time to address it. However, for now, just showing the flows as directly incoming/outgoing from the control node symbol is reasonable.

PR #644, opened 2025-04-30
ST6RI-836 Add option that allows changing the API basePath in a running Jupyter kernel

TheKorpos:
Previously, the REST API base path could be set in the Jupyter kernel only as part of the deployment process: either during the kernel installation by using `--api-base-path` or by using an environmental variable.

This PR adds a new `%repo` command that allows users to change the API base path in the running kernel. The usage of this command is:
```
%repo [<BASE PATH>]

Set the API base path for the repository accessed by the %projects, %publish and %load commands.
<BASE PATH> is a URL (possibly including port number), such as: https://my.domain.com/sysml_repo:9000.
If <BASE PATH> is not given, the current repository base path is printed.
If <BASE PATH> is given, the repository base path is set to this.
```

### DIFF
--- a/org.omg.sysml.interactive/src/org/omg/sysml/interactive/SysMLInteractive.java
+++ b/org.omg.sysml.interactive/src/org/omg/sysml/interactive/SysMLInteractive.java
@@ -489,7 +489,7 @@ public class SysMLInteractive extends SysMLUtil {
 	 * @return output of the command
 	 */
 	public String load(Map<String, String> parameters) {
-		counter++;
+		this.counter++;
 		
 		if (parameters.containsKey(HELP_KEY)) {
 			return SysMLInteractiveHelp.getLoadHelp();
@@ -502,6 +502,9 @@ public class SysMLInteractive extends SysMLUtil {
 		if (parameters.containsKey(BRANCH_ID_KEY) && parameters.containsKey(BRANCH_NAME_KEY)) {
 			return "ERROR:Branch name and id cannot be provided at the same time\n";
 		}
+		
+		System.out.println("API base path: " + apiBasePath);
+		System.out.println();
 		
 		final ProjectRepository repository = new ProjectRepository(apiBasePath);
 		final RemoteProject project;
@@ -541,8 +544,6 @@ public class SysMLInteractive extends SysMLUtil {
 			return "ERROR:Branch doesn't exist\n";
 		}
 		
-		System.out.println("API base path: " + apiBasePath);
-		System.out.println();
 		System.out.println("Selected branch " + branch.getName() + " (" + branch.getRemoteId().toString() + ")");
 		
 		Revision headRevision = branch.getHeadRevision();


seidewitz (changes requested):
`%api-base-path` is an awkward name for a command. How about something like `%repo` instead. The help text should then say:
```
Usage: %repo [<BASE PATH>]

If <BASE PATH> is not given, print the current repository base path.
If <BASE PATH> is given, set the repository base path.

<BASE PATH> is a URL giving the API base path for the repository access by the %projects, %publish and %load commands. 
For example: https://my.domain.com/sysml_repo
```

TheKorpos:
> `%api-base-path` is an awkward name for a command. How about something like `%repo` instead. The help text should then say:
> 
> ```
> Usage: %repo [<BASE PATH>]
> 
> If <BASE PATH> is not given, print the current repository base path.
> If <BASE PATH> is given, set the repository base path.
> 
> <BASE PATH> is a URL giving the API base path for the repository access by the %projects, %publish and %load commands. 
> For example: https://my.domain.com/sysml_repo
> ```

Hi @seidewitz,
I've udpated the magic command name and help strings.

For the example URL I used the default value (http://sysml2-dev.intercax.com:9000), it's more consistent with the new help strings, and default values we used in [ST6RI-178](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/633).

Also, I removed the part 'The api base path is:' from the output of this command so now it simply prints the current base path.

seidewitz:
> For the example URL I used the default value (http://sysml2-dev.intercax.com:9000), it's more consistent with the new help strings, and default values we used in [ST6RI-178](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/633).

I specifically do _not_ want to put a real domain in the help text. The Intercax deployment will not be the default forever, and may even change in the near future. And the default should particularly _not_ be the `sysml2-dev` server. This is for internal development use, and the URL is not supposed to be public.

TheKorpos:
> I specifically do _not_ want to put a real domain in the help text. The Intercax deployment will not be the default forever, and may even change in the near future. And the default should particularly _not_ be the `sysml2-dev` server. This is for internal development use, and the URL is not supposed to be public.

I see. I changed it to the one you wrote.

seidewitz:
This probably has more to do with ST6RI-682 than ST6RI-836, but I didn't notice it until I changed repositories.

After doing `%repo "http://sysml2-sst.intercax.com:9000"`, I got an exception when I tried reading a crtain project that was already in that repository:
```
%load "Spacecraft project with Element CRUD - 2025-05-07 11:54:28.544540"

API base path: http://sysml2-sst.intercax.com:9000/

Selected branch main (7824e4cc-c649-405b-9871-44ad2ff07e9b)
Downloading model...

---------------------------------------------------------------------------
java.lang.UnsupportedOperationException: Revision has no remote!
	at org.omg.sysml.util.repository.Revision.fetchRemote(Revision.java:84)
	at org.omg.sysml.interactive.SysMLInteractive.load(SysMLInteractive.java:555)
	at org.omg.sysml.interactive.SysMLInteractive.load(SysMLInteractive.java:529)
	at org.omg.sysml.jupyter.kernel.magic.Load.load(Load.java:69)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at io.github.spencerpark.jupyter.kernel.magic.registry.Magics.invoke(Magics.java:89)
	at io.github.spencerpark.jupyter.kernel.magic.registry.Magics.access$000(Magics.java:8)
	at io.github.spencerpark.jupyter.kernel.magic.registry.Magics$LineReflectionMagicFunction.execute(Magics.java:149)
	at io.github.spencerpark.jupyter.kernel.magic.registry.Magics.applyLineMagic(Magics.java:26)
	at org.omg.sysml.jupyter.kernel.SysMLKernel.eval(SysMLKernel.java:95)
	at io.github.spencerpark.jupyter.kernel.BaseKernel.handleExecuteRequest(BaseKernel.java:334)
	at io.github.spencerpark.jupyter.channels.ShellChannel.lambda$bind$0(ShellChannel.java:64)
	at io.github.spencerpark.jupyter.channels.Loop.lambda$new$0(Loop.java:21)
	at io.github.spencerpark.jupyter.channels.Loop.run(Loop.java:78)
```
Other projects seem to load OK. The name of this particular project is not consistent with the old format for publishing from Jupyter, so I suspect it was added by someone using the API directly and may be ill-formed in some way. However, if that is the case, it would be better if we generated some sort of error message, rather than an exception, if possible

TheKorpos:
> java.lang.UnsupportedOperationException: Revision has no remote!

When a branch has no head commit the logic creates and returns a local (no remote state) revision. The future head commit of the branch. Once this commit is published (delta between empty remote state and local state) it becomes 'remote'.
The `UnsupportedOperation `exception happens when `fetch` is called on a revision that has no remote.

In this case a project is empty (no commits) but `%load` still tires to `fetch` the non-existing remote state.

 I added a check to prevent this from happening and show the following error for empty branches:
```
ERROR:Branch has no head commit
```

I've made another cosmetic change. When `%load --id=<PROJECT ID>` is used and the project is not found an exception was thrown (informing about error code: 404). I changed that to the message `ERROR:Project doesn't exist` to match the output with `%load --name=<PROJECT NAME>`.

Additionally, it might be a good idea to move the printing of repository URL to an earlier stage in `%load`. Currently it happens after the error messages, so there is no information in case of a `Project doesn't exist` error about witch repository being used.

seidewitz:
> Additionally, it might be a good idea to move the printing of repository URL to an earlier stage in `%load`.

Done.

seidewitz (approved):

seidewitz:
Note: PR description has been updated per help text changes made in PR #649.

PR #645, opened 2025-04-30
ST6RI-840 Constructor expression evaluation (KERML_-224)

seidewitz:
The resolution of the following issue, approved in KerML FTF2 Ballot 7 allowed constructor expression to be model-level evaluable.

- [KERML_-224 ](https://issues.omg.org/issues/KERML_-224)Corrections to Ballot 5 resolutions

The update to the `ConstructorExpression::modelLevelEvaluable operation` from this resolution was already implemented. This PR updates `ModelLevelExpressionEvaluator` to actually evaluate constructor expressions.

A constructor expression is parsed as an expression whose `instantiatedType` is instantiated by its `result` parameter, with `argument` expressions bound to appropriate features of the `instantiatedType`. For example, given the (KerML) declaration
```
struct S {
   feature a;
   feature b;
}
```
then the constructor expression `S(1,2)` is semantically equivalent to the `result` of
```
expr :> Performances::constructorEvaluations {
    return feature :>> result : S {
        feature :>> a = 1;
        feature :>> b = 2;
    }
}
```
Therefore, the model-level evaluation of a constructor expression simply returns its `result`. So, given the following:
```
feature s = new S(1,2);
feature s_a = s.a;
feature s_b = s.b;
```
the feature `s` evaluates to the `result` parameter of the constructor expression, so that `s_a` evaluates to the literal integer `1` and `s_b` evaluates to the literal integer `2`.

The following is an equivalent SysML model that evaluates in the same way:
```
part def P {
    attribute a;
    attribute b;
}
part p = new P(1,2);
attribute p_a = p.a;
attribute p_b = p.b;
```

### DIFF
--- a/org.omg.sysml.interactive.tests/src/org/omg/sysml/interactive/tests/ModelLevelEvaluationTest.java
+++ b/org.omg.sysml.interactive.tests/src/org/omg/sysml/interactive/tests/ModelLevelEvaluationTest.java
@@ -300,6 +300,21 @@ public class ModelLevelEvaluationTest extends SysMLInteractiveTest {
 	}
 	
 	@Test
+	public void testConstructorEvaluation() throws Exception {
+		SysMLInteractive instance = getSysMLInteractiveInstance();
+		process(instance,
+				"part def P { attribute a; attribute b; } " +
+				"part p1 = new P(1, 2); " +
+				"part p2 = new P(b = p1.b, a = p1.a);");
+		assertEquals(true, evaluateBooleanValue(instance, null, "p1 istype P"));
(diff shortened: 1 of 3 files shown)


himi:
I tried your examples, but it did not work as they are:

* `Performances::constructorExpressions` should be `Performances::constructorEvaluations`
* `S(1, 2)` should be `new S(1, 2)` 
*  `P(1, 2)` should be `new P(1, 2)`

After that, I could visualize then with `EVAL` style:
Screenshot 2025-04-30 at 11 05 20 PM

himi (approved):
I confirmed it worked with many patterns including the cases using `calc`. And I also confirmed that case is not model level evaluable.

PR #646, opened 2025-05-03
ST6RI-837: Do not render <<variation>> for EnumerationDefinition

himi:
By the side effect of the fix by https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/pull/560, `<<variation>>` is rendered for `EnumerationDefinition` because its `isVariation` is true.  However it should not be rendered.  This PR fixes this issue.


himi:
The fix is quite simple and you can easily check it by rendering `enum def`.  For example,
```
package ST6RI837S {
	enum def E1 {
		enum e1;
		enum e2;
	}	
	enum def E2;
}
```

is rendered as:

Screenshot 2025-05-03 at 4 56 29 PM

seidewitz (approved):

PR #655, opened 2025-05-30
ST6RI-844 Update to Eclipse 2025-03

ujhelyiz:
This PR updates the Eclipse platform to Eclipse version 2025-03. 

- Updates Java requirement to version 21.
- Updates Target platform and Oomph setup accordingly.
- Fixes a few minor issues that were found during build review.
  - All projects now explicitly state UTF-8 character encoding.
  - Fixes a few other metadata issues.
  - Removes some redundant configuration from the POM.
  - Updates Asciidoctor dependency to latest release.
  - Fixes some deprecated code issues.


seidewitz:
I updated various README files to reflect the new Eclipse and Java versions, and bring them up to date otherwise as necessary.

seidewitz:
> I have fixed the launching issues by adding a specific dependency that is only used inside Eclipse. The root cause of the issue is that those tests are run as JUnit tests over Eclipse plugin projects, which behave differently regarding transitive dependencies. By explicitly specifying the problematic dependencies, these issues can be fixed.

Thanks, it works now. Why wasn't this a problem before?

PR #663, opened 2025-06-26
ST6RI-843 Annotate adapter transformation code with implemented semantic constraints

TheKorpos:
This PR annotates semantic constraints in the form of comments and javadoc tags in the transformation code.


seidewitz:
Replaced with PR #665.

PR #667, opened 2025-07-03
ST6RI-863 Actions in a case will cause a crash of the visualization (PlantUML)

himi:
Since VCase cannot handle actions properly, the model below causes a crash.
```
case c1 {
        first start;
        then action a;
}
```
This PR fixes this issue and lets the visualizer properly render actions as well as cases.

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VBehavior.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VBehavior.java
@@ -45,11 +45,11 @@ import org.omg.sysml.lang.sysml.TransitionUsage;
 
 public abstract class VBehavior extends VDefault {
     protected boolean isDoneAction(Feature f) {
-        return "Actions::Action::done".equals(f.getQualifiedName());
+        return f != null && f.specializesFromLibrary("Actions::Action::done");
     }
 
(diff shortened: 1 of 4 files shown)


seidewitz:
This fixes the crash, so the example case renders as
image
If the example is changed to an analysis case or a verification case, it renders similarly. However, a use case renders as
image
without the start symbol. Why is that?

himi:
Opps, it is because `UseCases::UseCase::start` redefines `Actions::Action::start` and the current code uses:
        `return "Actions::Action::start".equals(f.getQualifiedName());`

Is there any good utility function to check if the element belongs to `start` or `done`?   If not, I will simply compare the qualified name with `UseCases::UseCase::start`  as well.

seidewitz:
> Opps, it is because `UseCases::UseCase::start` redefines `Actions::Action::start` and the current code uses: `return "Actions::Action::start".equals(f.getQualifiedName());`
> 
> Is there any good utility function to check if the element belongs to `start` or `done`? If not, I will simply compare the qualified name with `UseCases::UseCase::start` as well.

If `f` is a `Type`, then you can check `f.specializesFromLibrary("Actions::Action::start")`.

himi:
@seidewitz,  thank you!  It works quite well.

image

seidewitz (approved):

PR #687, opened 2025-09-04
ST6RI-683 Programmatic interface

TheKorpos:
This PR introduces a new programmatic API that provides an easy setup for the language infrastructure (inlcuding reading the standard library) and interfaces for model reading/processing.

### Example usages
```
SysMLAccess access = SysMLAccess.createFullyFeatured("path_to_lib"); //init the lang. infrastructure and set library path
access.loadLibrary(); //load the library
access.parse("myres.sysml", "package A { }"); //parse a string and put it in a resource
Element element = access.resolve("$::A"); //resolve and element by qualified name
```
**with REST access**
```
SysMLAccess access = SysMLAccess.createFullyFeaturedWithAPIAccess("path_to_lib", "http://localhost:9000");
access.loadLibrary();
List<Resource> loadedResources = access.getAPIAccess().load(new APIProjectDescriptor().projectName("myProject"));
```

### Changes
**Added** 

- ```SysMLAccess```: main interface to interract with the language infrastructure
  - extends SysMLUtil 
  - library load
  - parsing string input
  - parsing files
  - resolving by qn
 - ```SysMLAPIAccess```: simplified interfaces for publish/load
   - loading
   - publishing
   - listing projects from repository
 - ```SysMLParseResult```: wrapper class for parsing/validation issues 
 - ```SysMLAccess.Builder```: builder for SysMLAccess instaces (mainly for future extendability and configuration)

- ```SysMLAccessTest```: test cases for the new API

**Modified**

- `SysMLUtil`
  - revised handing folders and file extensions
  - pushed some exception handling to the user side (e.g IOExceptions)
  - moved static helpers from `SysMLInteractiveUtil` (these are mostly for printing parts of the model in a human readable format)
- `SysMLInteractive`
  - now extends `SysMLAccess`. It's interfaces remain the same but there are some implementation changes (e.g moving functionality up to `SysMLAccess`)
- comand line utils:
   - added added try/catch blocks for newly thrown exceptions

**Removed**

- SysMLInteractiveUtil (moved static helpers to `SysMLUtil`)


himi:
I noticed SysMLAccess.build() and SysMLInteractive.createInstance() overlaps in many ways.  Can we make createInstance() use build()?  Also don't we need to call 
`KerMLxStandaloneSetup.doSetup();`
`SysMLxStandaloneSetup.doSetup();`
in build()?

So, for extension purpose, I think we change build() and createInstance() by using change generics.  For example,

```
	public static <T extends SysMLAccess> T build(ISetup setup, java.lang.Class<T> cls) {
                       Injector injector = setup.createInjectorAndDoEMFRegistration();
			
			T access = injector.getInstance(cis);
                       return access;
        }

        public static SysMLAccess build() {
                 return build(new SysMLStandaloneSetup(), SysMLAccess.class);
	}
```
I can propose a commit to use this mechanism.  So SysMLInteractive().createInstance() can call build(setup, SysMLInteractive.class) to create a SysMLInteractive instance.

ujhelyiz:
Where to call the doSetup is quite an interesting question, for two reasons:

1. These methods should be called _exactly_ once. If called multiple times, e.g., an application wants to start multiple SysMLAccess instances for separation purposes, it could cause tricky surprises.
2. If you want the setups to be customizable,  it might be better for this service not to call it, as in that case, the replacement is trivial.

PR #689, opened 2025-09-07
ST6RI-867 Trigger is not shown properly (PlantUML)

himi:
The trigger name of `Sig` below is wrongly rendered as `SigOccurrence`.  This PR fixes it by rendering the first parameter (`Sig`) only.

```
    attribute def Sig;
    
    state s1 {
    	state s11;
    	accept Sig then done;
    }
```

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VBehavior.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VBehavior.java
@@ -263,9 +263,13 @@ public abstract class VBehavior extends VDefault {
                 return "";
             }
         };
-        for (Feature f: s.getParameter()) {
-            v.visit(f);
-        }
+        List<Feature> ps = s.getParameter();
+
+        // We render the first parameter only because the second parameter of
+        // AcceptActionUsage is receiver, which should not be rendered.
+        if (ps.isEmpty()) return "";
+        v.visit(ps.get(0));
+
         String str = v.getString();
         if (str.isEmpty()) {
             return getText(s);
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VBehavior.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VBehavior.java
@@ -1,6 +1,7 @@
 /*****************************************************************************
  * SysML 2 Pilot Implementation, PlantUML Visualization
- * Copyright (c) 2020-2024 Mgnite Inc.
+ * Copyright (c) 2020-2025 Mgnite Inc.
+ * Copyright (c) 2025 Model Driven Solutions, Inc.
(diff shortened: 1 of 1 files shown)


seidewitz:
I updated `VBehavior::convertToDescription` so it uses `TransitionUsage` operations to get the trigger action, guard expression and effect action. I also updated `triggerToText` so it is similar to the corresponding code in `addAcceptActionUsage`, accounting for trigger actions with declared names and receivers. The model
```
package ST6RI_867 {
    attribute def Sig;
    
    state s0 {
    	state s01;
    	then done;
    }
    state s1 {
    	state s11;
    	accept Sig then done;
    }
    state s2 {
    	state s21;
    	accept s : Sig then done;
    }
    state s3 {
    	state s31;
    	accept Sig via this then done;
    }
}
```
is now visualized as
image

Please review and let me know if this is acceptable.

himi:
@seidewitz, thank you!  That is great, but I found two issues on `AcctionActionUsage` rendering.  So far, we use `:` to show type name even if it does not have a default name (`payload`).  So for example:
```
attribute def Sig;
action a0 {
    accept : Sig;
}
```
is rendered as
image

I want to make it consistent and should we eliminate `:` if it does not have a default name?

Also I found if there is no declared name, a proper effective name is not rendered properly:

```
	attribute def Sig;
	attribute ss : Sig;
    state s1 {
    	state s11;
    	accept :>> ss then done;
    }
```
Screenshot 2025-09-22 at 1 39 46 PM

I can fix these if we always omit or render `:` in the case we do not have a proper payload name.  `appendNameType()` is designed to support such case and I want to apply it to trigger's `AcceptActionUsage` as well.

seidewitz:
> I can fix these if we always omit or render `:` in the case we do not have a proper payload name. `appendNameType()` is designed to support such case and I want to apply it to trigger's `AcceptActionUsage` as well.

I think it is better to always omit the `:` if there is not a proper payload name.

himi:
Thank you.  I updated based on your comment:
```
package ST6RI867 {
	attribute def Sig;
	
	action a0 {
	    accept Sig;
	    accept s:Sig;
	}
	
	    
    state s0 {
    	state s01;
    	then done;
    }
    
    state s1 {
    	state s11;
    	accept Sig then done;
    }


    state s2 {
    	state s21;
    	accept s : Sig then done;
    }
    
    state s3 {
    	state s31;
    	accept Sig via this then done;
    }
    
    
    attribute ss : Sig;
	state s4 {
    	state s41;
    	accept :>> ss then done;
    }
    state s5 {
    	state s51;
    	accept :>> ss : Sig then done;
    }
   
}
```
is rendered as:
image

Note that we do not show `Sig` in state `s4` example because we do not render inherited type name by default.

seidewitz (approved):
Looks good!

PR #696, opened 2025-10-02
ST6RI-766 Update to JupyterLab version 4.x in the Jupyter deployment

TheKorpos:
This PR includes dependency and code updates for the `org.omg.sysml.jupyter.jupyterlab` labextension to support Jupyter 4.x. Older versions are no longer supported.

This PR doesn't change how the extension is packaged. Installing it into Jupyter is still possible with `jupyter labextension install` (which is deprecated but still works in 4.x).

#### Dependency updates
- updated dependencies in `package.json`
  - `"@jupyterlab/application": "3.x"` -> `"4.x"`
  - `"typescript": "<4.4.0"` -> `"~5.8.3"`
  - `"@jupyterlab/codemirror": "^0.0.98"` -> `">=4.0"`
- added new dependency
  - `"@codemirror/legacy-modes": "^6.3.2"`
- updated compiler options
  - `es2018` -> `es2023`

#### Code changes
Previously, the syntax highlight was registered by adding a new "mode" to Codemirror. This is no longer an option. Instead, the language has to be registered via `IEditorLanguageRegistry` by the front-end plugin. Luckily, the underlying parser is still present and can be registered in "legacy" mode, so most of the configuration can be easily ported. However, a new `foldService` had to be implemented to support code folding between braces (`{ ... }`).

`org.omg.sysml.jupyter.jupyterlab/src/main/`
- `plugin.ts` – updated the implementation of the front-end plugin
  - language registry is used to register the `sysml` language
  - editor extension registry is used to register the `sysml-fold-extension`
- `mode.ts` – created a `clike` parser to replace the CodeMirror mode
  - reused (most of) the configuration from older code
- `fold.ts` – implemented `sysmlFoldServiceSelection` to support code folding

`tool-support/syntax-highlighting/jupyter/`
- `xtext_grammar_converter.py` – removed support for keyword highlighting in classic Jupyter Notebook, Jupyter Lab should always be used instead
- `jupyter/mode_template.js` – updated the template for `mode.js` used by the `xtext_grammar_converter`

### DIFF
--- a/tool-support/syntax-highlighting/jupyter/mode.ts
+++ b/tool-support/syntax-highlighting/jupyter/mode.ts
@@ -28,7 +28,7 @@ import { clike } from '@codemirror/legacy-modes/mode/clike';
 
 const f_wordify = (h: any, s: string) => ({...h, [s]: true});
 export const sysmlparser = clike({
-        name: 'clike',
+        name: 'sysml',
         keywords: [
             "about", "abstract", "accept", "action", "actor", "after", "alias", "all", "allocate", "allocation",
             "analysis", "and", "as", "assert", "assign", "assume", "at", "attribute", "bind", "binding", "by",
--- a/tool-support/syntax-highlighting/jupyter/mode_template.ts
+++ b/tool-support/syntax-highlighting/jupyter/mode_template.ts
@@ -28,7 +28,7 @@ import { clike } from '@codemirror/legacy-modes/mode/clike';
 
 const f_wordify = (h: any, s: string) => ({...h, [s]: true});
 export const sysmlparser = clike({
-        name: 'clike',
+        name: 'sysml',
         keywords: [
             "$KEYWORDS"
         ].reduce(f_wordify, {}),


seidewitz (changes requested):
I noticed that the new extension does not support folding of code between braces. That is, the old mode definition included:
```
modeProps: {
    fold: ['brace'],
}
```
but I don't see anything similar in the new `sysmlparser` definition. If there is no way to do this in the new approach, that is a serious degradation of functionality that I know is commonly used by users.

seidewitz:
Also, the file `org.omg.sysml.jupyter.kernel/src/main/resources/sysml/kernel.js` still sets CodeMirror mode for classic Jupyter Notebook. Will this still work? Do we even need to support classic Jupyter Notebook anymore?

TheKorpos:
> Also, the file `org.omg.sysml.jupyter.kernel/src/main/resources/sysml/kernel.js` still sets up a CodeMirror mode for classic Jupyter Notebook. Will this still work? Do we even need to support classic Jupyter Notebook anymore?

No, this shouldn't work. I always go for jupyter lab. I'm not sure if there is a high demand for classic Notebook.

seidewitz:
> I think we need to use `foldService` for CodeMirror V6.

The problem is that the v6 `foldService` doesn't work for legacy language support, because it requires real parsing of the language text. As far as I could determine, it is possible to get legacy code folding to work, but one needs to write a routine to determine the scope of the folding (e.g., based on '{ ... }' delimiters). Surprisingly, I couldn't find anything in CodeMirror or otherwise publicly available to do this. I actually got what looked like a pretty good code suggestion from AI (even though using AI for code is against my usual principles!), so I might try seeing if that can work.

In any case, if we can't find a way to get code folding incorporated fairly quickly, I am going to defer this PR to the next sprint. I don't want to deploy it without code folding.

himi:
We can use `foldService` with `@codemirror/legacy-modes/mode` but I do not know how we enable CodeMirror 6 extensions in legacy JupyterLab extensions.  (As I said, I already switched to new JupyterLab, Python Wheel-based extension) Anyway, it would not be a simple task.

himi:
@seidewitz, I made ST6RI-766-1 branch and could successfully make the fold service work.  Can I change the source branch to it?

himi:
Opps.  `mode.ts` define it as 'variable'.  I will fix it.

seidewitz (approved):

seidewitz:
@himi 
Looks good! I merged branch ST6RI-766-1 into ST6RI-766 and changed this PR to a ready status again.

@TheKorpos 
Would you be able to review Hisashi's update?

TheKorpos:
@himi Thank you for finishing this up!

The code looks good and seems to work as intended.

I also caught a mistake I made in the packaging that caused some infinite inclusion of the output directory. I managed to get a 50GB tgz as a result... I fixed the issue by re-adding the 'files' field, that I had removed, to the `package.json.`

himi:
I found the first cell in newly created notebooks is initialized as 'text/plain' and it won't enable foldService.  So I enabled foldSerivce to 'text/plain' as well and check if the language is `sysml` also in the foldService.  For this purpose, I named the language as `sysml` instead of `clike` in `mode.ts`

PR #701, opened 2025-10-08
ST6RI-881 Conditional computation of baseType of semantic metadata doesn't work

seidewitz:
This PR corrects the evaluation of the feature value expression of the `baseType` of semantic metadata.

Consider the following KerML model:
```
class C;
struct S;

metaclass M :> Metaobjects::SemanticMetadata {
  :>> annotatedElement : KerML::Class
  :>> baseType = 
      if annotatedElement istype KerML::Structure ? 
          S meta SysML::Type else C meta SysML::Type;
}

#M struct T;
```
The structure `T` should have an implicit specialization to the base structure `S`, but, instead, it previously just received an implied specialization to `Objects::Object`. This is because the evaluation of the feature value expression for the `baseType` feature was computed using the `annotatedElement` `T` as the target, rather than the annotating metadata feature (where `annotatedElement` is a feature). The PR corrects this.

### DIFF
--- a/kerml/src/examples/Simple Tests/MetadataTest.kerml
+++ b/kerml/src/examples/Simple Tests/MetadataTest.kerml
@@ -38,4 +38,18 @@ package MetadataTest {
 	    }
 	}
 	
+    class CC;
+    struct SS {
+        feature cc : CC;
+    }
+    
+    metaclass M :> Metaobjects::SemanticMetadata {
+      :>> annotatedElement : KerML::Class;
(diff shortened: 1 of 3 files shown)


seidewitz:
@gke3 
The sample is correct, and it works for me, both in the Eclipse and Jupyter deployments of 2025-10.

Try this:
```
part p {
    protected attribute x;
}
attribute y = p.x; // ERROR:Couldn't resolve reference to Element 'x'.
```
If you _don't_ get an error, then you don't have 2025-10 installed.

PR #710, opened 2025-11-01
ST6RI-892 State transition visualization cause null pointer exception (PlantUML)

seidewitz:
This PR fixes a bug that can cause a null-pointer exception when visualizing a state transition in a model that has been loaded from a repository.

**Background**

Consider the following simple state model:
```
package StateVizTest {
    item def Cmd {
        attribute turnOn : ScalarValues::Boolean;
    }
    state def States {
        entry; then off;
        state off;                                  
        state on;
        transition off_To_on
            first off
            accept cmd : Cmd
                if cmd.turnOn
            then on;
        transition on_To_off
            first on
            accept Cmd
                do action Test
            then off;
    }
}
```
This would normally be visualized as
image
However, if the model is published to a repository and then later loaded back into memory rather than being parsed from text, attempting to visualize it results in a null pointer exception.

Normally, the guard expressions and effect actions are rendered using the textual notation from which they were parsed. However, when the model is loaded from a repository, the original textual notation is no longer available. The exception occurs because the code was attampting to call `trim()` on null text strings.

**Code changes**

This PR fixes the bug by updating the code in `VBehavior.convertToDescription` to check if textual notation strings are null before calling `trim()` on them. After this change, the above model, when loaded from a repository, is visualized without an exception as:
image
This is probably the best that can be done for now, without re-serializing the text from the abstract syntax or storing the original text as an annotation in the repository.

**Note.** The lack of rendering of typing on features in the loaded model is due to the `featuredType` property of `FeatureTyping` relationships not being set properly for the loaded model. This is not a visualization issue.

### DIFF
--- a/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VBehavior.java
+++ b/org.omg.sysml.plantuml/src/org/omg/sysml/plantuml/VBehavior.java
@@ -259,8 +259,9 @@ public abstract class VBehavior extends VDefault {
         StringBuilder sb = new StringBuilder();
 
         Expression payload = aau.getPayloadArgument();
-        if (payload != null) {
-            sb.append(getText(payload));
+        String payloadText = getText(payload);
+        if (payloadText != null) {
+            sb.append(payloadText);
         } else {
             ReferenceUsage ru = aau.getPayloadParameter();
             appendNameAndType(sb, ru, "payload");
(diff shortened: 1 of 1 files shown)


himi (approved):
The fix is straightforward and if it's urgent to fix, please go ahead.  If we have time for this issue, rather I want to introduce expressionToText() even if it's just calling getText() and trim() because it's getting important to render expressions without the original text when we get models via v2 API.

seidewitz:
@himi
If you can make the updates you want in the next few days, go ahead. I would like to get it into the 2025-10 release, though.

himi:
I tried to improve this but I found it involves `ActionUsage` as well as `Expreesion`.   I need to travel now and am not sure to finish in a few days.   So please go ahead with these changes.  Thank you!

PR #716, opened 2025-11-16
ST6RI-893 KerML grammar does not allow constant end features

seidewitz:
This PR corrects the KerML grammar as implemented, to allow association structures to have ends marked with **`const`** as constant variable features (per the KerML Specification, 8.2.4.3.1). For example, the following model previously did not parse, but, with this correction, it does:
```
assoc struct Test {
    const end feature a;
    const end feature b;
}
```


org.omg.kerml.xtext.ui/src-gen/org/omg/kerml/xtext/ui/internal/XtextActivator.java
```diff
@@ -29,7 +29,7 @@ public class XtextActivator extends AbstractUIPlugin {
 	
 	private static XtextActivator INSTANCE;
 	
-	private Map<String, Injector> injectors = Collections.synchronizedMap(Maps.<String, Injector> newHashMapWithExpectedSize(1));
+	private Map<String, Injector> injectors = Collections.synchronizedMap(new HashMap<>(2));
```
vincemolnar: What justifies this change?

vincemolnar (approved):
All good, but I had some questions - see comments.

PR #718, opened 2025-11-27
ST6RI-884 Model-level evaluation conformance

seidewitz:
This PR improves the conformance of the implementation of the evaluation of model-level evaluable expressions with the SysML 2.0 specification.

**Changed library function implementations**

The following functions had been implemented as model-level evaluable, but they are not specified as model-level evaluable for SysML 2.0. They are no longer considered model-level evaluable in the implementation, and their implementation has been moved from `org.omg.sysml.expressions.functions` to `org.omg.sysml.execution.expressions.functions`.

- `NumericalFunctions::prod`
- `NumericalFunctions::sum`
- `SequenceFunctions::excludes`
- `SequenceFunctions::includes`
- `SequenceFunctions::isEmpty`
- `SequenceFunctions::notEmpty`
- `SequenceFunctions::size`
- `StringFunctions::Length`
- `StringFunctions::Substring`

**New library function implementations**

The following functions have now been implemented as model-level evaluable, per the specification. Unfortunately, a problem with the current specification of the model-level evaluablity of body expressions (see [KERML11-178](https://issues.omg.org/issues/KERML11-178)) prevents them from be used in useful cases in model-level expressions. However, they can now be used in non-model-level evaluable expressions.

- `ControlFunctions::collect`
- `ControlFunctions::select`


himi:
I found `StringFunctions::length` and `size` should be `Length` and `Substring`.    Also `SequenceFunctions::size` is missing in your description.   I checked they could not be used in metadata since they are no longer model-level evaluable.

himi:
I confirmed `collect` can be used:
```
@MD { test = ControlFunctions::collect((1, 2, 3), true); }
```
But as you said, it's pity that I could not use:
```
@MD { test = ControlFunctions::collect((1, 2, 3), { in x; x * 2 }); }
```

himi (approved):

himi:
Thank you for fixing the description.  (But `SequenceFunction::size` is not correctly quoted)

PR #731, opened 2026-01-27
ST6RI-906 Evaluation of functions from TrigFunctions

seidewitz:
This PR implements the (non-model-level) evaluation of all  the functions from the Kernel Function Library package `TrigFunctions`.

**Implemented Functions**

The following functions from `TrigFunctions` are implemented in this PR.

- `deg`
- `rad`
- `sin`
- `cos`
- `tan`
- `cot`
- `arcsin`
- `arccos`
- `arctan`

All functions are implemented using the corresponding functions from `java.lang.Math` (even for those functions whose definition is given in `TrigFunctions`).

In addition, the feature `TrigFunctions::pi` is implemented to have the value `java.lang.Math.PI` (even though this has lower precision than what is specified in `TrigFunctions`).

**Other Changes**

1. Added a general capability to implement values for "library features" like `TrigFunctions::pi`.
2. Created an `ExpressionEvaluator` interface, which can be used to type `evaluator` parameters and fields more generally than `ModelLevelExpressionEvaluator`. `ModelLevelExpressionEvaluator` implements `ExpressionEvaluator`.
3. Fixed some problems with the `org.omg.sysml.execution.expressions.ExpressionEvaluator` extension of `ModelLevelExpressionEvaluator`.
   - Corrected `instantiateInvocation` to copy `instantiatedType` from the original `InvocationExpression`, rather than `ownedTyping`.
   - Updated the evaluation of `Features` so that unbound input parameters are considered to have a null value (rather than returning the `Feature` itself).

### DIFF
--- a/org.omg.sysml.execution/src/org/omg/sysml/execution/expressions/ExpressionEvaluator.java
+++ b/org.omg.sysml.execution/src/org/omg/sysml/execution/expressions/ExpressionEvaluator.java
@@ -1,6 +1,6 @@
 /*******************************************************************************
  * SysML 2 Pilot Implementation
- * Copyright (c) 2022-2025 Model Driven Solutions, Inc.
+ * Copyright (c) 2022-2026 Model Driven Solutions, Inc.
  *    
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
@@ -27,7 +27,6 @@ import org.omg.sysml.expressions.functions.LibraryFunction;
 import org.omg.sysml.lang.sysml.Element;
 import org.omg.sysml.lang.sysml.Expression;
 import org.omg.sysml.lang.sysml.Feature;
-import org.omg.sysml.lang.sysml.FeatureTyping;
 import org.omg.sysml.lang.sysml.FeatureValue;
 import org.omg.sysml.lang.sysml.Function;
 import org.omg.sysml.lang.sysml.InvocationExpression;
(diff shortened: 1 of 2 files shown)


himi:
I tested these functions
```
part TestTrigFunctions {
    private import TrigFunctions::*;

    attribute pi = rad(180);
    attribute flatDeg = deg(pi);
    attribute sin_pi = sin(pi);
    attribute cos_pi = cos(pi);
    attribute tan_pi4 = tan(pi / 4);
    attribute tan_pi2 = tan(pi / 2);
    attribute cot_pi2 = cot(pi / 4);

    attribute as1d = 2 * arcsin(1);
    attribute ac1 = arccos(1);
    attribute at1q = 4 * arctan(1);
}
```
But because of `UnitBoundedReal`, I got the warnings below due to `arcsin()` and `arccos()` even though the given argument satisfies the constraint:

```
WARNING:Bound features should have conforming types (19.sysml line : 12 column : 33)
WARNING:Bound features should have conforming types (19.sysml line : 13 column : 28)
```

The evaluation results looked good.

image

seidewitz:
> But because of `UnitBoundedReal`, I got the warnings below due to `arcsin()` and `arccos()` even though the given argument satisfies the constraint:

The problem is that `Integer` is a subtype of `Real`, but it is a _peer_ type of `UnitBoundedReal`. The type checking isn't smart enough to understand that the range constraint means that the extent of `UnitBoundedReal` includes the integers 0 and 1. Indeed, the system currently doesn't actually check the constraint at all.

Really, in KerML terms, the type `UnitBoundedReal` should be declared as "sufficient" (i.e., "`*datatype* *all* UnitBoundedReal :> Real"). This would give the type the intended ontological semantics that it includes all `Real` values that meet the constraint. We haven't fully worked out the type checking for sufficient types, though.

For now, if you use `1.0` instead of `1`, the type checking warnings go away.

himi (approved):
I confirmed the warnings disappeared when I changed 1 to 1.0.  Thank you.

PR #737, opened 2026-02-13
ST6RI-913 Exception opening .kermlx and .sysmlx files in editor

seidewitz:
One of the updates in PR #734 was to move the `OperandEList` class to the `org.omg.sysml.util` package and re-implementation. This new implementation did not support the `iterator` method. However, when a `.kermlx` and `.sysmlx` file is opened in the Eclipse editor, `iterator` is called to get an iterator with which to iterate over the values of structural features. Even though the `operands` list is not included in XMI serialization, the system still calls `iterate` on the `OperandEList` for the `operand` structural feature, which resulted in a `UnsupportedOperationException`.

This PR updates `EOperandList.iterator` so that it returns the `emptyIterator`.


seidewitz:
> It might be a different topic, but I found `SysML2XMI sysml/src/validation/10-Analysis and Trades/10d-Dynamics Analysis.sysml` failed with NullPointerException:

This happens when the library directory arguments are not included when using `SysML2XMI`. A default feature value has an implied binding connector whose featuring type is `this.startShot`. But `this` and `startShot` cannot be found without the library models.

In the spirit of being more robust, I have pushed a fix on this branch that avoids the NPE.

PR #743, opened 2026-03-17
ST6RI-921 Invalid feature references are allowed in filter expressions

seidewitz:
This PR removes an exception made in the Pilot Implementation that allowed invalid feature references in filter expressions.

**Background**

Both the KerML Specification (in 7.4.14 Packages) and the SysML Specification (in 7.5.4 Import Filtering) show examples of referencing features using a qualified name notation. For example:
```
filter @Annotations::ApprovalAnnotation and
    Annotations::ApprovalAnnotation::approved and
    Annotations::ApprovalAnnotation::level > 1;
```
This example has subexpressions that are feature reference expressions for non-package-level features, such as `Annotations::Approval::approved`. However, filter expressions are at package level (have no featuring type), and therefore do not have access to non-package-level features. 

This should have caused a `checkConnectorTypeFeaturing` error on the implied binding connector between the referent feature and the expression result. But the Pilot Implementation had an explicit check coded in `FeatureReferenceExpressionAdapter::addReferenceConnector` to _not_ add the binding connector if the expression was in a filter, so no error was reported. There is no such exception in the normative metamodel, despite the examples given in Clause 7 of the specifications (see also [KERML11-183](https://issues.omg.org/issues/KERML11-183) on the specification inconsistency).

**Changes**

Updated `FeatureReferenceExpressionAdapter::addReferenceConnector` to remove the exception for filter expressions.

Feature expressions such as in the example above now cause errors. The corrected version of the example is:
```
filter @Annotations::ApprovalAnnotation and
    (as Annotations::ApprovalAnnotation).approved and
    (as Annotations::ApprovalAnnotation).level > 1;
```

### DIFF
--- a/sysml/src/examples/Vehicle Example/SysML v2 Spec Annex A SimpleVehicleModel.sysml
+++ b/sysml/src/examples/Vehicle Example/SysML v2 Spec Annex A SimpleVehicleModel.sysml
@@ -1536,7 +1536,7 @@ package SimpleVehicleModel{
         package MandatorySafetyGroup {
             /* Parts that contribute to safety AND are mandatory. */
             public import vehicle_b::**;
-            filter @Safety and Safety::isMandatory;
+            filter (as Safety).isMandatory;
         }
     }
     package Views_Viewpoints{
--- a/sysml/src/examples/Vehicle Example/SysML v2 Spec Annex A SimpleVehicleModel.sysml
+++ b/sysml/src/examples/Vehicle Example/SysML v2 Spec Annex A SimpleVehicleModel.sysml
@@ -1536,7 +1536,7 @@ package SimpleVehicleModel{
         package MandatorySafetyGroup {
             /* Parts that contribute to safety AND are mandatory. */
             public import vehicle_b::**;
-            filter (as Safety).isMandatory;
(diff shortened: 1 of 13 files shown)


sysml/src/examples/Vehicle Example/SysML v2 Spec Annex A SimpleVehicleModel.sysml
```diff
@@ -1536,7 +1536,7 @@ package SimpleVehicleModel{
         package MandatorySafetyGroup {
             /* Parts that contribute to safety AND are mandatory. */
             public import vehicle_b::**;
-            filter @Safety and Safety::isMandatory;
+            filter (as Safety).isMandatory;
         }
```
hpdekoning: I think it is correct to leave out `@Safety and ` in the filter expression. In other words: `.isMandatory` is only evaluated on elements of type `Safety` and thus such elements are 'filtered in' if `isMandatory` equals true, correct? This appears to be defined in KerML 7.3.4.6 Feature Chaining and 7.4.9.3 Feature chain expression, but I am not completely sure. For the Annex A model we may want to insert a short comment that explains this.

hpdekoning (approved):
Looks good to merge.

PR #747, opened 2026-03-28
ST6RI-924 Problems with TypeAdapter and AssignmentActionUsageAdapter

seidewitz:
This PR fixes bugs in `TypeAdapter` and `AssignmentUsageAdapter`.

`TypeAdapter`

- Previously, `TypeAdapter::removeUnnecessaryImplicitGeneralTypes` could throw an `NullPointerException` if `implicitEClassGenerals` was null for a certain specialization `EClass` (which can happen during processing of an ill-formed model).
- Added a check on that `implicitEClassGenerals` is non-null before processing it further.

`AssignmentUsageAdapter`

- Previously,  `AssignmentActionUsageAdapter::addTargetRedefinition` called `FeatureAdapter::addFeatureWriteTypes` to satisfy redefinition constraints on `target::StartingAt` and `target::startingAt::accessedFeature` nested in an `AssignmentActionUsage`. However, placing this within `AssignmentActionUsageAdapter` means that the transformation is not done if `Feature::computeImplicitGeneralTypes` is called directly on these features, which can happen during Xtext editor reconciliation of text changes. The result can be spurious validation errors.
- Moved the implementation satisfying these constraints to `FeatureAdapter`, called from `addComputedRedefinitions`.

    **Note.** This change results in the KerML-level `FeatureAdapter` depending on SysML `AssignmentActionUsage`. However, in general, "on-demand" calls to `computeImplicitGeneralTypes` can only be properly handled if the satisfaction of all semantic constrains effecting a metaclass are implemented in the adapter for _that_ metaclass.

### DIFF
--- a/org.omg.sysml/src/org/omg/sysml/adapter/AssignmentActionUsageAdapter.java
+++ b/org.omg.sysml/src/org/omg/sysml/adapter/AssignmentActionUsageAdapter.java
@@ -1,6 +1,6 @@
 /*******************************************************************************
  * SysML 2 Pilot Implementation
- * Copyright (c) 2021 Model Driven Solutions, Inc.
+ * Copyright (c) 2021, 2026 Model Driven Solutions, Inc.
  *    
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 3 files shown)


himi:
I encountered this issue during Xtext editing, but I do not remember any concrete steps to reproduce that NullPointerException.   Do you know anything about it?  So far, it works quite well without NPE :).

himi (approved):
Thank you.  I tried many patterns and could not get any NPEs so far.  In my experience, analysis and subject edit has had caused an NPE often, but I could not reproduce such problems in this version.

PR #749, opened 2026-04-01
ST6RI-925 Problem with deriveTypeFeatureMembership (KERML11-191)

seidewitz:
This PR proactively implements the resolution to the following issue, which the KerML 1.1 RTF has elevated to be a "blocker" issue.

* [KERML11-191](https://issues.omg.org/issues/KERML11-191)  Problem with deriveTypeFeatureMembership

**Background**

Currently (as of KerML 1.0), the OCL for `deriveTypeFeatureMembership` is
```
featureMembership = ownedFeatureMembership->union(
    inheritedMembership->selectByKind(FeatureMembership))
```
The intent here is clearly to not include imported feature memberships. The problem is that `inheritedMembership` _does_ include imported feature memberships from supertypes. Since `Type::feature` is derived from `featureMembership`, this means that a type may get features that are not owned features of any of its supertypes, which may have owning types incompatible with the inheriting type.

**Overview**

The proposed change to the OCL is to simply further filter the inherited feature memberships with `select(mem | self.specializes(mem.owningType))`. The implementation in `Type_featureMembership_SettingDelegate` could be similarly updated with such a filter. However, doing this naively has a notable negative effect on performance. This is because `specializes` is implemented by computing the `allSupertypes` transitive closure and then checking against that. And the set of `allSupertypes` gets recomputed on each call to `specializes`.

The performance problem can be mitigated by computing `allSupertypes` once, outside the filtering loop, and then using `allSupertypes.contains` rather than `this.specializes`. With this change (and caching), the performance is essentially unaffected by the change.

**Changes**

* `Type_featureMembership_SettingDelegate` – Revised `basicGet`to simply call `TypeUtil.getFeatureMembershipOf`.
* `TypeUtil` – Added `getFeatureMembershipOf` to provide access to `TypeAdapter.getFeatureMembership`.
* `TypeAdapter` – Implemented `getFeatureMembership` as discussed above. Implementing this in the adapter allows the computed `featureMembership` to be cached (alongside `inheritedMembership`).

### DIFF
--- a/org.omg.sysml.interactive.tests/src/org/omg/sysml/interactive/tests/DerivedPropertyAndOperationTest.java
+++ b/org.omg.sysml.interactive.tests/src/org/omg/sysml/interactive/tests/DerivedPropertyAndOperationTest.java
@@ -43,11 +43,13 @@ import org.omg.sysml.lang.sysml.EnumerationDefinition;
 import org.omg.sysml.lang.sysml.EnumerationUsage;
 import org.omg.sysml.lang.sysml.Expression;
 import org.omg.sysml.lang.sysml.Feature;
+import org.omg.sysml.lang.sysml.FeatureMembership;
 import org.omg.sysml.lang.sysml.ItemUsage;
 import org.omg.sysml.lang.sysml.Membership;
 import org.omg.sysml.lang.sysml.Namespace;
 import org.omg.sysml.lang.sysml.Relationship;
 import org.omg.sysml.lang.sysml.TriggerInvocationExpression;
+import org.omg.sysml.lang.sysml.Type;
 import org.omg.sysml.lang.sysml.Usage;
 import org.omg.sysml.lang.sysml.ViewUsage;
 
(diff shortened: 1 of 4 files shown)


himi:
It tool much time to understand what the critical issue, but now I think I can understand.  We need to properly filter `featureMembership` to keep type compatibility.  And I did not notice import memberships are inherited!  So I tested it with:
```
package Test {
    part def A {
        part f;
    }
    part def Bprivate {
        private import A::*;
        part g;
    }
    part def Bpublic {
        public import A::*;
        part g;
    }
    part def Bprotected {
        protected import A::*;
        part g;
    }
    part def Cprivate :> Bprivate;
    part def Cpublic :> Bpublic;
    part def Cprotected :> Bprotected;
}
```

And get owningType of thier featureMembership with the following code:
```
from mgpy.mg import iMg
def showFeatureMembership(q):
    C = iMg.dq(q)[0]
    print(f'{C.qualifiedName}')
    for ms in C.featureMembership:
        e = ms.memberElement
        if not e.isStandardLibrary:
            print(f'  {e.qualifiedName} owningType: {ms.owningType.qualifiedName}')

showFeatureMembership('Test.A')
showFeatureMembership('Test.Bpublic')
showFeatureMembership('Test.Bprivate')
showFeatureMembership('Test.Bprotected')
showFeatureMembership('Test.Cprivate')
showFeatureMembership('Test.Cpublic')
showFeatureMembership('Test.Cprotected')
```

And then I obtained:
```
Test::A
  Test::A::f owningType: Test::A
Test::Bpublic
  Test::Bpublic::g owningType: Test::Bpublic
Test::Bprivate
  Test::Bprivate::g owningType: Test::Bprivate
Test::Bprotected
  Test::Bprotected::g owningType: Test::Bprotected
Test::Cprivate
  Test::Bprivate::g owningType: Test::Bprivate
Test::Cpublic
  Test::Bpublic::g owningType: Test::Bpublic
Test::Cprotected
  Test::Bprotected::g owningType: Test::Bprotected
```
They are correct!

himi:
Just FYI, before this fix, I got:
```
Test::A
  Test::A::f owningType: Test::A
Test::Bpublic
  Test::Bpublic::g owningType: Test::Bpublic
Test::Bprivate
  Test::Bprivate::g owningType: Test::Bprivate
Test::Bprotected
  Test::Bprotected::g owningType: Test::Bprotected
Test::Cprivate
  Test::Bprivate::g owningType: Test::Bprivate
Test::Cpublic
  Test::Bpublic::g owningType: Test::Bpublic
  Test::A::f owningType: Test::A
Test::Cprotected
  Test::Bprotected::g owningType: Test::Bprotected
  Test::A::f owningType: Test::A
```

himi (approved):

PR #750, opened 2026-04-27
ST6RI-916 org.omg.sysml.model module

AxelRICHARD:
**Motivation**

This change decouples the generated SysML EMF metamodel from the `org.omg.sysml` Eclipse plug-in.

The generated model is now produced and published as a plain Java/Maven artifact, while Eclipse/Tycho consumers continue to access it through a dedicated OSGi wrapper bundle. This makes the model easier to reuse outside Eclipse, clarifies module ownership, and removes some unnecessary runtime dependencies from the generated code.

**What Changed**

  - Added `org.omg.sysml.model` as a new plain Java module containing:
      - the generated SysML EMF Java classes
      - `SysML.ecore`
      - `SysML.genmodel`
  - Added `org.omg.sysml.model.bundle` as a new Eclipse plug-in that:
      - wraps the `org.omg.sysml.model` JAR
      - packages the model resources for OSGi/Tycho consumption
      - exports the generated SysML EMF packages
  - Updated the root Maven reactor to build both new modules.
  - Removed generated model sources from `org.omg.sysml/syntax-gen`; `org.omg.sysml` no longer owns compilation of the generated EMF model.
  - Rewired dependent plug-ins to consume `org.omg.sysml.model.bundle`, including:
      - `org.omg.sysml`
      - `org.omg.sysml.edit`
      - `org.omg.sysml.execution`
      - `org.omg.sysml.xtext`
      - `org.omg.kerml.xtext`
  - Added `org.omg.sysml.model.bundle` to the Eclipse feature so it is shipped with the existing tooling.

**Model Generation Changes**

  - Updated the custom UML-to-Ecore importer to generate into:
      - plugin ID: `org.omg.sysml.model`
      - source directory: `/org.omg.sysml.model/src/main/java`
  - Updated the SysML namespace URI used by generation and QVTO references from:
      - `https://www.omg.org/spec/SysML/20230201`
      - to `https://www.omg.org/spec/SysML/20250201`
  - Added deterministic sorting of imported Ecore contents:
      - `EClassifiers`
      - `EStructuralFeatures`
      - `EOperations`

  This reduces noisy regeneration diffs and stabilizes generated artifacts.

**Primitive Type Simplification**

  - Normalized imported UML primitive types to built-in Ecore datatypes during import.
  - Removed types.ecore.
  - Removed the generated org.omg.sysml.lang.types package and related generated classes.

  This simplifies the metamodel structure and removes an extra generated dependency layer.

**Runtime Dependency Cleanup**

  The generated model previously depended on UML2 helper-list implementations from org.eclipse.uml2.common.

  This commit copies the required helper classes into the SysML model module and updates generation so the generated implementation classes import the local versions instead:

  - DerivedEObjectEList
  - DerivedUnionEObjectEList
  - UnionEObjectEList

 The org.omg.sysml.uml.ecore.importer has been updated to rewrite those imports automatically after generation.

**Impact**

  - The SysML metamodel is now reusable as a regular Java library outside Eclipse.
  - Eclipse/Tycho-based modules remain compatible through org.omg.sysml.model.bundle.
  - Ownership of generated EMF code is now explicit and isolated in org.omg.sysml.model.
  - Generated sources are cleaner and less coupled to UML2 runtime internals.
  - Most of the file churn in this commit is due to regeneration and relocation of generated EMF sources, not hand-written functional logic changes.


seidewitz:
As originally proposed, this refactoring worked in Jupyter and the Xpect tests, but model files failed to open in the Xtext IDE, ultimately due to the inability to load the EObject class. I have now pushed a commit that resolves this problem by adding dependencies on the `org.eclipse.emf.common` and `org.eclipse.emf.ecore` plugins.

AxelRICHARD:
@seidewitz great thank you!
For those who have compiling issues inside Eclipse, you have to do a Maven Update (inside eclipse with M2E) on /org.omg.sysml.model.bundle and then refresh plugins the workspace.

himi:
This PR radically changed the project structure and dependencies. 
* M2E is now mandatory to resolve dependencies.
* org.eclipse.emf.ant is a new dependent of org.omg.sysml.model.bundle.  Can you check whether or not the dependency of org.eclipse.emf.ant is really needed?  It is used to generate code from EMF model. I tried to remove the dependency and it seems to work without it.
* mvn install is critical to update org.omg.sysml.model bundle
* I needed to reset the target environment to resolve org.apache.commons.logging (I have no idea why it's needed)

I could make it work in my environment by fixing the issues.

seidewitz:
@himi 
> * org.eclipse.emf.ant is a new dependent of org.omg.sysml.model.bundle.  Can you check whether or not the dependency of org.eclipse.emf.ant is really needed?  It is used to generate code from EMF model. I tried to remove the dependency and it seems to work without it.

The dependency on `org.eclipse.emf.ant` happened when I (not @AxelRICHARD) added the `org.eclipse.emf.common` and `org.eclipse.emf.ecore` dependencies to `org.omg.sysml.model.bundle`. I only added the latter two explicitly, and the `ant` dependency was added automatically. But we aren't using it in the bundle, so I think it is OK to delete it.

himi:
Thank you for the comment.  So may I remove it?

seidewitz:
@himi
> Thank you for the comment. So may I remove it?

Sure, if you like.

I wasn't sure from you message of 2 days ago if there still any changes you would like @AxelRICHARD to make. What is the status of updating the version number updates (other than plugin version)?

himi (approved):
I removed org.eclipse.emf.ant dependency in org.omg.sysml.model.bundle and it worked correctly.

himi:
@seidewitz 
> What is the status of updating the version number updates (other than plugin version)?
I'm not so sure what you are mentioning about but is it related to the jar in org.omg.sysml.model.bundle/lib?

AxelRICHARD:
@seidewitz @himi I still have to update the pom.xml to handle the version number updates (other than plugin version).
I will add a new commit on this PR asap and let you know when it will be done.
Thank you

AxelRICHARD:
@seidewitz @himi The update is ok.

AxelRICHARD:
@seidewitz @himi It is ok, you can do a last review, thank you.

seidewitz:
@AxelRICHARD 
I can confirm that the version update is working. However, I am still having problems cleanly building after changing the version number in the parent POM. Here is what I am doing:

1. Change version number in parent POM.
2. Eclipse clean and build all projects in repository.
3. Maven clean repository.
4. Maven install repository.
5. Maven update `org.omg.sysml.model.bundle`.

However, the first time I do step 4 (Maven install), the build fails because the `org.omg.sysml.model-0.60.0-SNAPSHOT.jar` is (oddly) empty. If I rerun the install, it then completes successfully.

Can you reproduce this behavior? Can you fix it? Or do I need to do something different?

AxelRICHARD:
@seidewitz Could you please try to:
- change step 5 from Maven update org.omg.sysml.model.bundle to Maven update org.omg.sysml.model.bundle AND org.omg.sysml.model 
- and move step2 after step5?

seidewitz:
> @seidewitz Could you please try to:
> 
> * change step 5 from Maven update org.omg.sysml.model.bundle to Maven update org.omg.sysml.model.bundle AND org.omg.sysml.model
> * and move step2 after step5?

It doesn't solve the problem. The Maven install produces and exception when building `org.omg.sysml`.
```
Failed to init Classpath for jar file /Users/seidewitz/Documents/Work/git/SysML-v2-Pilot-Implementation/org.omg.sysml.model.bundle/lib/org.omg.sysml.model-0.60.0-SNAPSHOT.jar
java.util.zip.ZipException: zip file is empty
```
And now this doesn't go away, because the workspace is in an inconsistent state when I don't rebuild it in Eclipse before the Maven build.

seidewitz:
I have explored a lot of different things, but the only way I am able to get a build completed without error (when the version changes) is the following:

1. Change version number in parent POM.
2. Eclipse clean and build all projects in repository.
3. Maven clean repository.
4. Maven install (_not_ update) _just_ `org.omg.sysml.model`.
5. Maven update `org.omg.sysml.model.bundle`.
6. Maven install repository.

It almost seems easier to just do the build as before, let it fail, and then succeed on the second try!

On the other hand, as far as I can tell so far, this build problem only happens when moving to a new version, so having a slightly more complicated build procedure is not really that bad. But I worry about what is going to happen as we add more non-plugin modules. And it would still be better anyway if we understood what was causing the problem.

AxelRICHARD:
@seidewitz Here is a procedure that could be more suitable:

  1. Close Eclipse.
  2. Update the revision and then run the root set-version goal from the repository root. (mvn -N antrun:run@set-version)
  3. Run a full Maven build from the root. (mvn clean install)
  4. Reopen Eclipse.
  5. Inside Eclipse, run Maven > Update Project... on all Maven projects.
  6. If Eclipse still shows stale PDE/JDT issues, run Project > Clean....

  Why this order fits this repo:

  - The root set-version step does not only change Maven-facing versions. It also rewrites committed Eclipse/PDE metadata:
      - MANIFEST.MF
      - feature.xml
      - build.properties
      - .classpath
      - versioned embedded JAR paths for org.omg.sysml.model.bundle
  - Because of that, keeping Eclipse open during set-version is more likely to leave the workspace in a half-stale state.
  - The full Maven build matters before reopening Eclipse because the wrapper bundles expect the versioned embedded JARs to exist in lib/ consistently with the rewritten metadata.

seidewitz:
> Here is a procedure that could be more suitable:
> 
> 1. Close Eclipse.
> 2. Update the revision and then run the root set-version goal from the repository root. (mvn -N antrun:run@set-version)
>3. Run a full Maven build from the root. (mvn clean install)
> 4. Reopen Eclipse.
> 5. Inside Eclipse, run Maven > Update Project... on all Maven projects.
> 6. If Eclipse still shows stale PDE/JDT issues, run Project > Clean....

I am afraid I don't really find it acceptable to have to leave Eclipse to carry out a build. Especially since I do have a procedure that works entirely inside Eclipse. I was hoping that there was something simpler, but what you are proposing is more complicated.

Perhaps @himi can comment. He implemented the set-version mechanism.

seidewitz:
I have now established the the following works, entirely within Eclipse:

1. Change version number in parent POM.
2. Maven build `antrun:run@set-version`.
3. Refresh the repository
4. Eclipse clean and build all projects in repository.
5. Maven clean repository.
6. Maven install repository.
7. Maven update `org.omg.sysml.model.bundle`.

Not sure it is much simpler...

seidewitz (approved):
I am a still concerned with the difficulties in the build process with the new architecture. But the build seems manageable for now, so I am going to approve and merge this PR so we can move on. We can consider further improvements in the build process in conjunction with future PRs.

But I would also still be interested a deeper review by @himi of the updates to the build process.

PR #752, opened 2026-05-11
ST6RI-922 Change licensing to the EPL

seidewitz:
All copyright holders have agreed to change the licensing for the Pilot Implementation from the Lesser General Public License (LGPL) to the more permissive Eclipse Public License (EPL). This PR makes this change for the [SysML v2 Pilot Implementation](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation) repository. The change will be made separately for the [SysML v2 API and Services](https://github.com/Systems-Modeling/SysML-v2-API-Services) and [SysML v2 Release](https://github.com/Systems-Modeling/SysML-v2-Release) repositories.

**Changes**

1. Replaced `LICENSE` and `LICENSE-GPL` files with a `LICENSE` file containing the EPL 2.0.
2. Revised the portion of the repository `README` related to licensing and copyright.
3. Updated (by global replace of specific phrases) all existing file copyright headers that referenced the LGPL.


AxelRICHARD:
@seidewitz do you also plan to add/update copyrights to the files in org.omg.sysml.model module?

seidewitz:
> @seidewitz do you also plan to add/update copyrights to the files in org.omg.sysml.model module?

I reviewed`org.omg.sysml.module`, and it it is almost entirely generated code, with only a very little handwritten (e.g., to handle default redefinition). We have generally not put license header comments on this code, so there was nothing for the the global replace to replace. However, I have a utility that would make it easy to put the headers on generated code if we want to. From your point of view, would it be better to have the licensing header on, e.g., generated EMF Java classes?

himi (approved):
Thank you!  It looks good to me.

AxelRICHARD:
> > @seidewitz do you also plan to add/update copyrights to the files in org.omg.sysml.model module?
> 
> I reviewed`org.omg.sysml.module`, and it it is almost entirely generated code, with only a very little handwritten (e.g., to handle default redefinition). We have generally not put license header comments on this code, so there was nothing for the the global replace to replace. However, I have a utility that would make it easy to put the headers on generated code if we want to. From your point of view, would it be better to have the licensing header on, e.g., generated EMF Java classes?

@seidewitz No, I don't have strong opinion about that. I just wanted to be sure it was not an omission. So it's ok on my side too, thank you.

AxelRICHARD (approved):

PR #768, opened 2026-06-16
ST6RI-937 Variants are incorrectly referential

### LINKED ISSUE #763: Issue about isComposite of variant usage
Hello,
I'm tried to upgrade the Pilot from 2025-10 to 2026-02 and found the isComposite value of variant usage is changed. Such as this SysML example:
```
part part0 {
	variation part part2 {
		variant part part21;
		variant ref part part22;
	}
}
```
The isComposite of part21 is changed from true to false.
After look through the code, seems the value of isComposite is calculated in the postProcess of UsageAdapter:
```
if (target.getDirection() != null || target.isEnd() || 
	// Note: A parsed Usage can only get a featuring type if it is owned via a FeatureMembership.
	!(target.eContainer() instanceof FeatureMembership)) {
	target.setIsComposite(false);
}
```
However, the Usage owned via VariantMembership can also have featuring type, such as the part21 and part22 in the previous SysML example.
Is this difference designed intentionally, or is truly an issue?

seidewitz:
This PR corrects the computation of an effective `featuringType` of a `Usage` in the `postProcess` method for `UsageAdapter`.

**Background**

Previously, the `UsageAdapter::postProcess `methods contained the code:
```
if (target.getDirection() != null || target.isEnd() || 
	// Note: A parsed Usage can only get a featuring type if it is owned via a FeatureMembership.
	!(target.eContainer() instanceof FeatureMembership)) {
	target.setIsComposite(false);
}
```
The comment note in this code is incorrect. A variant usage is owned by a `VariantMembership`, which is not a `FeatureMembership,` but, per semantic constraint `checkUsageVariationUsageTypeFeaturing`, has the same featuring types as its owning variation, which can be non-empty.

Note that the reason `getFeaturingType` isn’t just called here is to avoid computing derived values during parse post-processing.

**Changes**

In project `org.omg.sysml.logic`:

1. `UsageUtil` – Added:
    - `getExpectedFeaturingType` – Gets the expected `featuringType` for a `Usage`, assuming limitations in the SysML syntax. This is computed recursively as the `owningType` of the nearest non-variation container if this exists. Note that the special `featuringTypes` for variable features are skipped in this computation (not necessary for how this utility is used in postprocessing).
    -  `hasFeaturingType` – Whether `getExpectedFeaturingType` is non-null, meaning the `Usage` will have a `FeaturingType`.
2. `UsageAdapter` – Revised the method `postProcess` to use `UsageUtil.hasFeaturingType` in the condition on whether to set `isComposite` to false.
3. `PortAdapter` – Revised the method `postProcess` to use `UsageUtil.getExpectedFeaturingType`. This revision was necessary because the change to `UsageAdapter` resulted in variant `PortUsages` being considered composite, violating the constraint `validatePortUsageIsReference`.

**Test**

- `org.omg.sysml.logic.UsagePostProcessTest`

### DIFF
--- a/org.omg.sysml.logic/src/main/java/org/omg/sysml/util/UsageUtil.java
+++ b/org.omg.sysml.logic/src/main/java/org/omg/sysml/util/UsageUtil.java
@@ -26,6 +26,7 @@ import java.util.stream.Stream;
 
 import org.eclipse.emf.common.util.BasicEList;
 import org.eclipse.emf.common.util.EList;
+import org.eclipse.emf.ecore.EObject;
 import org.omg.sysml.adapter.UsageAdapter;
 import org.omg.sysml.lang.sysml.AcceptActionUsage;
 import org.omg.sysml.lang.sysml.ActionUsage;
@@ -90,6 +91,23 @@ public class UsageUtil {
 		return getUsageAdapter(usage).mayTimeVary();
 	}
 		
+	// Featuring Types
+	
+	/**
+	 * Determine with a given Usage should have a featuringType after transformation,
+	 * without actually computing the full derivation for featuringType. Assumes that
+	 * a Usage in SysML can only get a featuringType in one of two ways:
+	 * 1. If it is an ownedFeature.
+	 * 2. If it is a variant of a variation Usage that has a featuringType.
+	 */
+	public static boolean hasFeaturingType(Usage usage) {
+		EObject container = usage.eContainer();
+		return container instanceof FeatureMembership ||
+			   container.eContainer() instanceof Usage &&
+			   ((Usage)container.eContainer()).isVariation() &&
+			   hasFeaturingType((Usage)container.eContainer());
+	}
+
 	// Variants
 	
 	public static boolean isVariant(Usage usage) {
--- a/org.omg.sysml.logic/src/main/java/org/omg/sysml/util/UsageUtil.java
+++ b/org.omg.sysml.logic/src/main/java/org/omg/sysml/util/UsageUtil.java
@@ -94,18 +94,28 @@ public class UsageUtil {
 	// Featuring Types
 	
 	/**
-	 * Determine with a given Usage should have a featuringType after transformation,
+	 * Determine the featuringType a Usage is expected to have after transformation,
 	 * without actually computing the full derivation for featuringType. Assumes that
 	 * a Usage in SysML can only get a featuringType in one of two ways:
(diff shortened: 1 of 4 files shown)


org.omg.sysml.logic/src/main/java/org/omg/sysml/util/UsageUtil.java
```diff
@@ -90,6 +91,33 @@ public static boolean mayTimeVary(Usage usage) {
...
+	 * Determine the featuringType a Usage is expected to have after transformation,
+	 * without actually computing the full derivation for featuringType. Assumes that
+	 * a Usage in SysML can only get a featuringType in one of two ways:
+	 * 1. If it is an ownedFeature.
+	 * 2. If it is a variant of a variation Usage that has a featuringType.
+	 * Note that the special featuringTypes of variable features are ignored.
+	 */
+	public static Type getExpectedFeaturingTypeOf(Usage usage) {
+		EObject container = usage.eContainer();
+		if (container instanceof FeatureMembership) {
```
AxelRICHARD:
```
if (container instanceof FeatureMembership featureMemebership) {
    return featureMemebership.getOwningType();
}
```

org.omg.sysml.logic/src/main/java/org/omg/sysml/util/UsageUtil.java
```diff
@@ -90,6 +91,33 @@ public static boolean mayTimeVary(Usage usage) {
...
+			return ((FeatureMembership)container).getOwningType();
+		} else if (container != null &&
```
AxelRICHARD, line 108:
```
} else if (container != null &&
           container.eContainer() instanceof Usage usageContainer &&
           usageContainer.isVariation()) {
     return getExpectedFeaturingTypeOf(usageContainer.eContainer());
}
```

org.omg.sysml.logic/src/main/java/org/omg/sysml/util/UsageUtil.java
seidewitz: Revised as proposed.

org.omg.sysml.logic/src/main/java/org/omg/sysml/util/UsageUtil.java
seidewitz: Revised essentially as proposed.

AxelRICHARD:
@seidewitz I don't have any other comments on this pull request. You can merge it.

PR #770, opened 2026-06-17
ST6RI-939 Fix StackOverflow when loading a ResourceSet

AxelRICHARD:
Fix a `StackOverflowError` when loading a `ResourceSet` containing `TransitionUsage` elements.

### Context

Loading a SysML-XMI `ResourceSet` can trigger a `StackOverflowError` when derived properties are evaluated on `TransitionUsage` elements.

The recursion involves the derived transition properties and the implicit member computation performed by `TransitionUsageAdapter`, especially around:

- `TransitionUsage::source`
- `TransitionUsage::sourceFeature()`
- `TransitionUsage::succession`
- Implicit transition members and binding connectors

### Observed Behavior

  The application fails with a recursive stack similar to:

```
java.lang.StackOverflowError
  at org.omg.sysml.delegate.setting.TransitionUsage_source_SettingDelegate.basicGet(...)
  at org.omg.sysml.lang.sysml.impl.TransitionUsageImpl.getSource(...)
  at org.omg.sysml.adapter.TransitionUsageAdapter.computeTransitionLinkConnectors(...)
  at org.omg.sysml.adapter.TransitionUsageAdapter.addAdditionalMembers(...)
  at org.omg.sysml.util.NamespaceUtil.addAdditionalMembersTo(...)
  at org.omg.sysml.util.UsageUtil.getSourceFeatureOf(...)
  at org.omg.sysml.lang.sysml.impl.TransitionUsageImpl.sourceFeature(...)
  at org.omg.sysml.delegate.setting.TransitionUsage_source_SettingDelegate.basicGet(...)
  ...
```

### SysML-Level Explanation

In SysML, `TransitionUsage::source` is a derived property. It is computed from `sourceFeature()`, which identifies the feature used as the source of the transition succession.

At the same time, `TransitionUsageAdapter` may add implicit members required by transition semantics, such as:

- the implicit source member;
- transition link connectors;
- binding connectors between the succession, the transition link, and parameters.

The issue is that this implicit member computation is not reentrant-safe.

### Recursion Cycle

The problematic cycle is:

1. Code reads `transition.getSource()`.
2. This evaluates the derived property `TransitionUsage::source`.
3. The derived property calls `transition.sourceFeature()`.
4. `sourceFeature()` delegates to `UsageUtil.getSourceFeatureOf(transition)`.
5. `UsageUtil.getSourceFeatureOf(...)` calls `NamespaceUtil.addAdditionalMembersTo(transition)`.
6. `TransitionUsageAdapter.addAdditionalMembers()` calls `computeTransitionLinkConnectors()`.
7. `computeTransitionLinkConnectors()` reads `transition.getSource()`.
8. The same cycle starts again.

This leads to unbounded recursion and eventually a `StackOverflowError`.

### Typical Trigger

The issue can occur when a `TransitionUsage` has parameters, but its succession-related implicit members are not fully available at the time source is evaluated.

In that state, `computeTransitionLinkConnectors()` attempts to compute source-related binding connectors and reads `transition.getSource()`, which re-enters the same implicit member computation.

### Changes

Added a Boolean `isComputingTransitionLinkConnectors` that is set to true during the computation of transition link connectors and reset to false when the computation is done. However, if `computeTransitionLinkConnectors` ends up being recursively called, it returns without calling `TransitionUsage.getSource()`.

### DIFF
--- a/org.omg.sysml.logic/src/main/java/org/omg/sysml/adapter/TransitionUsageAdapter.java
+++ b/org.omg.sysml.logic/src/main/java/org/omg/sysml/adapter/TransitionUsageAdapter.java
@@ -38,6 +38,8 @@ import org.omg.sysml.util.UsageUtil;
 
 public class TransitionUsageAdapter extends ActionUsageAdapter {
 
+	private boolean isComputingTransitionLinkConnectors = false;
+
 	public TransitionUsageAdapter(TransitionUsage element) {
 		super(element);
 	}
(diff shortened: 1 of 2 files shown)


AxelRICHARD:

seidewitz (commented):
Really, this computation needs to be revised so the adding of additional members can be eliminated. For now, though, the simple change proposed in this PR is sufficient.

seidewitz (approved):

PR #771, opened 2026-06-17
ST6RI-940 ResourceSetModelLibraryProvider Implementation does not take into account the root namespace

AxelRICHARD:
This PR fixes the `ResourceSetModelLibraryProvider` Implementation to take into account the root namespace.

**Background**

The standalone `ResourceSetModelLibraryProvider` did not correctly resolve qualified names for elements stored in standard SysML/KerML model library resources when those resources followed the structure required by the SysML/KerML specification. The previous implementation attempted to resolve a qualified name such as `Base::Anything` by matching the first segment (`Base`) directly against each root object contained in an EMF Resource. This worked only if the named library package itself was the direct root object of the resource. However, this is not the structure defined by the specification for model interchange and model libraries.

According to KerML and SysML, a model interchange file represents a single root namespace. This root namespace is implicit, unnamed, and has no owner. The actual model elements, including top-level library packages such as Base, are owned members of that root namespace.

Therefore, in a conforming resource, the structure is:
```
Resource
  Root Namespace              // implicit, unnamed
    OwningMembership
      Library Package "Base"  // top-level element
        OwningMembership
          Type "Anything"
```
The old lookup logic expected this instead:
```
Resource
  Library Package "Base"
    Type "Anything"
```
As a result, resolving `Base::Anything` failed for correctly structured resources, because the provider compared `Base` with the unnamed root namespace and stopped before inspecting the root namespace memberships.

Instead, `Base::Anything` must be resolved by entering the resource root namespace, then matching `Base` among its owned memberships, then matching `Anything` inside the `Base` namespace.

**Changes**

The model library provider now treats the resource root objects as root namespaces, not as named elements to be matched against the first qualified name segment.

The lookup algorithm:

1. Iterates over resource contents.
2. Considers only root objects that are Namespace instances.
3. Resolves the first qualified-name segment against the namespace's owned memberships.
4. Continues recursively through member elements that are themselves namespaces.
5. Matches each segment against membership names, membership short names, declared names, and declared short names, preserving alias and short-name behavior.
6. Ignore non-namespace root elements for qualified-name lookup, because an element without an owning namespace has no valid qualified name according to the metamodel rules.

### DIFF
--- a/org.omg.sysml.logic/src/test/java/org/omg/sysml/logic/SysMLLogicStandaloneSetupTest.java
+++ b/org.omg.sysml.logic/src/test/java/org/omg/sysml/logic/SysMLLogicStandaloneSetupTest.java
@@ -29,6 +29,7 @@ import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.resource.impl.ResourceImpl;
 import org.eclipse.emf.ecore.resource.impl.ResourceSetImpl;
 import org.junit.Test;
+import org.omg.sysml.lang.sysml.Element;
 import org.omg.sysml.lang.sysml.Namespace;
 import org.omg.sysml.lang.sysml.OwningMembership;
 import org.omg.sysml.lang.sysml.SysMLFactory;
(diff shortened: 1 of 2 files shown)


org.omg.sysml.logic/src/test/java/org/omg/sysml/logic/SysMLLogicStandaloneSetupTest.java
```diff
@@ -83,4 +87,10 @@ public void standaloneSetupResolvesLibraryElementsWithoutXtext() {
 		SysMLLogicStandaloneSetup.doSetup();
 		assertSame(anything, SysMLLibraryUtil.getLibraryType(context, "Base::Anything"));
 	}
+
+	private void addAsOwnedMember(Element parent, Element child) {
+		OwningMembership owningMembership = SysMLFactory.eINSTANCE.createOwningMembership();
```
seidewitz: There is already a utility method for his: `NamespaceUtil.addOwnedMemberTo(Namespace, Element)`.

AxelRICHARD: Comment has been taken into account.

PR #772, opened 2026-06-17
ST6RI-941 NPE when evaluating derived properties on feature chains

AxelRICHARD:
Evaluating some SysML derived properties can fail with a _NullPointerException_ when the model contains an unresolved feature chain.

The problem occurs in `TypeUtil.getOwnedParametersOf(Type type)`. When the provided type is a `Feature`, the method first resolves its basic feature through `FeatureUtil.getBasicFeatureOf(...)`. For a chained feature whose target has not been resolved yet, this call may return null. The method then unreferences that _null_ value by calling
`type.getOwnedFeature()`.

A concrete failing scenario is an `AssignmentActionUsage` containing an unresolved `FeatureChaining`. When its derived _valueExpression_ property is evaluated, the computation asks `TypeUtil`
for owned parameters and triggers the NPE.

Expected behavior: unresolved feature chains should be treated as having no owned parameters at this stage, so derived property evaluation should return null or an empty result instead
of crashing.

This issue prevents robust handling of partially linked SysML models and can break workflows that evaluate derived properties before all feature chain references are resolved.

### DIFF
--- a/org.omg.sysml.logic/src/main/java/org/omg/sysml/util/TypeUtil.java
+++ b/org.omg.sysml.logic/src/main/java/org/omg/sysml/util/TypeUtil.java
@@ -229,12 +229,10 @@ public class TypeUtil {
 		if (type instanceof Feature) {
 			type = FeatureUtil.getBasicFeatureOf((Feature)type);
 		}
-		if (type == null) {
-			return new ArrayList<>();
-		}
-		return type.getOwnedFeature().stream().
-				filter(FeatureUtil::isParameter).
-				collect(Collectors.toList());
+		return type == null? Collections.emptyList():
+			    type.getOwnedFeature().stream().
+					filter(FeatureUtil::isParameter).
+					collect(Collectors.toList());
 	}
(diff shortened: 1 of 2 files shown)


AxelRICHARD:

seidewitz:
@AxelRICHARD 
I revised the code for `TypeUtil::getOwnedParametersOf` to have a style consistent with other similar methods (e.g., `getEndFeaturesOf` and `getOwnedFeaturesOf`. The functionality is the small (Junit test passes).

Please confirm you have no problem with this change.

AxelRICHARD:
> @AxelRICHARD I revised the code for `TypeUtil::getOwnedParametersOf` to have a style consistent with other similar methods (e.g., `getEndFeaturesOf` and `getOwnedFeaturesOf`. The functionality is the small (Junit test passes).
> 
> Please confirm you have no problem with this change.

I confirm have no problem with this change, thank you, you can merge the PR.

seidewitz (approved):

PR #773, opened 2026-06-22
ST6RI-944 Invalidate TypeAdapter#getFeatureMembership cache when needed

AxelRICHARD:
This PR revises the implementation for `getFeatureMembership` on a `Type` element so that, if additional `FeatureMemberships` are added, they are reflected in subsequent calls to `getFeatureMembership`.

**Background**

PR #749 (ST6RI-925) introduced caching in `TypeAdapter` for `featureMemberships`. This caching happens on the first call to `getFeatureMembership`. After this, unless the cache is cleared, `getFeatureMemberships` always returns the cached value, even if new `FeatureMembership` relationships have been added to the source element.

The following methods directly or indirectly use `getFeatureMembership` and,so, also reflect this behavior:
- `Type.getFeature`, `getDirectedFeature`, `getEndFeature`, `getInput`, `getOutput`
- `InstantiationExpression.getArgument`
- `ConnectionDefinition.getConnectionEnd`
- `InterfaceDefinition.getInterfaceEnd`

**Note.** `Association.getAssociationEnd`, `Connector.getConnectorEnd` and `FlowDefinition.getFlowEnd` and `FlowUsage.getConnectorEnd` have the same functionality as `getEndFeature`.

**Changes**

For each of the methods above, define a corresponding static utility method. The original property methods are re-implemented to first call `ElementUtil.clearCache` on the method argument, then call the corresponding utility method. Internally, the utility methods are used directly to allow caching to improve performance.

| Property Getter Method            | Utility Method                                       |
| -------------------------------|---------------------------------------|
| `Type.getFeatureMembership` | `TypeUtil.getFeatureMembershipOf ` |
| `Type.getFeature`                     | `TypeUtil.getFeatureOf`                       |
| `Type.getDirectedFeature`       | `TypeUtil.getDirectedFeatureOf`         |
| `Type.getEndFeature`               | `TypeUtil.getEndFeatureOf`                 |
| `Type.getInput`                         | `TypeUtil.getInputOf`                            |
| `Type.getOutput`                      | `TypeUtil.getOutputOf`                         |
| `InstantiationExpression.getArgument`      | `ExpressionUtil.getArgumentOf`    |
| `ConnectionDefinition.getConnectionEnd` | `UsageUtil.getConnectionEndOf` |
| `InterfaceDefinition.getInterfaceEnd`         | `UsageUtil.getInterfaceEndOf`      |

With this approach, performance (as measured using `SysMLInteractiveProfiler`) is essentially the same as before the changes in this PR. Without caching, there is a performance degradation of 10 to 20%.


AxelRICHARD:

org.omg.sysml.logic/src/test/java/org/omg/sysml/logic/TypeFeatureMembershipCacheTest.java
```diff
@@ -0,0 +1,103 @@
+/*******************************************************************************
+ * SysML 2 Pilot Implementation
```
AxelRICHARD, line 2: Sorry I put the LGPL license here, but we need EPL license

org.omg.sysml.logic/src/main/java/org/omg/sysml/util/ExpressionUtil.java
```diff
@@ -191,4 +194,28 @@ public static EList<Expression> getOperandsOf(InvocationExpression expression) {
...
+			}
+		}
+		return arguments;
+	}
+	
+	private static void addArguments(List<Expression> arguments, List<Feature> instantiatedTypeFeatures, List<Feature> argumentFeatures) {
+		instantiatedTypeFeatures.stream().
+			flatMap(f->argumentFeatures.stream().filter(rf->rf.redefines(f))).
+			map(FeatureUtil::getValueExpressionFor).
+			filter(e->e != null).
```
AxelRICHARD: `filter(e->e != null)` => `filter(Objects::nonNull)` as an alternative

org.omg.sysml.logic/src/main/java/org/omg/sysml/util/TypeUtil.java
```diff
@@ -308,6 +332,26 @@ private static Set<ResultExpressionMembership> getResultExpressionMembershipsOf(
...
 
+	public static EList<Feature> getInputOf(Type type) {
+		EList<Feature> inputs = new NonNotifyingEObjectEList<>(Feature.class, (InternalEObject)type, SysMLPackage.TYPE__INPUT);
+		TypeUtil.getFeatureOf(type).stream().filter(f->FeatureUtil.isInputParameter(f, type)).forEachOrdered(inputs::add);
+		return inputs;
+	}
+
+	public static EList<Feature> getOutputOf(Type type) {
+		EList<Feature> inputs = new NonNotifyingEObjectEList<>(Feature.class, (InternalEObject)type, SysMLPackage.TYPE__INPUT);
+		TypeUtil.getFeatureOf(type).stream().filter(f->FeatureUtil.isInputParameter(f, type)).forEachOrdered(inputs::add);
```
AxelRICHARD: - SysMLPackage.TYPE__INPUT => SysMLPackage.TYPE__OUTPUT
- isInputParameter => isOutputParameter

If the build succeeds with that implementation errors in `getOutputOf`, it probably means that we need a dedicated test for that, to ensure non-regression.

org.omg.sysml.logic/src/main/java/org/omg/sysml/util/TypeUtil.java
```diff
@@ -205,6 +202,19 @@ public static boolean isCompatible(Type subtype, Type supertype) {
...
 	
+	public static EList<FeatureMembership> getFeatureMembershipOf(Type type) {
+		return getTypeAdapter(type).getFeatureMembership();
+	}
+	
+	public static EList<Feature> getFeatureOf(Type type) {
+		EList<Feature> features = new NonNotifyingEObjectEList<>(Feature.class, (InternalEObject)type, SysMLPackage.TYPE__FEATURE);
+		getFeatureMembershipOf(type).stream().
+			map(FeatureMembership::getOwnedMemberFeature).
+			filter(f->f != null).
```
AxelRICHARD: `filter(e->e != null)` => `filter(Objects::nonNull)` as an alternative

org.omg.sysml.logic/src/test/java/org/omg/sysml/logic/TypeFeatureMembershipCacheTest.java
seidewitz: Fixed.

org.omg.sysml.logic/src/main/java/org/omg/sysml/util/ExpressionUtil.java
seidewitz: Done. (And, thanks, I have been looking for something like this, but didn't know about the `Objects` class!)

org.omg.sysml.logic/src/main/java/org/omg/sysml/util/TypeUtil.java
seidewitz: Corrected.

To be honest, we really don't have much testing on the derived property computation. I have been hoping that the Conformance WG will fill this in eventually. I haven't written tests for any of the new methods I added here.

I would be OK if you added tests for these methods. Or we could just move on for now.

org.omg.sysml.logic/src/test/java/org/omg/sysml/logic/TypeFeatureMembershipCacheTest.java
```diff
@@ -0,0 +1,102 @@
...
+ * Copyright (c) 2026 Obeo
+ *    
+  * This program is free software: you can redistribute it and/or modify
```
AxelRICHARD, line 5: extra space

org.omg.sysml.logic/src/main/java/org/omg/sysml/util/TypeUtil.java
AxelRICHARD: Ok I will add new tests

seidewitz (approved):

PR #754, opened 2026-05-13
ST6RI-926 org.omg.sysml.logic module

AxelRICHARD:
### Summary

This PR extracts model-independent SysML runtime logic from `org.omg.sysml` into new standalone modules, `org.omg.sysml.logic` and `org.omg.sysml.logic.bundle`, and updates dependent projects to consume that logic through explicit bundle and Maven dependencies.

The change also introduces a standalone bootstrap for EMF delegate registration and library resolution, so core SysML logic can run outside Eclipse/Xtext startup, and adds plain-EMF regression tests for that standalone path.

### Language Features

  1. Standalone runtime logic extraction.
     The model-independent SysML runtime layer has been split out of `org.omg.sysml` into a dedicated `org.omg.sysml.logic` module. This includes:
      - invocation delegates
      - setting delegates
      - adapters
      - expression evaluators
      - library function infrastructure
      - shared runtime utilities
  2. Namespace resolution delegates hardened.
     The extracted runtime now carries the delegate implementations for `Namespace::resolve`, `resolveLocal`, `resolveVisible`, and `resolveGlobal`, with runtime behavior aligned to the intended standalone delegation flow.
  3. Standalone delegate bootstrap.
     A new `SysMLLogicStandaloneSetup` class registers the SysML derived-property setting delegates and invocation delegates directly with EMF, allowing runtime use without requiring the Xtext runtime modules to be initialized first.

 ### Model Libraries

  1. Standalone library resolution support.
     The commit introduces `SysMLLibraryUtil` and `IModelLibraryProvider` as runtime-facing abstractions for library lookup, together with a standalone R`esourceSetModelLibraryProvider `implementation that resolves library elements from the EMF `ResourceSet.`
  2. Runtime-module provider wiring.
     KerML and SysML Xtext runtime modules and XMI runtime modules are updated to install the provider lookup used by the shared library utility, so both Eclipse/Xtext and standalone EMF runtimes rely on the same library resolution entry point.

 ### Backward Incompatibilities

  1. Package/module relocation for runtime logic.
     Code that previously referenced runtime logic classes from `org.omg.sysml` must now resolve them from `org.omg.sysml.logic` or from the exported `org.omg.sysml.logic.bundle` plugin.
  2. Removed legacy utility entry points from `org.omg.sysml`.
     The following legacy classes/interfaces are deleted from the old location and replaced by logic-module equivalents:
      - `SysMLLibraryUtil`
      - `IModelLibraryProvider`
      - `ISysMLScope`
[...]

PR #774, opened 2026-06-23
ST6RI-943 Restore ability to install and run without Maven build

seidewitz:
This PR restores the ability to install the Pilot Implementation development environment and run the Xtext editor in a runtime instance, without first doing a full Maven install. (However, a Maven clean install should still be done before changing any code.)

**Background**

PRs #750 (ST6RI-916 in 2026-04) and #754 (ST6RI-926 for 2026-05) refactored code from the `org.omg.sysml` plugin into the non-plugin module projects `org.omg.sysml.model` and `org.omg.sysml.logic`, respectively. After the implementation of these PRs, a Maven install generats JARs for the new projects thatich are used by corresponding bundle projects to provide the modularized functionality to the `org.omg.sysml` plugin. However, these JARs were not being stored in the repository. So, after installing the development environment, a user had to do a Maven build to generate the JARs before the IDE could be run.

**Changes**

1. Updated `org.omg.sysml.model.bundle/.gitignore` and `org.omg.sysml.logic.bundle/.gitignore` so that the corresponding `lib` subdirectories are _not_ ignored.
2. Updated `org.omg.sysml.logic/pom.xml` to use the same approach as `org.omg.sysml.logic.bundle`/pom.xml to copy the JAR file from `org.omg.sysml.logic`. This approach does not produce an error if the source JAR file does not exist.


org.omg.sysml.model.bundle/lib/org.omg.sysml.model-0.59.0-SNAPSHOT.jar
AxelRICHARD, line 1: Why there are two jars in this lib folder? Unless I'm missing something here, there should only be the last one.

org.omg.sysml.logic.bundle/pom.xml
```diff
@@ -28,24 +28,33 @@
 	
 	<build>
     <plugins>
-      <!-- Copy org.omg.sysml.logic jar in lib/ before the Tycho build -->
+      <!-- Copy org.omg.sysml.model jar in lib/ before the Tycho build -->
```
AxelRICHARD: The maven-resources-plugin copies org.omg.sysml.logic so the change of the comment in not necessary

org.omg.sysml.model.bundle/lib/org.omg.sysml.model-0.59.0-SNAPSHOT.jar
seidewitz: Oops. My fault for not cleaning up locally after going back and forth between versions. The 0.59.0 JAR is gone now.

org.omg.sysml.logic.bundle/pom.xml
seidewitz: Done.

AxelRICHARD:
@seidewitz no more remarks on my side, thank you

PR #780, opened 2026-07-20
ST6RI-953 - Extract parsing and linking completion concerns from SysML

adaussy:
model adapters

Bug: https://openmbee.atlassian.net/browse/ST6RI-953


seidewitz:
This pull request was not authorized within the SMC RIWG process.

PR #783, opened 2026-08-04
ST6RI-957 KerML 1.1 Ballot #3 - Abstract Syntax

seidewitz:
This PR implements resolutions for issues that were approved on KerML 1.1 RTF Ballot 3 related to abstract syntax.

**Changes**

- [KERML11-1](https://issues.omg.org/issues/KERML11-1) Binding connectors should not be considered owned cross features
    - Updated `Feature.ownedCrossFeature` operation per resolution.
- [KERML11-3](https://issues.omg.org/browse/KERML11-3) Featuring type of the multiplicity of a cross feature
    - Revised implementation of `checkMultiplicitRangeExpressionTypeFeaturing`.
    - **Note.** There was a previous pro-active implementation of a resolution for this issue, but it was removed, because it was not consistent with the RTF-approved resolution.
- [KERML11-68](https://issues.omg.org/browse/KERML11-68) End feature cannot be used in a feature chain expression
    - Revised implementation of check for `validationRedefinitionEndConformance` per resolution.

Resolutions for the following issues had been implemented previously.
- [KERML11-69](https://issues.omg.org/browse/KERML11-69) Indexing Collection types (PR #669)
- [KERML11-191](https://issues.omg.org/browse/KERML11-191) Problem with deriveTypeFeatureMembership (PR #749)

**Tests**

- Added `org.omg.sysml.logic.KERML11_Ballot3_Tests`, which includes standalone tests for KERML11-1, KERML11-3 and KERML-191, which affect only the logic module. (It was too difficult to set up a programmatic test for KERML-69.)
- Updated KerML Xpect test `Redefinition_End_Invalid.kerml.xt` for KERML11-68.

### DIFF
--- a/org.omg.sysml.logic/src/main/java/org/omg/sysml/adapter/ExpressionAdapter.java
+++ b/org.omg.sysml.logic/src/main/java/org/omg/sysml/adapter/ExpressionAdapter.java
@@ -113,11 +113,11 @@ public class ExpressionAdapter extends StepAdapter {
 		if (owner instanceof MultiplicityRange multiplicity &&
 			multiplicity.getBound().contains(expression)) {
 			owner = multiplicity.getOwningNamespace();
-			if (owner instanceof Feature) {
-				Namespace owningEnd = owner.getOwningNamespace();
-				if (FeatureUtil.isOwnedCrossFeature((Feature)owner) && 
+			if (owner instanceof Feature ownerFeature) {
+				if (FeatureUtil.isOwnedCrossFeature(ownerFeature) && 
 					isImplicitFeaturingTypesEmpty()) {
-					addFeaturingTypes(((Feature)owningEnd).getFeaturingType());
+					Feature owningEnd = (Feature) ownerFeature.getOwningNamespace();
+					addFeaturingTypes(owningEnd.getFeaturingType());
 				} else {
 					super.addImplicitFeaturingTypesIfNecessary();
 				}
--- a/org.omg.sysml.logic/src/main/java/org/omg/sysml/adapter/ExpressionAdapter.java
+++ b/org.omg.sysml.logic/src/main/java/org/omg/sysml/adapter/ExpressionAdapter.java
@@ -26,9 +26,11 @@ import org.omg.sysml.lang.sysml.Element;
 import org.omg.sysml.lang.sysml.Expression;
 import org.omg.sysml.lang.sysml.Feature;
 import org.omg.sysml.lang.sysml.FeatureValue;
-import org.omg.sysml.lang.sysml.Multiplicity;
+import org.omg.sysml.lang.sysml.MultiplicityRange;
+import org.omg.sysml.lang.sysml.Namespace;
 import org.omg.sysml.lang.sysml.Type;
 import org.omg.sysml.util.ExpressionUtil;
+import org.omg.sysml.util.FeatureUtil;
 import org.omg.sysml.util.ImplicitGeneralizationMap;
 
 public class ExpressionAdapter extends StepAdapter {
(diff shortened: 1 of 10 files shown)


org.omg.sysml.logic/src/main/java/org/omg/sysml/util/FeatureUtil.java
```diff
@@ -233,12 +234,15 @@ public static Feature getCrossFeatureOf(Feature feature) {
 	}
 
 	public static Feature getOwnedCrossFeatureOf(Namespace namespace) {
-		return !(namespace instanceof Feature) || !((Feature)namespace).isEnd()? null:
+		return !(namespace instanceof Feature feature) ||
```
AxelRICHARD: This condition is very difficult to read imho. The ternary operator does not help.
Could you consider rework it a little bit to be more readable please?

org.omg.sysml.logic/src/main/java/org/omg/sysml/adapter/ExpressionAdapter.java
```diff
@@ -99,20 +101,40 @@ protected List<Type> getGeneralTypes(Type type, Element skip) {
...
 	 * @satisfies checkMultiplicityRangeExpressionTypeFeaturing
 	 */
 	@Override
+	protected void addImplicitFeaturingTypesIfNecessary() {
+		Expression expression = getTarget();
+		Namespace owner = expression.getOwningNamespace();
+		if (owner instanceof MultiplicityRange multiplicity &&
+			multiplicity.getBound().contains(expression)) {
+			owner = multiplicity.getOwningNamespace();
+			if (owner instanceof Feature) {
```
AxelRICHARD: you could replace this by "owner instanceof Feature ownerFeature" and use _ownerFeature_ variable after that

org.omg.sysml.logic/src/main/java/org/omg/sysml/adapter/ExpressionAdapter.java
```diff
@@ -99,20 +101,40 @@ protected List<Type> getGeneralTypes(Type type, Element skip) {
...
+				Namespace owningEnd = owner.getOwningNamespace();
+				if (FeatureUtil.isOwnedCrossFeature((Feature)owner) &&
```
AxelRICHARD: `owningEnd` could be null? it is not tested in the _if_ statement just below

org.omg.sysml.logic/src/main/java/org/omg/sysml/adapter/ExpressionAdapter.java
```diff
@@ -99,20 +101,40 @@ protected List<Type> getGeneralTypes(Type type, Element skip) {
...
-	
+
 	/**
 	 * @satisfies checkExpressionTypeFeaturing
-	 * @satisfies checkExpressionResultBindingConnector
 	 * @satisfies checkMultiplicityRangeExpressionTypeFeaturing
 	 */
 	@Override
+	protected void addImplicitFeaturingTypesIfNecessary() {
+		Expression expression = getTarget();
```
AxelRICHARD, line 111: `expression` could be null? it is not tested in the code below

org.omg.sysml.logic/src/main/java/org/omg/sysml/util/FeatureUtil.java
seidewitz: I replaced the ternary conditional with an if statement. The second part of the condition (now in the "else" of the if) is uses stream operations to reflect the structure of the OCL for the `ownedCrossFeature` operation in the spec.

org.omg.sysml.logic/src/main/java/org/omg/sysml/adapter/ExpressionAdapter.java
seidewitz: OK, done (though `owner` is really only used one time in the if body specifically as a Feature).

org.omg.sysml.logic/src/main/java/org/omg/sysml/adapter/ExpressionAdapter.java
seidewitz: I have moved the call to `getOwningNamespace` inside the body of the following if statement. This makes it clear that the `owningEnd` is only used id `isOwnedCrossFeature` is true. One of the conditions for this to be true is that the feature has an owning namespace that is an end feature (i.e., non-null `owningEnd`).

org.omg.sysml.logic/src/main/java/org/omg/sysml/adapter/ExpressionAdapter.java
seidewitz: The target of an adapter can never be null within the Eclipse architecture.

AxelRICHARD (approved):

PR #786, opened 2026-08-10
Cross-file qualified references through a root Element short name fail to resolve (#778)

### LINKED ISSUE #778: Cross-file qualified references through a package's short name fail to resolve (global scope indexes `declaredName` only)
# Cross-file qualified references through a package's short name fail to resolve (global scope indexes `declaredName` only)

## Summary

A qualified name whose **first segment is the short name** (`<...>`) of a **top-level package defined in another file** fails to resolve. The same reference resolves when written with the package's `declaredName`, and short names resolve correctly in every *same-resource* position. The asymmetry comes from the global-scope index being built from `declaredName` only, while the local scope provider correctly matches both `memberName` and `memberShortName`.

## Environment

- Reproduced on the pilot as packaged in **`jupyter-sysml-kernel` 0.59.0** (`SysMLInteractive` / `IResourceValidator.validate(..., CheckMode.ALL, ...)`).
- Source references below are from tag **`2026-05`**. The relevant code path is unchanged there (see *Root cause*), so the behavior is expected to persist on the latest release.

## Steps to reproduce

Load each case as one or more `.sysml` files in the same resource set and validate.

**T1 — package short name as first segment, same file → OK**
```sysml
package <P> Alpha { part def Foo; }
package Beta { part x : P::Foo; }
```

**T2 — member short name as a non-first segment → OK**
```sysml
package Alpha { part def <F> Foo; }
package Beta { part x : Alpha::F; }
```

**T3 — member short name, unqualified, same namespace → OK**
```sysml
package Alpha { part def <F> Foo; part x : F; }
```

**T4 — package short name as first segment, references another file → FAILS**
```sysml
// a.sysml
package <P> Alpha { part def Foo; }
```
```sysml
// b.sysml
package Beta { part x : P::Foo; }
```

## Actual result

Only **T4** fails:

```
Couldn't resolve reference to Type 'P::Foo'.
An occurrence, item or part must be typed by occurrence definitions.   // cascading, from the now-untyped feature
```

Writing the same reference with the declared name — `part x : Alpha::Foo;` — resolves. Importing by short name across files (`private import P::*;`) fails the same way; `private import Alpha::*;` works.

## Expected result

T4 resolves, consistent with T1–T3. A package's short name is a valid `memberShortName` of its namespace, and per the KerML name-resolution rules a namespace provides names for its members via *both* `memberNames` **and** `memberShortNames`. There is no stated reason why short-name qualification should work within a resource but not across resources.

## Root cause

The local scope provider already handles short names. `KerMLScope.xtend` checks each membership under both names:

```xtend
// org.omg.kerml.xtext/.../scoping/KerMLScope.xtend
var memberName = ... else mem.memberName
if (checkElementName(memberName, qn, ...)) return true
...
var memberShortName = ... else mem.memberShortName
if (checkElementName(memberShortName, qn, ...)) return true
```

That is why T1–T3 (and non-first segments in general) work.

For a **cross-resource** reference, however, the **first** segment is resolved through the default global index, not by model traversal:

```xtend
// org.omg.kerml.xtext/.../scoping/KerMLGlobalScope.xtend  (getSingleElement)
val rootName = QualifiedName.create(qualifiedName.firstSegment)
val root = outer.getSingleElement(rootName)   // outer == default global scope (the exported index)
...
// subsequent segments are resolved by scopeFor(root as Namespace) — which DOES honor short names
```

The exported index is produced by `KerMLQualifiedNameProvider`, which derives qualified names from **`declaredName` only**:

```java
// org.omg.kerml.xtext/.../naming/KerMLQualifiedNameProvider.java
private Function<EObject, String> resolver =
        SimpleAttributeResolver.newResolver(String.class, "declaredName");
```

So a top-level element is indexed under its `declaredName`-based qualified name but **never under its `declaredShortName`**. When the first segment of a cross-resource reference is a short name, `outer.getSingleElement(rootName)` returns `null`, and resolution stops before the (short-name-aware) local scope is ever consulted.

This is long-standing and deliberate — introduced in the 2022-12 baseline (commit **ST6RI-628**, *"Updated KerMLQualifiedNameProvider to use declaredName for global scope name resolution"*) and unchanged through `2026-05`.

## Impact

Multi-file models that reference each other's top-level packages by short name do not validate. A common pattern is a project split into several files, each declaring a package with a short name (`package <ABC> SomeLongName { … }`) and referring to the others by that short name (`ABC::Something`). Under the pilot every such cross-file reference fails, and each failure cascades into secondary rule violations (e.g. `An occurrence, item or part must be typed by occurrence definitions.`, `A requirement must be typed by one requirement definition.`) on the now-untyped features. Rewriting every short-name qualifier to the corresponding `declaredName` makes the whole model validate with zero errors, which confirms the single root cause.

## Suggested fix

Make cross-resource resolution consistent with local resolution by exposing short names to the global index. Options:

1. In `KerMLQualifiedNameProvider`, also produce a `declaredShortName`-based qualified name (e.g. override `getFullyQualifiedNames` / contribute an additional `IEObjectDescription`), so a top-level element is indexed under both its name and its short name; **or**
2. In `KerMLGlobalScope.getSingleElement`, fall back to looking up the first segment by short name when the `declaredName` lookup misses.

The local scope already honors `memberShortName`, so the change is isolated to the export/index side.

tkanov:
Fixes #778. A qualified name whose first segment is the short name of a root Element does not resolve across resources: `part x : P::Foo;` fails where `part x : Alpha::Foo;` succeeds, and `private import P::*;` fails likewise, although short names resolve in every same-resource position.

**Cause.** `KerMLQualifiedNameProvider` exports `declaredName`-based names only, so a root Element is never indexed under its `declaredShortName` – and `KerMLGlobalScope.getSingleElement` resolves the first segment of a cross-resource qualified name through that index, before the short-name-aware `KerMLScope` is reached.

**Changes**

1. `KerMLResourceDescriptionStrategy` exports root Elements under their escaped `declaredShortName`. Nested Elements are unaffected, since `KerMLScope` already handles those.
2. Short names are exported under a separate qualifier (`<short-name>`) and consulted by `KerMLGlobalScope.getRootElement` only after a `declaredName` lookup misses, so a declared name takes precedence structurally rather than by index order. Without this, `package <ScalarValues> MyPackage` in one resource made `ScalarValues::Real` unresolvable in another.
3. Bound via `bindIDefaultResourceDescriptionStrategy` in `KerMLRuntimeModule`, `SysMLRuntimeModule`, `KerMLxRuntimeModule` and `SysMLxRuntimeModule`. Also exported `org.omg.kerml.xtext.resource`.
4. `ShortNameScopeTest`, 7 cases. The cross-resource short-name cases fail before change 1. The three collision cases fail before change 2.

`org.omg.sysml.interactive.tests` (215), `org.omg.kerml.xpect.tests` (963) and `org.omg.sysml.xpect.tests` (308) all pass.

**Notes**

This applies to every root Element with a short name, not only Packages – a root `part def <F> Foo;` now resolves as `F` – consistent with a Namespace giving its members both `memberName` and `memberShortName`.

Where a short name and a declared name collide, the declared name wins and the short name is unreachable as a first segment. That cannot break a reference that resolves today, but it is a semantics call, so I am happy to report the collision as an error instead. Related: the `declaredName`-only index looks deliberate (ST6RI-628), so a check against the original intent may be warranted.

Also included, unrelated and latent: `KerMLScope.qualifiedNameConverter` is `@Inject`ed but every `KerMLScope` is constructed with `new`, so it is always null and `getElement(String)` would fail. It now comes from the injected `KerMLScopeProvider`. Happy to split out.


seidewitz:
Sorry, we cannot accept pull requests from outside our development team. We will take your issue (#778) into consideration (this is actually a known problem).

If you would like to contribute the the Pilot Implementation, consider joining (or having your organization join) the [OMG Systems Modeling Community](https://www.omg.org/communities/systems-modeling-community.htm). The Pilot Implementation is maintained by the Reference Implementation Working Group of the SMC.

PR #793, opened 2026-08-31
ST6RI-953 Extract parsing and linking completion concerns to Xtext modules

### LINKED ISSUE #761: Problem with Commit : ST6RI-897 Moved name escaping/body processing to adapter postProcess.
Hello SysML Team!
Since this commit moves from setDeclaredName to adapter's postProcess, and not all adapter's post process call super.postProcess(), whick makes the name handles differently for different kind of elements, is this modification work as design?

Best reguards!

adaussy:
This PR moves parsing/linking-specific `postProcess()` behavior out of the model adapters and into dedicated Xtext parser adapters.

**Background**

The textual notation grammars in the KerML and SysML specifications have several productions that parse some segment of text into an element that is synthesized into a meta-property, but then also assigned to another property. This second step is done using the `{ ... }` notation in the specification BNF. 

For example, in the specification BNF, an owned `Specification` is parsed this way:
```
OwnedSpecialization : Specialization =
    GeneralType

GeneralType : Specialization =
      general = [QualifiedName]
    | general = OwnedFeatureChain
      { ownedRelatedElement += general }
```
That is, and owned `Specialization` is specified with just a `general` type (the owner is the `specific` type). The `general` type is parsed as either as a named reference or an owned feature chain. In the latter case, the feature chain is also added to the `ownedRelatedElements` to establish ownership.

Unfortunately, the Xtext grammar language doesn't support doing this sort of thing. In the Xtext grammars, the corresponding production is
```
OwnedSpecialization returns SysML::Specialization :
	  general = [SysML::Type | QualifiedName]
	| ownedRelatedElement += OwnedFeatureChain
;
```
That is, an `OwnedFeatureChain` gets added to `ownedRelatedElement` (because Xtext requires every element in the parse tree to have an owner), but does _not_ get assigned to `general`.

In order to make up for Xtext deficiencies such as this, after parsing a model, the implementation does post-processing on each model element. For example, the post-processing for `Specialization` checks whether `general` is null. If so, it sets `general` to the first `ownedRelatedFeature`. In the end, this achieves the parsing required by the specification production.

Previously, post-processing code was included in a `postProcess` method on the `ElementAdapters` for elements that needed post-processing. However, this post-processing is specific for handling the limitations of the Xtext grammar language. Other parsers may not have such limitations, or may have different ones. Since the code in `org.omg.sysml.logic` (where the adapters now are) is supposed to be parser-independent, the Xtext parser-post-processing code needed to be separated from the other adapter code.

**Changes**
1. Added `omg.org.kerml.xtext.postprocessing` package, with `KerMLParserPostProcessorFactory` to instantiate various stateless`ParserPostProcessor` classes for KerML element post-processing. Updated `org.omg.kerml.scoping.KerMLLinker.postProcessAll()` to use the new factory, while preserving previous functionality.
2. Added `omg.org.sysml.xtext.postprocessing` package with `SysMLParserPostProcessorFactory`, extending `KerMLParserPostProcessorFactory` and adding additional `ParserPostProcessor` classes for SysML. Added `org.omg.sysml.scoping.SysMLLinker` to use the new `SysMLParserPostProcessorFactory`.

    This keeps parser-specific model completion concerns in the Xtext layer and separates them from the long-lived semantic model adapters.

3. Revised `postProcess` methods as necessary so they all call `super` (except for the top-level `ElementParserPostProcessor`).


seidewitz:
I made two changes myself.

1. Moved `ParserAdapterTest` and `UsagePostProcessingTest` from `org.omg.sysml.semantics.tests` to `org.omg.sysml.interactive.tests`. Parser post-processing is syntactic, not semantic. The `semantics.tests` are only for testing implementation of implied relationships (based on semantic constraints). I also revised `UsagePostProcessingTest` to use the interactive test parsing framework (it didn't seem worth doing this for `ParserAdapterTest`.
2. Updated `postProcess` methods as necessary so they always call super. This has always been the intent, even though it seems quite a few of the post processing methods for various kinds of relationships were missing these calls. Fortunately, it seems that the only post processing that was missed in these cases was the unescaping of declared names (in `ElementParserAdapter`).

org.omg.kerml.xtext/src/org/omg/kerml/xtext/scoping/KerMLLinker.java
```diff
@@ -70,14 +72,29 @@ protected void postProcessAll(EObject model) {
...
+							doGetAdapter(element).postProcess();
+						}
+					}
+				} finally {
+					TreeIterator<EObject> iterator = getAllLinkableContents(model);
+					while (iterator.hasNext()) {
+						EObject obj = iterator.next();
+						if (obj instanceof Element element) {
+							KerMLParserAdapterFactory.removeAdapter(element);
+						}
```
seidewitz: When I saw that you were adding and then removing the `ParserAdapters`, I realized that these classes should actually no be adapters at all. The reason to add an adapter to an element is so that it can maintain possibly changing state over time. But this not the case for the so-called `ParserAdapters`. They are just functionality that is applied once to an element when it is first parsed.

So, I recommend the following:
1. Rename the `...ParserAdapter` classes to `...ParserPostProcessor`. 
2. Revise the factories for KerML and SysML post-processor classes to _not_ add them to the `eAdapters` list or check whether there is already a post-processor instantiated for an element. (They can still use switches for selecting which class to instantiate.)
3. Here in `postProcessAll`, remove the `try ... finally ...` block (but keep the loop calling the post-processors!).

org.omg.sysml.xtext/src/org/omg/sysml/xtext/adapter/SysMLParserAdapterFactory.java
```diff
@@ -0,0 +1,102 @@
...
+		if (element instanceof PortUsage portUsage) {
+			adapter = new PortUsageParserAdapter(portUsage);
+		// Some concrete OccurrenceUsage subtypes have multiple inheritance. In the generated
+		// SysMLSwitch, caseFeature may then be visited before caseOccurrenceUsage and would select
+		// a FeatureParserAdapter, skipping the occurrence-specific portion post-processing.
+		} else if (element instanceof OccurrenceUsage occurrenceUsage) {
+			adapter = new OccurrenceUsageParserAdapter(occurrenceUsage);
+		// For the same reason, some other concrete Usage subtypes reach caseFeature before caseUsage.
+		// Selecting UsageParserAdapter here guarantees that their common Usage post-processing runs.
+		} else if (element instanceof Usage usage) {
```
seidewitz: We also had issues with element multiple inheritance when doing the logic adapters. However, using if statements outside the switch to resolve this can be brittle and easy to overlook in later maintenance (even with your helpful comments). 

For the logic adapters, what we ended up doing was creating adapter classes for all classes with multiple inheritance and all intermediate classes, generally following the inheritance structure of the metamodel `Impl` classes.

The fact that the logic adapter `PortUsageAdapter` extends `UsageAdapter` instead of `OccurrenceUsageAdapter` seems to actually be a bug. According to the SysML Specification, a port can be snapshot or timeslice or individual like any other occurrence, and that is handled in the `OccurrenceUsageAdapter`. I will create a bug report for this, which we can fix after this PR is merged. But we shouldn't compound the problem here, so `PortUsageParsePostProcessor` (currently called `PortUsageParseAdapter`) should extend `OccurrenceUsagePostProcessor`, _not_ `UsageParsePostProcessor`.

For other subtypes with multiple inheritance, the idea would be to create post-processor classes for them, even though they have no specific post-processing functionality. They can then be part of a single-inheritance `extend` hierarchy that provides a path to the correct post-processing. For example, `ConnectionUsagePostProcessor` would be in the hierarchy ConnectionUsagePostProcessor` ->`PartUsagePostProcessor` -> `ItemUsagePostProcessor` -> `OccurrenceUsagePostProcessor`.

This requires some more effort up-front, but experience has shown it is easier to maintain later on.

org.omg.sysml.xtext/src/org/omg/sysml/xtext/postprocessing/PortUsageParserPostProcessor.java
```diff
@@ -0,0 +1,39 @@
...
+
+package org.omg.sysml.xtext.adapter;
+
+import org.eclipse.emf.ecore.EObject;
+import org.omg.sysml.lang.sysml.PortDefinition;
+import org.omg.sysml.lang.sysml.PortUsage;
+import org.omg.sysml.util.UsageUtil;
+
+public class PortUsageParserAdapter extends UsageParserAdapter {
+
```
seidewitz, line 20: Per my previous comment, this should extend `OccurrenceParserAdapter`/`OccurrenceParserPostProcessor`. It is a bug that the logic adapter `PortUsageAdapter` extends `UsageAdapter` and not `OccurrenceUsageAdapter`.

seidewitz (changes requested):

org.omg.kerml.xtext/src/org/omg/kerml/xtext/scoping/KerMLLinker.java
adaussy: I totally agree with you and it was in my initial design.

org.omg.sysml.xtext/src/org/omg/sysml/xtext/adapter/SysMLParserAdapterFactory.java
adaussy: Ok. I have made PortUsageParserPostProcessor inherite OccurrenceUsageParserPostProcessor but I'm not about the "other subtypes". Do you expect me to do something?

org.omg.sysml.xtext/src/org/omg/sysml/xtext/postprocessing/PortUsageParserPostProcessor.java
adaussy: Done.

adaussy:
@seidewitz I have a new version I hope that match all your comment because I'm not sure about what to do with "For other subtypes with multiple inheritance"

seidewitz:
> @seidewitz I have a new version I hope that match all your comment because I'm not sure about what to do with "For other subtypes with multiple inheritance"

Yeah, I wasn't entirely clear here. What I meant was to fill in the post-processors for intermediate superclasses of any existing processing class, so the post-processor inheritance hierarchy for KerML and SysML is a subset of the abstract syntax hierarchy, with no level skipped. I went ahead and pushed a commit with the additional post-processor classes.

Please review my latest update. If you agree with it, we can go ahead and merge this PR. Or, if you see problems, please comment.

seidewitz:
I also update `ParserPostProcessorTest`, because I tracked some intermittent build failures to it. It turns out that the logic module setup you were doing in this test class was interacting badly with the Xtext-based `SysMLInteractiveTest` framework. This somehow caused problems with the global-scope name resolution. In any case, you don't need to initialize the logic module for this test, because you are not actually using any of the logic capabilities (derived properties, operations and semantic transformation). I removed the logic module set up, and now it builds without any problem.

seidewitz (approved):

adaussy:
> > @seidewitz I have a new version I hope that match all your comment because I'm not sure about what to do with "For other subtypes with multiple inheritance"
> 
> Yeah, I wasn't entirely clear here. What I meant was to fill in the post-processors for intermediate superclasses of any existing processing class, so the post-processor inheritance hierarchy for KerML and SysML is a subset of the abstract syntax hierarchy, with no level skipped. I went ahead and pushed a commit with the additional post-processor classes.
> 
> Please review my latest update. If you agree with it, we can go ahead and merge this PR. Or, if you see problems, please comment.

OK this is what I tough but I was not sure if you expect me to do it with this PR or later. Thanks you for making it happen.

I reviewed it quickly and it seems ok. You can merge this PR. (I'm not sure if I'm allowed to merge it myself)

When its done I will rebase my change for ST6RI-950.

Thanks for your review.

